PR #1, opened 2015-04-21
Dev/15.0/dizzy

bmouring:
We reference internal NI servers, preventing parsing from happening. Revert to the yocto kernels for the time being.


bmouring:
I'll re-request with the updated commit message

PR #9, opened 2020-01-30
NILRT_GIT isn't publicly available for librtpi

pschneiders:
The git repository pointed at by meta-nilrt/recipes-rt/librtpi/librtpi_0.0.1.bb does not appear to exist or is not publicly available.  This changes allows builds to complete.


amstewart:
Thanks for bringing this to my attention!

I'm going to push a fixup which will direct the recipe to [https://github.com/gratian/librtpi/tree/ni/latest](https://github.com/gratian/librtpi/tree/ni/latest); which has the latest commits that we're referencing, and I expect will eventually be the canonical upstream.

I'll update this PR when it is pushed.

amstewart:
Pushed an updated SRC_URI in commit [51504c1ed9b8c1116631df4fbe5ae55c540387d3](https://github.com/ni/meta-nilrt/commit/51504c1ed9b8c1116631df4fbe5ae55c540387d3).

Thanks!

PR #11, opened 2020-06-17
Accomplish basic layer compatibility with oe/dunfell

amstewart:
This patchset establishes basic layer compatibility with the dunfell release of OE-core, specifically those rebased layer branches available at:
github.com/amstewart/$layer:dev/dunfell/master.

The recipe commits in this patchset allow a selection of essential tool recipes (eg. glibc, busybox, util-linux, et c.) to succeed. Further layer fixes are required for full dunfell compatibility, but that work will be tracked in other work-items and manifest in later pull requests.

@ni/rtos


shruthi-ravi (commented):
Are the cp936 patches here related to the condvar patches?

amstewart:
> Are the cp936 patches here related to the condvar patches?

Hmm... I'm not sure, but I don't see anything which obviously says that they are.
@gratian Can you comment on if we should carry and reapply those few cp936 patches to the new glibc?

recipes-core/busybox/busybox/defconfig
```diff
@@ -78,10 +81,12 @@ CONFIG_NO_DEBUG_LIB=y
 # CONFIG_EFENCE is not set
 
 #
-# Busybox Library Tuning
+# Library Tuning
 #
 # CONFIG_FEATURE_USE_BSS_TAIL is not set
+CONFIG_FLOAT_DURATION=y
 CONFIG_FEATURE_RTMINMAX=y
+CONFIG_FEATURE_RTMINMAX_USE_LIBC_DEFINITIONS=y
```
shruthi-ravi, line 89:
```suggestion
# CONFIG_FEATURE_RTMINMAX_USE_LIBC_DEFINITIONS is not set
```
based on this [file](https://github.com/openembedded/openembedded-core/blob/dunfell/meta/recipes-core/busybox/busybox/defconfig#L88). Or is this something you intended to do?

gratian:
> > Are the cp936 patches here related to the condvar patches?
> 
> Hmm... I'm not sure, but I don't see anything which obviously says that they are.
> @gratian Can you comment on if we should carry and reapply those few cp936 patches to the new glibc?

The cp936 patches are not related to condvar support. They are likely related to simplified Chinese localization support. We do support simplified Chinese localization on LV RT and as such they probably need to be carried forward.

recipes-core/util-linux/util-linux_2.%.bbappend
```diff
@@ -1,6 +1,6 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
-DEPENDS_class-target += "shadow-native pseudo-native niacctbase"
```
shruthi-ravi, line 3: Would you get an error if you tried to build the recipe with a missing dependency?

recipes-core/busybox/busybox/defconfig
```diff
@@ -658,7 +658,7 @@ CONFIG_MOUNT=y
 # CONFIG_FEATURE_MOUNT_VERBOSE is not set
 # CONFIG_FEATURE_MOUNT_HELPERS is not set
 # CONFIG_FEATURE_MOUNT_LABEL is not set
-CONFIG_FEATURE_MOUNT_NFS=y
+CONFIG_FEATURE_MOUNT_NFS=n
```
shruthi-ravi, line 661:
```suggestion
# CONFIG_FEATURE_MOUNT_NFS is not set
```

recipes-core/packagegroups/packagegroup-ni-base.bb
```diff
@@ -118,7 +119,6 @@ RDEPENDS_${PN} = "\
 	opkg-keyrings \
 	os-release \
 	run-postinsts \
-	start-stop-daemon \
```
shruthi-ravi, line 121: codesearch reveals that a lot of people use start-stop-daemon! Are you making a list of breaking changes that we can tell clients upfront? This should be added to that.

conf/distro/nilrt.conf
```diff
@@ -2,9 +2,9 @@ require nilrt.inc
```
shruthi-ravi, line 2: I suppose you won't be able to edit the commit message to say LV 2021 now?

recipes-core/util-linux/util-linux_2.%.bbappend
amstewart: Presumably, a build-time dependency which you actually need would cause some kind of error at build. The commits which added the original DEPENDS didn't go into detail about why `niacctbase` was necessary. I would assume it was thought to be, since the `niacctbase` recipe does some configuration of the lvrt user and NI-specific groups.

But I don't believe that recipe is a real dependency - the do_install operations in util-linux seem independent of what `niacctbase` is doing in the sysroot.

recipes-core/packagegroups/packagegroup-ni-base.bb
ScotSalmon: Um, yeah. We're really not supporting that anymore? That will break a ton of stuff.

amstewart: The `start-stop-daemon` binary is still provided, just through the `dpkg-start-stop` package in oe-core, rather than the `start-stop-daemon` recipe in meta-oe. Meta-oe removed their recipe because the one in oe-core is newer (and from the same source).

So initscripts can continue to use `start-stop-daemon` in the same way. The only impact is if they have declared an IPK dependency on `start-stop-daemon`, it will need to be updated to `dpkg-start-stop`.

conf/distro/nilrt.conf
amstewart: Oops. I'll fix that up in my v2. :)

recipes-core/busybox/busybox/defconfig
amstewart: The default value for this feature (which doesn't apply for kernels >2.6) is 'no'. So your suggestion seems like a NOOP. Why do you think it should be "not set" rather than explicitly 'no'?

recipes-core/busybox/busybox/defconfig
amstewart: It was enabled by the `make oldconfig` operation in busybox because we had the parent `CONFIG_FEATURE_RTMINMAX` enabled already. Codesearching around for which signal definitions we use suggests that both NI and OE-upstream prefer to use the LIBC definitions. So we should keep this option enabled (which is the default anyway).

@gratian ACKed that assessment.

amstewart:
# Patch V2
* Updated `glibc: remove 2.24 patches` to only remove the condvars patchset.
* Added a new glibc commit to refresh the cp936 patches against glibc_2.31.
* Fixed typo in `nilrt.conf: rev distro version to 9.0`.

amstewart:
# Patch V3
* Added `busybox: unset apparently unused defconfig values` to address feedback from @shruthi-ravi about unsetting unused values.

shruthi-ravi (approved):
Looks good!

PR #12, opened 2020-07-08
Updated /var/cache to a non-volatile storage location.

shruthi-ravi:
AzDo work item [US1023816](https://ni.visualstudio.com/DevCentral/_workitems/edit/1023816).

Tested changes on a cRIO-9068 and cRIO-9037:
1. /var/cache was mounted on the root filesystem.
2. opkg was able to install software
2. /var/cache directory was cleaned up on shutdown and initialization during a reboot

@ni/rtos

### DIFF
--- a/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
+++ b/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
@@ -32,6 +32,7 @@ SRC_URI = "file://nisetbootmode \
            file://nisetupirqpriority \
            file://nipopulateconfigdir \
            file://run-ptest \
+           file://cleanvarcache \
 "
 
 SRC_URI_append_x64 = "file://nidisablecstates \
@@ -70,6 +71,7 @@ do_install () {
 	install -m 0755 ${WORKDIR}/nisetupkernelconfig   ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/nisetcommitratio      ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/wirelesssetdomain     ${D}${sysconfdir}/init.d
+	install -m 0755 ${WORKDIR}/cleanvarcache         ${D}${sysconfdir}/init.d
 	install -m 0755 ${S}/nisetupirqpriority          ${D}${sysconfdir}/init.d
 
 	install -d ${D}${sysconfdir}/natinst
@@ -85,6 +87,7 @@ do_install () {
 	update-rc.d -r ${D} nisetupkernelconfig   start 3  4 5 .
 	update-rc.d -r ${D} nisetcommitratio      start 99 S .
 	update-rc.d -r ${D} wirelesssetdomain     start 36 S .
+	update-rc.d -r ${D} cleanvarcache         start 38 0 6 S .
 
 	# only for nilrt-nxg, on older nilrt it's installed via p4
 	if ${@oe.utils.conditional('DISTRO', 'nilrt-nxg', 'true', 'false', d)}; then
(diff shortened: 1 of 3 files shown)


recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
```diff
@@ -85,6 +87,7 @@ do_install () {
 	update-rc.d -r ${D} nisetupkernelconfig   start 3  4 5 .
 	update-rc.d -r ${D} nisetcommitratio      start 99 S .
 	update-rc.d -r ${D} wirelesssetdomain     start 36 S .
+	update-rc.d -r ${D} cleanvarcache         start 38 0 1 6 S .
```
chaitu236: According to [https://wiki.debian.org/BootProcess](https://wiki.debian.org/BootProcess), Runlevel _1_ is entered when switched from another runmode. So we don't need to specify it as _S_ takes care of running cleanvarcache during boot time.

bstreiff: I agree, this script seems like it could be specified as `start 38 S .` similar to how `nicleanstalelinks` is.

chaitu236: Shouldn't we keep 1 and 6 also, in case next boot is safemode?

bstreiff: We enter runlevel `6` when issuing a `reboot`, so if we want to clear /var/cache prior to reboot then yeah, adding it to `6` start could be appropriate.

Where would we enter runlevel `1`?

chaitu236: Sorry I meant 0 not 1 - we'd want to clean up on halt also

bstreiff: Oh, yeah, in that case, I agree-- if we clean up on shutdown (6) we should do the same on halt (0).

shruthi-ravi: I'll remove 1. I've just thought about another scenario that isn't handled. When there's a power outage of some sort and the target reboots into safe mode, cleanup doesn't happen until the next time the target reboots into runmode..

chaitu236: Isn't initscripts-nilrt_1.0.bb applicable to both safemode and runmode images ([this](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-core/packagegroups/packagegroup-ni-base.bb#L98), [this](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-core/images/niconsole-image-safe.bb#L4) [this](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-core/images/niconsole-image.inc#L16) suggest it is)? - in which case cleanup will happen during boot in either mode .. and my previous comment about needing 0 and 6 isn't accurate since we have S (but I prefer we still clean up in 0 and 6).

shruthi-ravi: Right, you are. I verified that safemode image had the script too.

chaitu236 (approved):

amstewart (changes requested):
Looks good to me. A few nits about the commit message.

Could you change the short message to present-tense, like:
`update /var/cache to a non-volatile storage location`

And fixup a typo in the message:

Since opkg uses /var/cache as its cache location, when /var/cache is set
to volatile storage, distribution upgrades could use up all system memory
and result in OOM on low memory targets. Move~~d~~ /var/cache to a non-volatile
location to address this. Add~~ed~~ an init script to delete the /var/cache
directory on startup and shutdown to clean up after any NI software that
assume**s** /var/cache is a volatile storage location.

amstewart (approved):

PR #13, opened 2020-07-13
local.conf.sample: Do not parametrize BUILDNAME to prevent reproducibility errors

chaitu236:
Signed-off-by: Chaitanya Vadrevu <chaitanya.vadrevu@ni.com>

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -250,4 +250,4 @@ CONF_VERSION = "1"
 # concatentation of the currrent $DATE and $TIME - which is not very meaningful
 # and can lead to taskhash issues in recipes which are sensitive to it (like
 # os-release.)
-BUILDNAME ?= "dev-${MACHINE_ARCH}"
+BUILDNAME ?= "dev-nilrt"


amstewart:
Do you have commentary on why using the `MACHINE_ARCH` in the build name produces hash mismatch errors? It's value should never change over the course of an image build, so is the error spurious?

chaitu236:
> TUNE_FEATURES{mx32} = UnsetAR${HOST_PREFIX}arARCH${@map_kernel_arch(d.getVar('TARGET_ARCH'), d)}AS${HOST_PREFIX}as ${HOST_AS_ARCH}ASNEEDED-Wl,--as-neededB${WORKDIR}/linux-${PACKAGE_ARCH}-${LINUX_KERNEL_TYPE}-buildBUILDNAMEdev-${**MACHIN  E_ARCH**}BUILD_AR

This is the log from the first time basehash is calculated - MACHINE_ARCH is not set by this time.
MACHINE_ARCH is set only in subsequent basehash calculations - it is not clear why this is the case though.

Replacing MACHINE_ARCH with other variables (like MACHINE) results in same behavior.

PS: This is not spurious and is always reproducible

amstewart:
Huh. I suppose that explains why it doesn't recur on subsequent builds. OK. Ship it.

amstewart:
@chaitu236 Could you submit a similar PR to the dunfell branch, when you're ready.

PR #14, opened 2020-07-13
Remove DISTRO_FEATURES_LIBC

joshcosta:
> The DISTRO_FEATURES_LIBC variable is no longer used. Remove it from meta-nilrt/conf/distro/nilrt.inc
> 
> https://www.yoctoproject.org/docs/3.0.2/ref-manual/ref-manual.html

Requires verification be done later as discussed. I did not build dunfell.

### DIFF
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -4,9 +4,6 @@ TARGET_VENDOR = "-nilrt"
 
 GRUB_BRANCH = "nilrt/19.0"
 
-# Inherit the default LIBC features superset from OE-core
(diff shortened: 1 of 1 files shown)


amstewart:
lgtm; Did you check to make sure that we don't depend on the libc features set?

shruthi-ravi (commented):
Can you sign off on the commit? And a second line of description to the commit, something like "This variable is no longer used per Yocto.".

joshcosta:
> lgtm; Did you check to make sure that we don't depend on the libc features set?

Yes it looks like any references to it were updated when comparing sumo to dunfell.

PR #17, opened 2020-07-20
glibc: fix EAGAIN retry loop in pthread_cond_broadcast()

gratian:
Commit ac38b649bcb9 ("glibc: pi-condvars: EAGAIN retry in
pthread_cond_ broadcast() and signal()") introduced a retry loop for
an EAGAIN error on a FUTEX_CMP_REQUEUE_PI futex syscall in
pthread_cond_broadcast(). However the futex value is not re-read
before re-invoking the futex syscall on retry. This can result in an
infinite retry loop if the reason the kernel returned EAGAIN is
because the futex value changed after the private lock was released.

Update the futex value and other associated sequence counters before
retrying the syscall.

Fixes: ac38b649bcb9 ("glibc: pi-condvars: EAGAIN retry in pthread_cond_ broadcast() and signal()")
Natinst-AZDO-ID: 1024009

### DIFF
--- a/recipes-core/glibc/glibc/0006-pi-condvars-EAGAIN-retry-in-pthread_cond_-broadcast-and-signal.patch
+++ b/recipes-core/glibc/glibc/0006-pi-condvars-EAGAIN-retry-in-pthread_cond_-broadcast-and-signal.patch
@@ -14,16 +14,31 @@ https://elixir.bootlin.com/linux/v4.9.47/source/kernel/futex.c#L1788
 Adds retry loop when lll_futex_cmp_requeue_pi() returns EAGAIN.
 
 Signed-off-by: Haris Okanovic <haris.okanovic@ni.com>
+[gratian: update the futex value when retrying EAGAIN on broadcast]
+Signed-off-by: Gratian Crisan <gratian.crisan@ni.com>
 ---
- nptl/pthread_cond_broadcast.c | 12 +++++++++---
+ nptl/pthread_cond_broadcast.c | 19 +++++++++++++++----
  nptl/pthread_cond_signal.c    | 35 +++++++++++++++++++++--------------
- 2 files changed, 30 insertions(+), 17 deletions(-)
+ 2 files changed, 36 insertions(+), 18 deletions(-)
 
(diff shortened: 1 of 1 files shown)


amstewart:
ACKed for the OE content.

Is it correct that we expect the glibc+librtpi implementation in dunfell is immune from this issue?

gratian:
Some notes for reviewers:

- tested with .c test case attached to associated Azdo work item (no live locks observed).
- LabVIEW 2020 and later uses librtpi so running the RT ATS doesn't add additional coverage.

Attn: @harisokanovic

gratian:
> Is it correct that we expect the glibc+librtpi implementation in dunfell is immune from this issue?

Yes. That is correct (assuming glibc was upgraded to the dunfell version).

bstreiff (approved):
LGTM. (Reviewing a diff of a diff is a bit mind-bending.)

gratian:
Further explanation if you are reviewing this and are wondering about this part of the change:

```
-      cond->__data.__futex = (unsigned int) cond->__data.__total_seq * 2;
+      ++cond->__data.__futex;

```
The futex word value (i.e. the value of cond->__data.__futex) is used for race detection. The user space code increments this value for: wait, timedwait, signal, cancellation, and timeout. In the broadcast case the value was set to "cond->__data.__total_seq * 2". The kernel  in the futex syscall can then detect a race by comparing the currently stored in the futex word with the futex value passed in by user space. The current futex word value has no meaning associated with it for the futex operations used in this implementation (i.e. all the kernel cares about is the fact that the value changed):

```
lll_futex_cmp_requeue_pi	FUTEX_CMP_REQUEUE_PI
lll_futex_requeue		FUTEX_CMP_REQUEUE
lll_futex_wait_requeue_pi	FUTEX_WAIT_REQUEUE_PI
lll_futex_wait			FUTEX_WAIT
lll_futex_timed_wait_requeue_pi	FUTEX_WAIT_REQUEUE_PI
lll_futex_timed_wait		FUTEX_WAIT
lll_futex_timed_wait_bitset	FUTEX_WAIT_BITSET
```
Given that the futex value is not used directly using an increment for the broadcast case simplifies the retry code. All we need to do is ensure that we increment the value before retrying so other concurrent waiters/signalers can detect the race.

PR #18, opened 2020-07-24
lsbinitscripts: Add recipe

chaitu236:
lsb support was removed from upstream but lot of initscripts
in NI drivers use functions from lsbinitscripts. So adding lsbinitscripts
from sumo back to meta-nilrt.

Also removed dead code referring to lsb and its files

Signed-off-by: Chaitanya Vadrevu <chaitanya.vadrevu@ni.com>


chaitu236:
@amstewart

recipes-extended/lsb/lsbinitscripts/functions.patch
```diff
@@ -0,0 +1,33 @@
+From e46b056282c8420f096d5c34d78c00f816788784 Mon Sep 17 00:00:00 2001
+From: Fan Xin <fan.xin@jp.fujitsu.com>
+Date: Mon, 5 Jun 2017 16:26:47 +0900
+Subject: [PATCH 1/2] Upstream-Status: Inappropriate [configuration]
```
amstewart: Could you comment on whether or not this patch is necessary for NILRT; it seems like something we could drop.

Which package ships the `/sbin/consoletype` binary?

recipes-core/images/host-toolchain-sysroot.bb
```diff
@@ -60,8 +60,6 @@ create_sdk_files_append() {
     bbnote "Target Sysroot: Remove escaping symlinks"
     rm -Rf "${SDK_OUTPUT}/${SDKTARGETSYSROOT}/etc/mtab"
     rm -Rf "${SDK_OUTPUT}/${SDKTARGETSYSROOT}/var/lock"
-    rm -Rf "${SDK_OUTPUT}/${SDKTARGETSYSROOT}/usr/lib/lsb/install_initd"
-    rm -Rf "${SDK_OUTPUT}/${SDKTARGETSYSROOT}/usr/lib/lsb/remove_initd"
```
amstewart, line 64: Could you split this patchset into two commits: one which removes the deprecated LSB references, and one which adds the lsbinitscripts recipe to meta-nilrt.

recipes-extended/lsb/lsbinitscripts/0001-functions-avoid-exit-1-which-causes-init-scripts-to-.patch
```diff
@@ -0,0 +1,21 @@
+Upstream-Status: Pending
```
amstewart: Has this patch been upstreamed to lsbinitscripts? If so, is there anything stopping us from bumping the source version of the recipe, rather than carrying this `.patch`?

recipes-extended/lsb/lsbinitscripts/functions.patch
chaitu236: The patch checks for existence of /sbin/consoletype before using it, so I think it is required - if we want to ship the same version of initscripts.

But we can bump to latest version so this patch wouldn't be required (as the relevant code was [removed](https://github.com/fedora-sysv/initscripts/commit/f88dbd98e992635ee569fb6349f49cb6d215c434)).

None of our feeds contain a package that provides /sbin/console, at-least not the feeds that are shipped with target.

recipes-extended/lsb/lsbinitscripts/0001-functions-avoid-exit-1-which-causes-init-scripts-to-.patch
chaitu236: It was never upstreamed so we'd still need this patch.

amstewart: Was it never even submitted to the sysvinit upstream?

The more I look at this .patch, the more I don't understand it. I don't see how the `strstr` function would "exit 1" in any case. Do you have an understanding of what Chen Qi was trying to fix here?

recipes-extended/lsb/lsbinitscripts/functions.patch
amstewart: Oh. Could you do an evaluation of what it would mean to upgrade the recipe to the latest (or later) version of `sysvinit` then. ie. what commits have gone into the upstream `functions` implementation between SRCREV `a51c1b4f7dcf55b568b2ee4c2b18078849943469` and the new rev?

recipes-extended/lsb/lsbinitscripts/0001-functions-avoid-exit-1-which-causes-init-scripts-to-.patch
chaitu236: Don't see any PR (open/closed) in upstream repo, so doesn't look like it was submitted.

[strstr](https://github.com/fedora-sysv/initscripts/blob/master/etc/rc.d/init.d/functions#L602) returns 1 when $1 doesn't contain $2 ([ref](https://stackoverflow.com/a/41806827)). So seems like a legitimate patch.

amstewart: `exit` and `return` are unrelated calls in bash. The `strstr` function returning 0 or 1 shouldn't ever cause either `functions` or a sourcing script to `exit` with code 1, *unless* the returning function is the last call which is made in the script. In which case, the script would `exit` with the last return value.

But since the final call in the script is always `return 0`, I think the bash exit code should always be 0 (unless there is a legitimate bash error).

So, I'm still not convinced this does something. Have you found any evidence of a reproducing case for this change? If it is legitimate, I would like us to adopt the patch and try to upstream it.

chaitu236: You're right. I tested it out and the patch isn't necessary. Don't know why it was added in first place.
Will remove it.

recipes-extended/lsb/lsbinitscripts/functions.patch
chaitu236: Only change with latest version is that any failures/warnings, etc would now be visible instead of being ignored like in current version. There are no other functional differences.
So I think we can go ahead with upgrading the recipe.

amstewart:
ACK; looks good to me.

PR #20, opened 2020-08-04
base-files-nilrt: add /etc/machine-info

sharpk:
This file is needed to store the system comment field that is visible in
MAX.  It needs to have proper permissions and group ownership to allow
System Web Server to modify its contents.

Signed-off-by: Ken Sharp <ken.sharp@ni.com>

### DIFF
--- a/recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb
+++ b/recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb
@@ -74,5 +74,5 @@ do_install () {
 	install -m 0644 ${WORKDIR}/sysctl.conf ${D}${sysconfdir}/
 
 	# add machine-info and allow System Web Server to modify it
-	install -m 0664 -g ni ${WORKDIR}/machine-info ${D}${sysconfdir}/
+	install -m 0664 -g ${LVRT_GROUP} ${WORKDIR}/machine-info ${D}${sysconfdir}/
 }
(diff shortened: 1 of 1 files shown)


sharpk:
I built this against nilrt/master/sumo and checked the resulting ipk to ensure that /etc/machine-info existed and had correct ownership and permissions.

recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb
```diff
@@ -69,4 +72,7 @@ do_install () {
 
 	# add sysctl.conf file to adjust system configuration parameters
 	install -m 0644 ${WORKDIR}/sysctl.conf ${D}${sysconfdir}/
+
+	# add machine-info and allow System Web Server to modify it
+	install -m 0664 -g ni ${WORKDIR}/machine-info ${D}${sysconfdir}/
```
amstewart: I think `LVRT_GROUP = "ni"` should be available as a variable via the [distro conf](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/conf/distro/nilrt.inc#L27). Is it available for you to use in the recipe here, instead of hard-coding the `ni` value?

amstewart (changes requested):
Looks good; pending the `LVRT_GROUP` suggestion below/above.

recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb
sharpk: Good call; I forgot about that var.

amstewart (approved):

amstewart:
Cherry-picked the same to `nilrt/master/dunfell` as https://github.com/ni/meta-nilrt/commit/243b1dfbd62e37bda94f6eb46052301bfdfaed41.

PR #21, opened 2020-08-05
Use kernel-devsrc to support building modules on target

chaitu236:
kernel-dev doesn't contain everything needed to build modules on target.
Notably, files needed for building tools/objtool aren't present preventing modules from getting built when elfutils-dev is installed.
kernel-devsrc already provides all files needed to build module on target but it can't be installed as there are conflicts
with kernel-dev.

The changes
1. Removes /lib/modules/<kernel>/build from kernel-dev.
2. Add kernel-devsrc to the distribution.
3. Add a postinst script to kernel-devsrc that prepares /lib/modules/<kernel>/build so external modules can be built.
4. Add kernel-devsrc as dependency to dkms.

### DIFF
--- a/recipes-core/packagegroups/packagegroup-kernel-module-build.bb
+++ b/recipes-core/packagegroups/packagegroup-kernel-module-build.bb
@@ -8,5 +8,6 @@ RDEPENDS_${PN} = "\
 	gcc \
 	gcc-symlinks \
 	kernel-dev \
+	kernel-devsrc \
 	make \
 	"
--- a/recipes-kernel/dkms/dkms_%.bbappend
+++ b/recipes-kernel/dkms/dkms_%.bbappend
@@ -1,0 +1,1 @@
+RDEPENDS_${PN} += "kernel-devsrc"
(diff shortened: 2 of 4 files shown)


chaitu236:
@amstewart @gratian

gratian:
Can you describe a bit the testing you've done so far for this change?

chaitu236:
This is how I tested

1. Built the changes on sumo.
2. Unpacked kernel-dev.ipk and verified that it doesn't contain _/lib/moduels/kernel/build_ and the postinst script.
3. On a VM provisioned with latest stack, removed _/lib/modules/kernel/build_ that was installed with existing kernel-dev and installed the new kernel-devsrc.ipk without issues.
4. Ran the [postinst commands](https://github.com/ni/meta-nilrt/pull/21/commits/f71192a52d46c519e3e22ea987a6e3f0b0736bd7#diff-083fb8fc63c4ded50efcb4039b0625ca) manually, installed elfutils-dev and built [this](https://github.com/cb22/macbook12-spi-driver) driver successfully. Without these changes the driver build used to fail due to objtool missing.
5. Installed some NI software from MAX to the target and verified their .ko driver files were present after reboot (which means dkms succeeded). modprobe'd a couple of these drivers without issues.

gratian (approved):
LGTM.

amstewart (approved):
Looks good; ship it.

PR #23, opened 2020-08-12
kernel-tests: Add kernel performance ptests

gratian:
Add cyclictest performance tests with various loads: idle, hackbench
scheduler load, fio based disk I/O load, and iperf based network load.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-performance-tests-files/common.cfg
+++ b/recipes-kernel/kernel-tests/kernel-performance-tests-files/common.cfg
@@ -4,3 +4,6 @@ TEST_DURATION=8h
 # Maximum expected latency in useconds.
 # The test will fail if cyclictest latency goes beyond this value.
 MAX_LATENCY=300
+
+# Location for benchmark logs and test results
+LOG_DIR="/var/local/ptest-results/kernel-performance-tests"
--- a/recipes-kernel/kernel-tests/kernel-performance-tests-files/common.cfg
+++ b/recipes-kernel/kernel-tests/kernel-performance-tests-files/common.cfg
@@ -1,0 +1,6 @@
+# Common configuration for all kernel performance tests
+TEST_DURATION=8h
+
+# Maximum expected latency in useconds.
+# The test will fail if cyclictest latency goes beyond this value.
+MAX_LATENCY=300
--- a/recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
+++ b/recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
@@ -1,0 +1,72 @@
+#!/bin/sh
+source ./common.cfg
+
+function add_system_info()
+{
+    {
+	DESC=$(fw_printenv DeviceDesc)
+	DEVICE=${DESC#*=}
+	echo "# Device: $DEVICE"
+
+	echo -n "# Kernel: "
+	uname -a
+
+	echo -n "# Kernel parameters: "
+	cat /proc/cmdline
+
+	# save info on the current security mitigations
+	echo "# Security vulnerabilities settings: "
+	if [ -d "/sys/devices/system/cpu/vulnerabilities" ]; then
+	    for f in  /sys/devices/system/cpu/vulnerabilities/*; do
+		VULN=$(basename "$f")
+		if [ -f "$f" ]; then
+		    echo -n -e "#     $VULN:\t"
+		    cat "$f"
+		fi
+	    done
+	else
+	    echo "#     vulnerability information not published"
+	fi
+
+	# save info on the current C state settings
+	echo "# C-states settings: "
+	for CPU in /sys/devices/system/cpu/cpu[0-9]; do
+	    NCPU=$(basename "$CPU")
+	    echo "#     [$NCPU]"
+	    for CSTATE in $CPU/cpuidle/state[0-9]; do
+		NSTATE=$(basename "$CSTATE")
+		if [ -f "$CSTATE/name" ] && [ -f "$CSTATE/disable" ]; then
+                    NAME=$(cat "$CSTATE/name" 2>/dev/null)
+                    STATUS=$(cat "$CSTATE/disable" 2>/dev/null)
+
+                    echo -n "#     $NSTATE: $NAME: "
+                    if [ "$STATUS" = "0" ]; then
+                        echo "enabled"
+                    else
+                        echo "disabled"
+                    fi
+		else
+                    echo "#     information not available"
+		fi
+	    done
+	done
+    } >> "$1"
+}
+
+function run_cyclictest()
+{
+    LOG="results/cyclictest-$1-`date +'%Y_%m_%d-%H_%M_%S'`.log"
+    cyclictest --smp --priority=98 --mlockall --notrace --interval=997 --quiet --duration="$TEST_DURATION" --histofall=1000 --histfile="$LOG" > /dev/null
+    add_system_info "$LOG"
+
+    LATENCY=$(grep -sw "# Max Latencies:" "$LOG" | awk '{max=$4; for(i=4; i<=NF; i++) if ($i>max) max=$i; gsub("^0*", "", max); print max}')
+    if [ "$LATENCY" -le "$MAX_LATENCY" ]; then
+	echo "PASS: cyclictest with $1 load, latency: $LATENCY (usec) is less than max latency: $MAX_LATENCY (usec)"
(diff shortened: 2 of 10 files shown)


gratian:
Attn: @bstreiff @amstewart

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
```diff
@@ -0,0 +1,72 @@
+#!/bin/sh
```
amstewart: Could you set the shebangs to `/bin/bash` only, since you're using non-POSIX syntax in most of the scripts.

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
```diff
@@ -0,0 +1,72 @@
...
+source ./common.cfg
+
+function add_system_info()
+{
+    {
+	DESC=$(fw_printenv DeviceDesc)
```
amstewart, line 7: Could you add a dependency on the `fw-printenv` package, which apparently [provides the utility](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-ni/ni-utils/fw-printenv_1.0.bb).

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
```diff
@@ -0,0 +1,72 @@
...
+                    echo "#     information not available"
+		fi
+	    done
+	done
+    } >> "$1"
+}
+
+function run_cyclictest()
+{
+    LOG="results/cyclictest-$1-`date +'%Y_%m_%d-%H_%M_%S'`.log"
```
amstewart: Can we change this to an absolute path under `/tmp` or something? We haven't enforced a consistent working dir from which to run the ptests.

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
```diff
@@ -0,0 +1,72 @@
...
+    cyclictest --smp --priority=98 --mlockall --notrace --interval=997 --quiet --duration="$TEST_DURATION" --histofall=1000 --histfile="$LOG" > /dev/null
+    add_system_info "$LOG"
+
+    LATENCY=$(grep -sw "# Max Latencies:" "$LOG" | awk '{max=$4; for(i=4; i<=NF; i++) if ($i>max) max=$i; gsub("^0*", "", max); print max}')
+    if [ "$LATENCY" -le "$MAX_LATENCY" ]; then
+	echo "PASS: cyclictest with $1 load, latency: $LATENCY (usec) is less than max latency: $MAX_LATENCY (usec)"
+	CYCLICTEST_RESULT=0
+    else
+	echo "FAIL: cyclictest with $1 load, latency: $LATENCY (usec) is above the expected max latency: $MAX_LATENCY (usec)"
```
amstewart: For ptest parsing, the contents of the results line needs to be consistent between runs (with the exception of the actual PASS/FAIL substring). I cover that in more detail [here](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/3560/Yocto-PTests?anchor=test-result-lines). You can print the load and latency info on lines before the PASS/FAIL line, though.

recipes-kernel/kernel-tests/kernel-performance-tests-files/common.cfg
```diff
@@ -0,0 +1,6 @@
+# Common configuration for all kernel performance tests
+TEST_DURATION=8h
```
amstewart, line 2: I suppose this isn't critical for what we're intending to use this ptest suite for, but it's worth noting that the `ptest-runner` wrapper binary is currently hardcoded to ETIMEOUT a test suite [if it runs for more than 300 seconds](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/3560/Yocto-PTests?anchor=timeliness). So we'll need to invoke this test manually in any testsuite we use it in (and/or upstream a patch to `ptest-runner` to allow for modifying that timeout).

recipes-kernel/kernel-tests/kernel-performance-tests-files/test_kernel_cyclictest_fio.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+source ./run-cyclictest
+
+# start background disk I/O load
+mkdir -p .fio
+fio fio.cfg --ioengine="sync" --runtime="$TEST_DURATION" --ramp_time=1m --output="results/fio-sync-`date +'%Y_%m_%d-%H_%M_%S'`.log" > /dev/null &
```
amstewart: Same comment about having a more absolute location to store the `.fio` directory.

recipes-kernel/kernel-tests/kernel-performance-tests-files/test_kernel_cyclictest_iperf.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+source ./run-cyclictest
+
+# start background network load
+source ./iperf.cfg
+if [ -z "$IPERF_SERVER" ]; then
+    echo "SKIP: iperf server not configured, skipping iperf based network load test."
```
amstewart: Same comment about the result line needing to be consistent.

amstewart (changes requested):

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
gratian: I don't like using /tmp for this because it's volatile and limited in space, but I can look for a different place if desired.

IMHO having the results stored where the tests reside (/usr/lib/kernel-performance-tests/ptest/results/) is not the worst thing in the world. It makes them easier to retrieve since for these particular tests the log files are the important part not the pass/fail print.

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
gratian: OK

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
gratian: Sounds good. Thanks for catching this.

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
gratian: Sure; I didn't realize there was convention doc for this. That's good to know. The Yocto manual doesn't mention anything beyond the PASS/FAIL substring.

recipes-kernel/kernel-tests/kernel-performance-tests-files/test_kernel_cyclictest_fio.sh
gratian: /tmp is not going to have enough space for this. The fio benchmark creates large files during the test. The folder is removed at the end of the test so it shouldn't matter but I can look for some other location in /var or something.

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
amstewart: I think the test location would be fine. Another possibility might be to store them under $HOME for the running user. I was only objecting to using a blind, relative path - since who knows where the shell is when these tests are run.

recipes-kernel/kernel-tests/kernel-performance-tests-files/run-cyclictest
amstewart: Yep. Believe me, that wiki page is written in blood. :)

recipes-kernel/kernel-tests/kernel-performance-tests-files/common.cfg
gratian: ptest-runner already has that option ('-t'). I'm currently currently doing a run using that option i.e. 'ptest-runner -t 360000' which should give me enough runway for these tests before it times out.

amstewart: Ah! Nevermind; I have old information then.

gratian:
Requested changes have been implemented. See fixup commit.

amstewart (approved):

PR #26, opened 2020-08-22
Fix warnings in Dunfell OE build.

shruthi-ravi:
There are still a few warnings left that I haven't figured out. Pushing what I have.
@ni/rtos


recipes-connectivity/openssl/openssl_1.0.2o.bbappend
```diff
@@ -1,31 +0,0 @@
-FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
-
-SRC_URI_remove = "file://debian/pic.patch"
```
amstewart, line 3: Do you know the rationale for dropping this OE file? Is this a change we need to carry forward into a new openssl `.bbappend`?

recipes-devtools/python/python_%.bbappend
```diff
@@ -1,5 +0,0 @@
-# we don't want python 2 to take precedence over the default python 3 because
-# it breaks salt if it gets installed from the feeds alongside python 3
-do_install_append () {
-	rm -f ${D}${bindir}/python
```
amstewart, line 4: This is fine.

Have you verified that `/usr/bin/python` evaluates to python3 on the images? The python links in OE are kind of odd and I wouldn't be surprise if a relatively simple change like this messes something up.

recipes-core/udev/udev/0001-ni-skip-kmod-for-wl12xx.patch
```diff
@@ -4,7 +4,7 @@ index 8551f47..ac8bb60 100644
 +++ b/rules/80-drivers.rules
 @@ -2,6 +2,9 @@
  
- ACTION=="remove", GOTO="drivers_end"
+ ACTION!="add", GOTO="drivers_end"
  
 +# skip kmod for wl12xx device to prevent 30s delay issues with 3.2 kernel
 +ENV{MODALIAS}=="platform:wl12xx", GOTO="drivers_end"
```
amstewart: I think this patch can be dropped, in its entirety, from the Dunfell release.

Also, you can close [this work item](https://dev.azure.com/ni/DevCentral/_workitems/edit/1066354) while you're at it. :)

recipes-core/coreutils/coreutils_8.%.bbappend
```diff
@@ -11,14 +11,16 @@ DEPENDS += "shadow-native pseudo-native niacctbase"
 RDEPENDS_${PN}-hostname += "niacctbase"
 
 # preserve existing behavior when installing base package
-RDEPENDS_${PN}_class-target += "\
```
amstewart, line 14: I think the way you're supposed to fix this error is by using something like
```bash
RDEPENDS_${PN}_class-target_append += ...
```
... or something like that. I might be mistaken, but I would like to preserve the `${PN}` semantic, if we can.

amstewart (changes requested):
Looks good overall. It's nice to see that we're able to drop so many `.patch`es!

recipes-connectivity/openssl/openssl_1.0.2o.bbappend
shruthi-ravi: Good question. The answer is likely? [This](https://github.com/ni/meta-nilrt/commit/46a320ade106e0744b1f42f244187448d1048188#diff-d12441f3749633acb88494958beb3909) is the commit that added the lines. [This](https://review-board.natinst.com/r/128139/) is the rb link describing the changes a little more. I guess NI-SSL is still a thing on Windows?

amstewart: I think so. We should probably keep this part of the .bbappend.

recipes-devtools/python/python_%.bbappend
shruthi-ravi: Verified that /usr/bin/python evaluates to python3. Python 3.8.

recipes-core/coreutils/coreutils_8.%.bbappend
shruthi-ravi: Tried that, but it didn't work. Take a look at the final comment from Richard Purdie in this [thread](https://patchwork.openembedded.org/patch/98957/) that might explain why.

amstewart: Ah; that's slightly annoying. Could you add a section to the bitbake wiki document with a link to that thread. I imagine this will come up many more times. :/

recipes-connectivity/openssl/openssl_1.0.2o.bbappend
shruthi-ravi: Kept the following line
`SRC_URI_remove = "file://debian/pic.patch"`
This patch is not in our files directory. So we wouldn't need this definition right?
`FILESEXTRAPATHS_prepend := "${THISDIR}/files:"`

recipes-core/udev/udev/0001-ni-skip-kmod-for-wl12xx.patch
shruthi-ravi: Done.

amstewart (changes requested):
> Though is old/deprecated/unmaintained

s/is/it is/
:)

recipes-connectivity/openssl/openssl_1.1.1g.bbappend
```diff
@@ -0,0 +1 @@
+SRC_URI_remove = "file://debian/pic.patch"
```
amstewart, line 8: Could you add a comment in this `.bbappend` which explains why we're removing this `.patch`. i think that would be easier than having our future selves dig back through the git log to rediscover the rationale.

amstewart (changes requested):

recipes-connectivity/openssl/openssl_1.0.2o.bbappend
amstewart: Correct; we don't need that `FILESEXTRAPATHS` assignment.

amstewart:
Looks good; pending the nits above.

recipes-connectivity/openssl/openssl_1.1.1g.bbappend
shruthi-ravi: Done.

shruthi-ravi:
I had updated the iproute2 version to 2.5 in another commit. I need to rebase this branch on top of those commits, which probably means another round of force pushes.

amstewart (approved):

PR #27, opened 2020-08-24
update salt recipe to point to 20.5 release

andzn:


chaitu236:
LGTM

shruthi-ravi:
This looks good, but I went back to look at previous release branches and nilrt/20.1/sumo is still at the master salt branch, auto-revving. Can that be updated separately?

amstewart:
> This looks good, but I went back to look at previous release branches and nilrt/20.1/sumo is still at the master salt branch, auto-revving. Can that be updated separately?

Since the 20.1 releases were never published, there was no reason to staple them to a particular commit. That's probably not something I would advocate that we do in future cycles, it it isn't harming anything on our end to have those branches stay on autorev.

PR #28, opened 2020-08-26
Add support for installing the 'next' kernel from package feeds

gratian:
This patch series adds support for building and installing a 'next' kernel on a NILRT distro via package feeds.

New kernels are installed as versioned image files (e.g. /boot/runmode/bzImage-5.6.17-rt9-next) and switching to the new kernel is accomplished by changing the '/boot/runmode/bzImage' symlink. A list of available kernel images is generated and kept for uninstall use cases. This patch series does not cover updating the initramfs images on kernel installs from feed.

Testing done:

- built packages for the 'next' kernel;

- verified that: 'opkg install packagegroup-ni-next-kernel' and 'opkg remove --autoremove packagegroup-ni-next-kernel' work correctly;

- verified that a full NI OE build succeeds and the output can be consumed w/o errors by the os-common component; 

- verified that the layout of the exported images has not changed.

### DIFF
--- a/recipes-core/images/nilrt-image-common.inc
+++ b/recipes-core/images/nilrt-image-common.inc
@@ -11,12 +11,9 @@ IMAGE_PREPROCESS_COMMAND += "rootfs_update_timestamp;"
 # Do not install license subpackages if they are only recommended.
 BAD_RECOMMENDATIONS_pn-${PN} += "*-lic"
 
-ROOTFS_POSTPROCESS_COMMAND += "\
-       ${@oe.utils.conditional('DISTRO', 'nilrt-nxg', '', 'move_kernel;', d)} \
-"
-
-# Older NILRT distro flavors expect the kernel is an actual image file
+# NILRT images expect the default kernel is an actual file
 # (e.g. bzImage) and not a symbolic link; fix it up here.
+ROOTFS_POSTPROCESS_COMMAND += "move_kernel;"
 move_kernel() {
 	for type in ${KERNEL_IMAGETYPES} ; do
 		if [ -L "${IMAGE_ROOTFS}/${KERNEL_IMAGEDEST}/${type}" ]; then
--- a/recipes-core/images/nilrt-image-common.inc
+++ b/recipes-core/images/nilrt-image-common.inc
@@ -11,4 +11,19 @@ IMAGE_PREPROCESS_COMMAND += "rootfs_update_timestamp;"
 # Do not install license subpackages if they are only recommended.
 BAD_RECOMMENDATIONS_pn-${PN} += "*-lic"
 
+ROOTFS_POSTPROCESS_COMMAND += "\
+       ${@oe.utils.conditional('DISTRO', 'nilrt-nxg', '', 'move_kernel;', d)} \
+"
+
+# Older NILRT distro flavors expect the kernel is an actual image file
+# (e.g. bzImage) and not a symbolic link; fix it up here.
+move_kernel() {
+	for type in ${KERNEL_IMAGETYPES} ; do
+		if [ -L "${IMAGE_ROOTFS}/${KERNEL_IMAGEDEST}/${type}" ]; then
+			mv "${IMAGE_ROOTFS}/${KERNEL_IMAGEDEST}/$(readlink "${IMAGE_ROOTFS}/${KERNEL_IMAGEDEST}/${type}")" \
+			   "${IMAGE_ROOTFS}/${KERNEL_IMAGEDEST}/${type}"
+		fi
+	done
+}
+
 inherit core-image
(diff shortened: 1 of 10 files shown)


gratian:
Attn: @chaitu236 @bstreiff @jeminor

recipes-kernel/linux/linux-nilrt-next_git.bb
```diff
@@ -0,0 +1,14 @@
+DESCRIPTION = "NILRT linux kernel next development build"
+
+require linux-nilrt-next.inc
+
+LINUX_VERSION_EXTENSION = "-next"
+KERNEL_PACKAGE_NAME = "kernel${LINUX_VERSION_EXTENSION}"
```
amstewart: Is there a reason that we're adding `-next` to the kernel version? I'm fine with adding it to the package name (`kernel-next`), but this will apply to the files on disk as well, right? That seems excessive.

recipes-kernel/linux/linux-nilrt.inc
```diff
@@ -78,16 +78,118 @@ do_install_append() {
...
+	fi
+
+	# Also make sure the kernel being uninstalled is listed in
+	# 'kernel-versions'. This simplifies error checking later.
+	if ! grep -xqs "${KERNEL_IMAGETYPE}-${KERNEL_VERSION}" kernel-versions; then
+		echo "${KERNEL_IMAGETYPE}-${KERNEL_VERSION}" >> kernel-versions
+	fi
+
+	# Sanity check the kernels listed in 'kernel-versions' are
+        # present on disk.
```
amstewart: Whitespace. :)

recipes-kernel/linux/linux-nilrt.inc
```diff
@@ -78,16 +78,118 @@ do_install_append() {
...
+	while read line; do
+		if [ -f "$line" ]; then
+			echo "$line"
+		fi
+	done <kernel-versions >kernel-versions.tmp
+	mv -f kernel-versions.tmp kernel-versions
```
amstewart: What do you think about throwing a `WARNING` if we find a kernel declaration which doesn't exist? Seems like a sysadmin might want to know that the `kernel-versions` file fell out of date at some point.

recipes-core/images/nilrt-image-common.inc
```diff
@@ -11,4 +11,19 @@ IMAGE_PREPROCESS_COMMAND += "rootfs_update_timestamp;"
 # Do not install license subpackages if they are only recommended.
 BAD_RECOMMENDATIONS_pn-${PN} += "*-lic"
 
+ROOTFS_POSTPROCESS_COMMAND += "\
+       ${@oe.utils.conditional('DISTRO', 'nilrt-nxg', '', 'move_kernel;', d)} \
+"
```
amstewart: The `nilrt-nxg` DISTROs are all deprecated at this point. We can avoid this conditional, I think.

recipes-core/images/nilrt-image-common.inc
```diff
@@ -11,4 +11,19 @@ IMAGE_PREPROCESS_COMMAND += "rootfs_update_timestamp;"
...
+
+# Older NILRT distro flavors expect the kernel is an actual image file
+# (e.g. bzImage) and not a symbolic link; fix it up here.
+move_kernel() {
```
amstewart: When we're saying "older NILRT distro flavors" here, we're talking about pre-RAUC images; right? All of the OneRT images are going to use the `nilrt` DISTRO value.

amstewart (changes requested):

recipes-kernel/linux/linux-nilrt-next_git.bb
gratian: I could make a weak argument that it would be good to know from the files on disk and the bzImage symlink if you are running a 'next' kernel but yeah, it's a bit redundant so I'll take it out.

recipes-core/images/nilrt-image-common.inc
gratian: Hmm, maybe I should remove the comment altogether since RAUC won't be able to use symlinks in the image either (due to UEFI/FAT32 partition limitation). I'll try to re-word it.

recipes-kernel/linux/linux-nilrt-current.inc
```diff
@@ -0,0 +1,6 @@
+NI_RELEASE_VERSION = "master"
+LINUX_VERSION = "4.14"
+KBRANCH = "nilrt/${NI_RELEASE_VERSION}/${LINUX_VERSION}"
+GIT_KERNEL_REPO = "linux.git"
+
+require linux-nilrt.inc
```
chaitu236: Should there be a newline at the end?

gratian: I don't think it's needed. Do you know of places this would cause an issue?

amstewart (approved):

recipes-kernel/linux/linux-nilrt-current.inc
chaitu236: I don't think there'll be any issues - just mentioned as its a convention on linux machines to end files with newline. I don't mind you dropping this though.

gratian: OK, fair point. I pushed a fixup commit just for consistency.

PR #30, opened 2020-09-04
packagegroup-ni-extra: Fix enchant,usbpath,python

shruthi-ravi:
Update enchant to enchant-2.
Remove usbpath dependency because the recipe was removed from
meta-openembedded.
Update python2 recipes to python3.

This should fix the [errors](https://heisenberg.amer.corp.natinst.com/view/nilrt-CG/view/CG%2021.0/job/NIOE_Dunfell_x64_temp_Export-9.0/5/consoleText) when building the feed.

@ni/rtos

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-extra.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-extra.bb
@@ -135,15 +135,15 @@ RDEPENDS_${PN} += "\
 	patchelf \
 	pax-utils \
 	prelink \
-	python-dbus \
-	python-distribute \
-	python-imaging \
-	python-setuptools \
+	python3-dbus \
+	python3-distribute \
+	python3-imaging \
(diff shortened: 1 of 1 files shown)


recipes-core/packagegroups/packagegroup-ni-extra.bb
```diff
@@ -601,54 +600,54 @@ RDEPENDS_${PN} += "\
...
+	python3-pytz \
+	python3-zopeinterface \
+	python3-pyparted \
+	python3-pyconnman \
 	python3-gevent \
 	python3-greenlet \
 	python3-gsocketpool \
 "
-# python-matplotlib: hang on do_install
+# python3-matplotlib: hang on do_install
```
amstewart, line 650: Does this still hang on do_install?

amstewart (approved):

PR #31, opened 2020-09-08
Fix issue where update did not work for uncompressed feeds

andzn:
There was a case in which opkg update did not work for uncompressed feeds. 
Steps to reproduce are:
1. add an unexisting compressed feed URL
2. add an existing uncompressed feed URL (they have to be in this order and they have to be URLs so that the opkg download curl backend is used)
3. make sure volatile cache is enabled
4. run opkg update
5. try to install something fom the uncompressed feed

Although opkg says that the uncompressed feed was updated, the uncompressed feed lists file is not downloaded and therefore no packages from that feed can be installed.

The issue was that after an unexisting compressed feed download failure, the opkg options were not reset and therefore a subsequent uncompressed feed download attempt failed.

**Testing**
- built recipe
- installed generated libopkg ipk on a x64 target where the issue was reproducing. Checked that after upgrading libopkg the issue doesn't reproduce anymore.

### DIFF
--- a/recipes-devtools/opkg/files/0002-fix-issue-where-update-did-not-work-for-uncompressed.patch
+++ b/recipes-devtools/opkg/files/0002-fix-issue-where-update-did-not-work-for-uncompressed.patch
@@ -1,0 +1,26 @@
+From bab8d6f1e74d52c53bb0600bbe5afd8943b65bce Mon Sep 17 00:00:00 2001
+From: Andrei Zene <andrei.zene@ni.com>
+Date: Tue, 8 Sep 2020 22:07:25 +0300
+Subject: [PATCH] Fix issue where update did not work for uncompressed feeds
+
+---
+ libopkg/opkg_download_curl.c | 6 ++++++
+ 1 file changed, 6 insertions(+)
+
+diff --git a/libopkg/opkg_download_curl.c b/libopkg/opkg_download_curl.c
+index d572070..8d6ca6a 100644
+--- a/libopkg/opkg_download_curl.c
++++ b/libopkg/opkg_download_curl.c
+@@ -245,6 +245,12 @@ static int opkg_validate_cached_file(const char *src, const char *cache_location
+         curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &error_code);
+         opkg_msg(ERROR, "Failed to download %s headers: %s.\n", src,
+                  curl_easy_strerror(res));
++        free(etag);
++        curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, NULL);
++        curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, NULL);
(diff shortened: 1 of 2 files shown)


amstewart (commented):
Content-wise, this patch seems reasonable to me. Thanks for fixing it!

I think I found the AZDO item for this issue (#1122119?) and it seems like this isn't blocking SL dev or testing and that it should go into the 20.6 release. If that is true, could you send the opkg .patch upstream to see if anyone on the opkg ML has feedback. It's always better to pull the same `.patch` as is already accepted upstream, in case the content changes. If it lingers upstream (unlikely), I can pull it prior to the 20.6 go-final in a pending state.

**A little commit feedback:** 
* for `.patch` recipe additions like this, it's preferable to have a single commit which both adds the new `.patch` file and the `SRC_URI` entry - since pulling one without the other isn't a sane recipe change.
* could you prefix the commit summary with the recipe name. eg. `opkg: add patch fix for uncompressed feed downloads`.

Lastly, when we do the actual PR for sumo (after you've gotten feedback from opkg upstream), you'll also need to rebase the patchset onto `nilrt/master/dunfell`, so that we can pull it into the 21.0 pipeline.

amstewart:
Oh, also: in the work item comments, you mentioned.
> I've managed to write a test for opkg that reproduces the issue.

Is that test a script you hacked up or is it something that you can add to the existing `:tests/` framework in opkg? If the latter, I'm sure upstream would appreciate having it added to the regression test suite.

andzn:
> Oh, also: in the work item comments, you mentioned.
> 
> > I've managed to write a test for opkg that reproduces the issue.
> 
> Is that test a script you hacked up or is it something that you can add to the existing `:tests/` framework in opkg? If the latter, I'm sure upstream would appreciate having it added to the regression test suite.

It was a test that I added to the existing `:tests/` framework in opkg (I actually started with it in order to understand when the issue reproduces). However there's a little problem with it. All the opkg tests currently use `file:` feeds and therefore do not exercise the curl download backend. For this issue, however, as stated in the steps to reproduce, the test would need to access an existing feed over http(s). My test was using a feed from download.ni.com, but it felt inappropriate to make the test suite depend on internet connection. Of course this can also be solved by starting a local HTTP server but that's a lot of infrastructure and it felt like it's not worth. This is why I didn't add the test in the patch file anymore.

amstewart:
Coincidentally, I happen to be working on an upgraded test framework for opkg which will start a local feed server. If you push your test to a dev branch somewhere, I'll see if I can work it into my project.

amstewart:
I'm going to go ahead and close PR, pending the .patch file going upstream.

andzn:
> Coincidentally, I happen to be working on an upgraded test framework for opkg which will start a local feed server. If you push your test to a dev branch somewhere, I'll see if I can work it into my project.

I've created this branch that also contains the test: https://github.com/andzn/meta-nilrt/tree/fix-opkg-update-issue-2

PR #34, opened 2020-09-16
opkg: add patch fix for uncompressed feed downloads

andzn:
The opkg_validate_cached_file method sets a dummy write function on
the global curl session pointer that is used by opkg. It also sets
CURL_NOBODY to 1 which will only download the header. It does this
to check if the timestamp of the source has changed and if a
redownload is needed.

However, if that curl call for the header fails, it doesn't reset
back the curl options. Because of this, if a subsequent call to
opkg_download doesn't use cache (which is the case for uncompressed
feeds list files) and therefore opkg_validate_cache is not called at
all, the download will look as if it succeeded where in fact it
didn't write anything to the disk.

The opkg_validate_cached_file function already had some code to
cleanup the curl options so that they are in the same state then
before the call. However, that cleanup logic was not executed in
case of an error. This change makes the cleanup logic to be executed
also on error.

This fixes an issue where an opkg update would look as if it
successfully updated an uncompressed feed source but it didn't.
The issue occured only under the following conditions:
- before updating the uncompressed feed source, it tried to
update a compressed feed source and it failed
- volatile_cache is enabled.

Note: The patch was accepted upstream: https://git.yoctoproject.org/cgit/cgit.cgi/opkg/commit/?id=0485ad453fc5f582e4d289cb35b0c70dc17f7544

Signed-off-by: Andrei Zene <andrei.zene@ni.com>

### DIFF
--- a/recipes-devtools/opkg/files/0002-libopkg-clear-curl-properties-on-download-error-to-p.patch
+++ b/recipes-devtools/opkg/files/0002-libopkg-clear-curl-properties-on-download-error-to-p.patch
@@ -1,0 +1,110 @@
+From 8ad7c8bdc8a77f357a6ee929184647a98c5b69cb Mon Sep 17 00:00:00 2001
+From: Andrei Zene <andrei.zene@ni.com>
+Date: Fri, 11 Sep 2020 23:20:34 +0300
+Subject: [PATCH] libopkg: clear curl properties on download error to prevent
+ polluting next call
+To: opkg-devel@googlegroups.com
+
+The opkg_validate_cached_file method sets a dummy write function on
+the global curl session pointer that is used by opkg. It also sets
+CURL_NOBODY to 1 which will only download the header. It does this
+to check if the timestamp of the source has changed and if a
+redownload is needed.
+
+However, if that curl call for the header fails, it doesn't reset
+back the curl options. Because of this, if a subsequent call to
+opkg_download doesn't use cache (which is the case for uncompressed
+feeds list files) and therefore opkg_validate_cache is not called at
+all, the download will look as if it succeeded where in fact it
+didn't write anything to the disk.
+
+The opkg_validate_cached_file function already had some code to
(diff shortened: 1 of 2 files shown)


recipes-devtools/opkg/files/0002-libopkg-clear-curl-properties-on-download-error-to-p.patch
```diff
@@ -0,0 +1,108 @@
...
+
+This fixes an issue where an opkg update would look as if it
+successfully updated an uncompressed feed source but it didn't.
+The issue occured only under the following conditions:
+- before updating the uncompressed feed source, it tried to
+update a compressed feed source and it failed
+- volatile_cache is enabled.
+
+Signed-off-by: Andrei Zene <andrei.zene@ni.com>
+---
```
amstewart: Could you add an `Upstream-Status: Accepted ...` trailer here ([like this](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-support/curl/curl/0007-Added-curl_multi_setopt_va-API-function.patch#L9)), which will help when we next need to rebase this recipe.

amstewart (changes requested):
Thanks for getting this upstreamed. Everything looks good to me, pending the trailer suggestion below.

recipes-devtools/opkg/files/0002-libopkg-clear-curl-properties-on-download-error-to-p.patch
andzn: Done.

amstewart:
@andzn Could you rebase your `.patch` and submit a similar PR to the `nilrt/master/dunfell` ref - which will become the 21.X releases?

andzn:
> @andzn Could you rebase your `.patch` and submit a similar PR to the `nilrt/master/dunfell` ref - which will become the 21.X releases?

Sure: https://github.com/ni/meta-nilrt/pull/35

PR #35, opened 2020-09-17
opkg: add patch fix for uncompressed feed downloads

andzn:
Cherry-picking #34 to nilrt/master/dunfell

Signed-off-by: Andrei Zene <andrei.zene@ni.com>


amstewart:
Did you do a `bitbake opkg` build and verify that the .patch applies cleanly on the newer opkg sources?

andzn:
> Did you do a `bitbake opkg` build and verify that the .patch applies cleanly on the newer opkg sources?

I can do that but nilrt/master/sumo and nilrt/master/dunfell use the same opkg sources. Besides, the opkg patch file is based on the latest opkg source (opkg master) and since it applied succesfully on nilrt/master/sumo, I am pretty sure it will apply just fine on dunfell too.

amstewart:
Ah; I had forgotten that we cherry-picked the 0.4.2 opkg recipes back to sumo. I agree that no additional testing is necessary.

Thanks.

PR #36, opened 2020-09-24
packagegroup-ni-extra: add makedumpfile

bstreiff:
This package is used in conjunction with a dump-capture kernel for kdump.


bstreiff:
fyi, this pull request is for dunfell, [37](https://github.com/ni/meta-nilrt/pull/37) is for sumo, this seems like a change that may as well go into both feeds.

PR #38, opened 2020-09-24
rauc image prep

amstewart:
Make a pass through the `conf/` and `images/` directories and strip out deprecated nilrt-nxg and `MACHINE=ARM` content, in preparation for the new RAUC images.

The new bblayers and local.conf files are a merge of the old (CG) content and the better ideas from the NXG content.

The `NILRT_ROOT` variable comes from [this nilrt.git PR](https://github.com/ni/nilrt/pull/22).

@ni/rtos


amstewart:
PATCH V2:
* Remove the `images/` changes, because those are becoming large enough that they will be submitted as their own PR.

conf/bblayers.conf.sample
```diff
@@ -1,40 +1,40 @@
+# LAYER_CONF_VERSION is increased each time build/conf/bblayers.conf
```
shruthi-ravi, line 1: Update name of the variable referenced to LCONF_VERSION.

conf/distro/nilrt.inc
```diff
@@ -6,23 +6,36 @@ GRUB_BRANCH = "nilrt/19.0"
...
+	zeroconf \
+	pci \
+	sysvinit \
+	pam \
+	ptest \
+	selinux \
 	virtualization \
 "
 
+# Clean *most* package work directories
```
shruthi-ravi, line 26: Can you comment on why you are excluding the directories below?

conf/local.conf.sample
```diff
@@ -230,6 +220,41 @@ BB_DISKMON_DIRS ??= "\
...
+# uid/gid.  This can lead to non-deterministic uid/gid values from one build
+# to another.  Use the following settings to specify that all user/group adds
+# should be created based on a static passwd/group file.
+#
+# Note, if you enable or disable the useradd-staticids in a configured system,
+# the TMPDIR may contain incorrect uid/gid values.  Clearing the TMPDIR
+# will correct this condition.
+#
+# By default the system looks in the BBPATH for files/passwd and files/group
+# the default can be overriden by spefying USERADD_UID/GID_TABLES.
```
shruthi-ravi: *specifying

conf/local.conf.sample
```diff
@@ -230,6 +220,41 @@ BB_DISKMON_DIRS ??= "\
...
+# Enable internal source mirroring. Set SOURCE_MIRROR_URL externally.
+INHERIT += "own-mirrors"
+BB_GENERATE_MIRROR_TARBALLS = "1"
+
+# The network based PR service host and port
+# Uncomment the following lines to enable PRservice.
+# Set PRSERV_HOST to 'localhost:0' to automatically
+# start local PRService.
+# Set to other values to use remote PRService.
+PRSERV_HOST = "versionator.amer.corp.natinst.com:8585"
```
shruthi-ravi, line 255: Can you explain this config? Should it commented since the comment above says, "Uncomment...".

conf/local.conf.sample
```diff
@@ -38,20 +38,11 @@ MACHINE ??= "x64"
```
shruthi-ravi, line 38: I've never understood, can you explain the purpose of these .sample files? Are they just suggested base .conf files that anyone can start from?

conf/bblayers.conf.sample
amstewart: I'm not sure what that variable actually does. I don't see any upstream documentation about it, nor do I see a place that we reference it in our pipelines. Is there something I'm missing here?

I can bump the version number, since this is a breaking change. I'm not sure which variable name is appropriate though.

conf/local.conf.sample
amstewart: When you source the initialization file in the nilrt.git repo, it will create your build directory if it doesn't already exist. If that build directory doesn't contain a `:conf/bblayers.conf` or `:conf/local.conf` file, they will be copied from the directory you [define](https://github.com/ni/nilrt/blob/nilrt/master/sumo/ni-oe-init-build-env#L1) in your environment's `TEMPLATECONF` variable - which is meta-nilrt, in our case.

The `.sample` files are the template that is copied. So, in a sense, these are the *default* values for those local files.

conf/local.conf.sample
amstewart: Fixed. :)

conf/distro/nilrt.inc
amstewart: When recipes are built, bitbake creates a subpackages directory in their workspace, into which it builds the package trees before compressing them into IPKs. Bitbake otherwise doesn't have a good mechanism for determining what packages *actually* resulted from a recipe. Event the `PACKAGES_$PN` variable only tells you what the superset of possible packages are, and once the IPKs are copied into the `deploy/` directory pool, the association with their parent recipe is lost.

So the recipes added to this variable, and the similar variable in local.conf, were added at one time or another because we had some workflow which needed to interrogate that temporary subpackages workspace for the IPKS owned by the recipe.

For this patchset, I've just set these variables to be a superset of what was defined in each of the CG and NXG versions of the pipeline .conf files - for maximum compatibility. In my later patchsets to remove the ARM and NXG components, I will remove entries from the variable which aren't need anymore (which might be all of them).

conf/local.conf.sample
amstewart: I think the comment is a leftover from the example version of this file in OE upstream (ie. the line was originally commented-out). This variable defines the location of bitbake's Patch-Rev server, which bitbake will interrogate to figure out what the PR value should be for new recipe builds. Within the NI network, versionator will resolve to a sqlite database that we host on the gustavo hardware.

I don't think it hurts anything to have this variable defined; even outside of the NI network. If it does, we can add a toggle for it to the build environment init.

shruthi-ravi: Can you remove this line then?
"# Uncomment the following lines to enable PRservice."

conf/bblayers.conf.sample
shruthi-ravi: I am similarly confused. There are lots of references to other people being confused about the same in the yocto mailing list, but I don't see a definitive answer. I guess just bump the version number for now.

shruthi-ravi (approved):
Looks good, pending responses on the outstanding comments.

amstewart:
Patch V4:
bblayers.conf.sample:
* Bumped bblayers conf version to 6 to reflect the breaking changes in this patchset.
* Also changed the soft tabs to hard tabs because tabs are better for indentation.

local.conf.sample:
* Removed the confusing line about uncommenting from the documentation comment on the PRSERV variable.

conf/bblayers.conf.sample
amstewart: Done.

conf/local.conf.sample
amstewart: Done.

PR #41, opened 2020-10-12
bblayers.conf.sample: fixup LCONF_VERSION

amstewart:
The LCONF_VERSION should be '7', to align to the minimum required
version for upstream OE/dunfell's sanity checker.

Ref oe-core commit 933088c74870b8c3aa5077d57bc85fc66a652291.
https://github.com/ni/openembedded-core/commit/933088c74870b8c3aa5077d57bc85fc66a652291

### DIFF
--- a/conf/bblayers.conf.sample
+++ b/conf/bblayers.conf.sample
@@ -1,6 +1,6 @@
-# LAYER_CONF_VERSION is increased each time build/conf/bblayers.conf
-# changes incompatibly
-LCONF_VERSION = "6"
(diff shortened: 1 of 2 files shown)


shruthi-ravi:
Can you check if some recipe builds as expected? You'll see an error right away if not.

amstewart:
> Can you check if some recipe builds as expected? You'll see an error right away if not.

Yeah; I see the issue now. I've updated the patchset to also bump the LAYER_CONF_VERSION to 7, which resolves the sanity checker.

shruthi-ravi (approved):
Looks good now!

PR #42, opened 2020-10-12
RAUC images rearchitecture (main patchset)

amstewart:
This patchset aims to:
* remove the deprecated ARM image content from recipes-core/images
* remove the deprecated x64 NIDKMS (aka. non-DKMS) images
* remove the deprecated safemode image content
* remove the deprecated NXG LVComms images
* add x64 RAUC images in the form of the `nilrt-base-*` family of content.

The x64 DKMS images are kept around to keep internal component exports compatible, until the RAUC content is ready for mainstream consumption.

I didn't dramatically alter the contents of these images, beyond refactoring names and connecting content to where it needs to go.

The `images:README` diff is a good overview of what has been changed. The other recipe changes are mostly incidental cleanup to affect the image recipe changes.

@ni/rtos


amstewart:
NI Internal note: there are some implications for the NIOE and os-common components as a result of this patchset. We will need to deprecate the nikms and safemode contents in both, and the recovery media exports in os-common will not boot until we finish integrating changes that @shruthi-ravi is working on now.

recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -374,16 +374,12 @@ elif [ "$ARCH" = "x86_64" ]; then
             mount -o rw /dev/niboot/niboota /mnt/efi-auto/niboota
 
             local img_written=false
-            for imgname in "minimal-nilrt-bundle-image-x64.tar.bz2" "lvcomms-nilrt-bundle-image-x64.tar.bz2"; do
-                local imgpath="/mnt/efi-auto/bundle/$imgname"
+            for imgpath in /mnt/efi-auto/bundle/*.tar.bz2; do
```
shruthi-ravi, line 377: Can there be more than one image in this location? Why do we need to loop?

recipes-core/images/nilrt-base-image.bb
```diff
@@ -0,0 +1,28 @@
+DESCRIPTION = "NI LinuxRT Base System Image for x64 Targets"
```
shruthi-ravi, line 1: Mention RAUC here? It can get confusing since we are keeping around the non-RAUC image too for now.

recipes-core/images/nilrt-restore-mode-image.bb
```diff
@@ -1,7 +1,7 @@
 DESCRIPTION = "Tiny initramfs image intended to run restore mode operations, uses minimal-nilrt-image"
```
shruthi-ravi: Update the reference to "minimal-nilrt-image".

recipes-core/images/nilrt-restore-mode-image.bb
```diff
@@ -1,7 +1,7 @@
...
-NIBOOT_BUNDLE_IMAGE = "minimal-nilrt-bundle"
-IMAGE_DISPLAY_NAME = "minimal runmode"
+NIBOOT_BUNDLE_IMAGE = "nilrt-base-bundle"
```
shruthi-ravi, line 2: This image doesn't look minimal to me. Are you sure it is a good replacement for the image from minimal-nilrt-bundle?

recipes-core/initrdscripts/files/init-nilrt-ramfs.sh
```diff
@@ -1,8 +1,4 @@
 #!/bin/bash
-
-# Any failure at any line of this script will cause the script to
-#  exit with error, this will then leads to a kernel panic
-#  because it's the init script
 set -euo pipefail
 
 # close stdin
```
shruthi-ravi, line 4: Where are you removing content referred to in the commit?

recipes-core/images/nilrt-base-bundle.bb
```diff
@@ -6,6 +6,8 @@ BUNDLE_IMAGE = "nilrt-base-bundle-image"
 
 DEPENDS = "${BUNDLE_IMAGE}"
 
+LICENSE_CREATE_PACKAGE = "0"
```
shruthi-ravi, line 9: Why not?

scripts/lib/wic/canned-wks/nilrt-restore-mode-image.x64.wks
```diff
@@ -2,7 +2,7 @@
 # long-description: Creates an EFI and legacy bootable hybrid ISO image
 # which can be used on optical media as well as USB media.
```
shruthi-ravi, line 4: The commit message mentions a replacement for the deleted files. Is that referring to replacements in a different commit?

recipes-core/images/nilrt-base-bundle.bb
amstewart: We haven't been creating `*-lic` subpackages for IPKs in the dist feed, largely because they are just thin wrappers around other content - which have their own licenses. I don't see a compelling reason to pollute the dist/ feed with IPKs that no one would ever care about.

recipes-core/images/nilrt-base-image.bb
amstewart: Yeah; I was kind of hoping that I would be able to delete all the non-RAUC stuff. :)

I didn't add "RAUC" anywhere in here because it doesn't seem like a relevant part of the image's purpose. The RAUC bundle image and bundle recipe are certainly RAUC-specific, but here it is just an implementation detail. I'm hoping that it isn't going to confuse anyone for the - hopefully - short time that it coexists with the `nilrt-dkms` images (which are already labeled as "legacy" in the README.)

recipes-core/images/nilrt-restore-mode-image.bb
amstewart: Good catch; will fix in a V2.

recipes-core/images/nilrt-restore-mode-image.bb
amstewart: Well, we don't really have anything more minimal at the moment. I entertained the idea of creating two image layers, one with NIFeeds content and the other without. But that requires more fixups to the recipes (like disabling NIAuth in getty) and I didn't know what your BUILD_FROM_FEEDS patchset would look like.

I'm fine refactoring this after your patchset is in, but I think this is fine for the moment.

Also the `minimal-nilrt-*` content is the remnants of the `nilrt-nxg` pipeline. These changes only show as a diff because they had some convenient logic for building RAUC bundles. They are otherwise totally deprecated.

recipes-core/initrdscripts/files/init-nilrt-ramfs.sh
amstewart: Nowhere. I was going to fixup the grammar in the comment and then decided that it was... obvious... and removed the comment entirely.

recipes-core/initrdscripts/files/ni_provisioning.common
amstewart: Under normal circumstances, there is only ever one image in the bundle - hence why both I and the previous author `break` out after the first loop. Both of our stanzas are just using the `for... do` loop to parse the first element from the wildcard expansion or inline element list respectively.

I could replace it with some explicit lines to parse out the first element. The only substantial difference would be that, as I have it written now, if there isn't an image file, the script will exit with a bash syntax error (as the wildcard expansion becomes ""). I think the ROI on using a more explicit codeblock is pretty low here though.

scripts/lib/wic/canned-wks/nilrt-restore-mode-image.x64.wks
amstewart: Yes; it's referring to the `nilrt-restore-mode-image*` recipes added in the other commits.

recipes-core/initrdscripts/files/ni_provisioning.common
shruthi-ravi: I was wondering if the script should return an error if there were more than one image, instead of silently accepting the first in the list. Leaving it to you, feel free to drop.

recipes-core/initrdscripts/files/init-nilrt-ramfs.sh
shruthi-ravi: I was referring to this text, "Remove the `nilrt-runmode-ramfs` content, as it is
no longer being used". Are you also talking about the same?

recipes-core/images/nilrt-base-bundle.bb
shruthi-ravi: Can you add a short comment describing the intent in the commit message?

recipes-core/initrdscripts/files/ni_provisioning.common
amstewart: I think it's pretty unlikely that we'll accidentally include more than one .tar.bz2 in the `bundle/` directory. To do that, we would need to intentionally modify the `bundle.bbclass` - which is owned by meta-rauc - and then also provide some alternate content in the bundle recipe. Maybe we'll feel differently about it once we start development, but I think just using a `for...do` loop is sufficient for this patchset.

recipes-core/initrdscripts/files/init-nilrt-ramfs.sh
amstewart: Oh! I had removed that in a prior version of the commit, but the changes got squashed out when we decided to keep the x64-dkms legacy images. I'll fixup the commit message.

recipes-core/images/nilrt-base-bundle.bb
amstewart: Sure.

amstewart:
Patch V2:
* Remove the `, uses minimal-nilrt-image` substring from the restore mode image recipe because: (a) the payload can change in the future and it is annoying to have to update that variable, (b) that fact is asserted in code by the very next line in the recipe, and (c) the "minimal-image" part is no longer accurate.
* Reword the first `initrdscripts` commit to correctly reflect its current content.
* Reword the `nilrt-base-bundle` commit to explain the rationale behind disabling license subpackages for that image recipe.

shruthi-ravi (approved):

PR #44, opened 2020-10-23
Clean up EFI variables left behind by efi-pstore

bstreiff:
Our 4.14-kernel-based releases have had CONFIG_EFI_PSTORE enabled, but we're
not currently doing anything with the dump information that gets left behind
as efi variables, since we do not have systemd-pstore to come along and
retrieve this information. Therefore, dump information can accumulate to the
point that other operations to add or update EFI variables will fail for lack
of storage.

As the quick-fix for dealing with this, delete all EFI variables associated
with the Linux crash GUID on startup, since we're not currently making use of
that information anyway. (Ideally, a future version will be able to remove
this script and utilize this information.)

Also delete all crash-related EFI variables during provisioning; these are
even less important to keep around at the time of provisioning, because the
user is seeking to remove any prior system state.

### DIFF
--- a/recipes-core/initrdscripts/files/ni_provisioning.common
+++ b/recipes-core/initrdscripts/files/ni_provisioning.common
@@ -289,10 +289,23 @@ elif [ "$ARCH" = "x86_64" ]; then
         ASK_BEFORE_REBOOT=1
     }
 
+    prune_efi_crash_vars()
+    {
+        # Previous kernels may have stored kernel crash information in EFI.
+        # Unfortunately if we have a backlog, this can prevent us from adding
+        # additional boot entries, since those are also EFI variables. Since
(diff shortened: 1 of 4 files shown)


bstreiff:
I've tested that the script operates as expected (deletes EFI variables, can be disabled via /etc/defaults), but I haven't tested it fully in installed-from-package form.

The provisioning codepaths also need tested. I added those as separate codepaths, because it doesn't look like the recovery/restore mode images use packagegroup-ni-base or even run initscripts.

amstewart (approved):
Looks good to me.

I assume this same issue exists in the RAUC bootflow as well. In that case, can you prepare a PR against the dunfell master as well.

bstreiff:
done: https://github.com/ni/meta-nilrt/pull/45

PR #48, opened 2020-10-31
meta-nilrt: Changes to build image from feeds

shruthi-ravi:
Preliminary PR for the changes to build image from feeds. The following are the outstanding workitems that will be addressed once the first set of changes are in:
	
- investigate ldconfig deletion of symlinks at the end of rootfs task
- replace NI_MAIN_FEED_URI with location to feed export derived from the package file when creating non-final builds within nibuild
- fix license copying, tar creation issue when building images from feed

@ni/rtos

### DIFF
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -25,3 +25,4 @@ KERNEL_IMAGEDEST = "boot/runmode"
 ROOT_HOME = "/home/admin"
 
 IMAGE_LINGUAS ?= "en-us en-us.iso-8859-1"
+
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -104,6 +104,9 @@ NILRT_FEEDS_URI_RELEASE            ?= "${NILRT_FEEDS_URI}/${NILRT_RELEASE_NAME}"
 NILRT_MACHINE_FEED_URI_x64         ?= "${NILRT_FEEDS_URI_RELEASE}/x64"
 NILRT_MACHINE_FEED_URI_xilinx-zynq ?= "${NILRT_FEEDS_URI_RELEASE}/arm"
 
+NILRT_MAIN_FEED_VERSION            ?= "2020.6"
+NILRT_MACHINE_FEED_URI_nimain      ?= "${NILRT_FEEDS_URI}/${NILRT_MAIN_FEED_VERSION}/ni-main"
+
 # OE Packages with TUNEARCHes
 PACKAGE_FEED_URIS       = "${NILRT_MACHINE_FEED_URI}"
 PACKAGE_FEED_BASE_PATHS = "extra main"
@@ -116,3 +119,19 @@ PACKAGE_FEED_URIS_ADDITIONAL = 'fake value for PACKAGE_FEED_URIS_ADDITIONAL'
 
 # NI distribution packages
 PACKAGE_FEED_URIS_ADDITIONAL[NI-dist] = "${NILRT_FEEDS_URI}/dist"
+
+# NI-main packages
+PACKAGE_FEED_URIS_ADDITIONAL[ni-main] = "${NILRT_MACHINE_FEED_URI_nimain}"
+
+BUILD_IMAGES_FROM_FEEDS = "1"
+
+# Get NIOE packages from local deploy feed and NI packages from web feed
+NILRT_LOCAL_FEED_URI ?= "file://${DEPLOY_DIR_IPK}"
+
+# Configure feed URIs for image builds
+IPK_FEED_URIS += "\
+    NIOE-all##${NILRT_LOCAL_FEED_URI}/all \
+    NIOE-${MACHINE}##${NILRT_LOCAL_FEED_URI}/${MACHINE} \
+    NIOE-${TUNE_PKGARCH}##${NILRT_LOCAL_FEED_URI}/${TUNE_PKGARCH} \
+    NI-software##${NILRT_MACHINE_FEED_URI_nimain} \
+"
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -4,6 +4,7 @@ TARGET_VENDOR = "-nilrt"
 
 GRUB_BRANCH = "nilrt/19.0"
 
+DISTRO_FEATURES += "${DISTRO_FEATURES_DEFAULT}"
 # Common features enabled on all NILRT flavours
 DISTRO_FEATURES += "\
         argp \
--- a/recipes-core/images/include/licenses.inc
+++ b/recipes-core/images/include/licenses.inc
@@ -10,5 +10,7 @@ tar_licenses() {
 	[ ! -e "${IMAGE_ROOTFS}/usr/share/common-licenses" ]
 }
 
-ROOTFS_POSTPROCESS_COMMAND += "tar_licenses; "
+# tar_licenses() does not work when using BUILD_IMAGE_FROM_FEEDS,
+# because licenses are not available in the rootfs.
+#ROOTFS_POSTPROCESS_COMMAND += "tar_licenses; "
 
--- a/recipes-core/images/nilrt-dkms-image.bb
+++ b/recipes-core/images/nilrt-dkms-image.bb
@@ -10,5 +10,6 @@ require niconsole-image.inc
 require nilrt-xfce.inc
 require nilrt-initramfs.inc
 require include/licenses.inc
+require nilrt-proprietary.inc
 
 IMAGE_FSTYPES += "squashfs"
--- a/recipes-core/packagegroups/packagegroup-ni-base.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-base.bb
@@ -115,6 +115,7 @@ RDEPENDS_${PN} = "\
 	openssh-sftp-server \
 	openssh-ssh \
 	opkg \
+	opkg-utils \
 	opkg-keyrings \
 	os-release \
 	run-postinsts \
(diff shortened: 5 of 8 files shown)


shruthi-ravi:
Btw, I'll likely need to merge some of these changes into the newer RAUC image recipes you, @amstewart, checked in in this [PR](https://github.com/ni/meta-nilrt/pull/42).

conf/distro/nilrt.conf
```diff
@@ -25,3 +25,23 @@ KERNEL_IMAGEDEST = "boot/runmode"
 ROOT_HOME = "/home/admin"
 
 IMAGE_LINGUAS ?= "en-us en-us.iso-8859-1"
+
+BUILD_IMAGES_FROM_FEEDS = "1"
+
+# Get NIOE packages from local deploy feed and NI packages from web feed
+IPK_NIOE_SUBFEED_URI ?= "file://${DEPLOY_DIR_IPK}"
```
amstewart: I think these lines would be better moved to `nilrt.inc`. In theory, the difference between *there* and *here* is that `nilrt.inc` is _how_ to build the distro, and `nilrt.conf` is _what_ to put in the distro. In reality, we don't follow that separation very well.

conf/distro/nilrt.conf
```diff
@@ -25,3 +25,23 @@ KERNEL_IMAGEDEST = "boot/runmode"
...
+
+FEED_SRC ?= "download.ni.com/ni-linux-rt/feeds"
```
amstewart: Could you name this something like `NI_FEEDS_URI` to keep it closer to the extant `NILRT_FEEDS_URI` variable? Or is this variable a part of the BUILD_FROM_FEEDS feature?

conf/distro/nilrt.conf
```diff
@@ -25,3 +25,23 @@ KERNEL_IMAGEDEST = "boot/runmode"
...
+NI_MAIN_FEED_VERSION ?= "2020.5"
+NI_SALT_FEED_VERSION ?= "20.0"
+IPK_NI_MAIN_SUBFEED_URI ?= "http://${FEED_SRC}/${NI_MAIN_FEED_VERSION}/ni-main/"
+IPK_NI_SALT_SUBFEED_URI ?= "http://${FEED_SRC}/SystemLink/salt/${NI_SALT_FEED_VERSION}/${MACHINE}/${TUNE_PKGARCH}/"
```
amstewart: I think we can drop the `NI_SALT` content here. Those salt feeds shouldn't contain anything that isn't already in the NIOE feeds. If they do, we need to reconcile the differences.

recipes-core/images/nilrt-internal.inc
```diff
@@ -0,0 +1,34 @@
+# NI infrastructure packages to be installed in the image
+IMAGE_INSTALL_NODEPS = "\
```
amstewart: Sort lines, pls. :)

recipes-core/images/nilrt-dkms-image.bb
```diff
@@ -10,5 +17,6 @@ require niconsole-image.inc
 require nilrt-xfce.inc
 require nilrt-initramfs.inc
 require include/licenses.inc
+require nilrt-internal.inc
```
amstewart: Could I suggest a name like `nilrt-proprietary.inc` instead? I think that would better communicate its purpose to customers - who are probably familiar with the "proprietary" and "nonfree" feeds in other distros.

recipes-core/images/nilrt-dkms-image.bb
```diff
@@ -1,5 +1,12 @@
 DESCRIPTION = "Runmode image for ethernet-based NI Linux Real-Time targets running XFCE (DKMS)."
 
+# Disable running ldconfig at the end of the rootfs creation process.
+# Running ldconfig removes symlinks created by ni packages installed 
+# to the rootfs, particularly the symlinks to libraries in 
+# NIWebServer. This needs to be investigated before ldconfig can be
+# reenabled.
+LDCONFIGDEPEND = ""
```
amstewart: Should this be a part of the `nilrt-internal.inc` instead? Is there a time that you want to include the proprietary content, but would like to keep `LDCONFIGDEPEND`?

recipes-core/packagegroups/packagegroup-ni-base.bb
```diff
@@ -139,3 +143,96 @@ RDEPENDS_${PN} = "\
 		'${NILRT_NXG_PACKAGES}', \
 		'${NILRT_PACKAGES}', d)} \
 "
+
+# Packages required by SystemLink
+RDEPENDS_${PN} = "\
```
amstewart: Thanks for moving these from extras. Sort lines, pls. :)

conf/distro/nilrt.conf
```diff
@@ -25,3 +25,23 @@ KERNEL_IMAGEDEST = "boot/runmode"
...
+
+# Configure feed URIs for image builds
+IPK_FEED_URIS += "\
+    NIOE-all##${IPK_NIOE_SUBFEED_URI}/all \
+    NIOE-${MACHINE}##${IPK_NIOE_SUBFEED_URI}/${MACHINE} \
+    NIOE-${TUNE_PKGARCH}##${IPK_NIOE_SUBFEED_URI}/${TUNE_PKGARCH} \
+    NI-software##${IPK_NI_MAIN_SUBFEED_URI} \
+    NI-salt##${IPK_NI_SALT_SUBFEED_URI} \
+"
```
amstewart: These lines can probably be moved to `nilrt.inc` as well. They fit well with the rest of the [OPKG SOURCE FEEDS](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/conf/distro/nilrt.inc#L112) section.

conf/distro/nilrt.conf
```diff
@@ -25,3 +25,23 @@ KERNEL_IMAGEDEST = "boot/runmode"
...
+
+BUILD_IMAGES_FROM_FEEDS = "1"
+
+# Get NIOE packages from local deploy feed and NI packages from web feed
+IPK_NIOE_SUBFEED_URI ?= "file://${DEPLOY_DIR_IPK}"
+
+FEED_SRC ?= "download.ni.com/ni-linux-rt/feeds"
+NI_MAIN_FEED_VERSION ?= "2020.5"
+NI_SALT_FEED_VERSION ?= "20.0"
+IPK_NI_MAIN_SUBFEED_URI ?= "http://${FEED_SRC}/${NI_MAIN_FEED_VERSION}/ni-main/"
```
amstewart: Is it true that the image which you build from this patchset won't have an opkg config line for the `ni-main` feed source? I think it won't because you haven't asked the package_manager bbclass to write one in the rootfs with either the `PACKAGE_FEEDS*` variables (upstream-standard) or the `PACKAGE_FEED_URIS_ADDITIONAL` (nilrt-specific).

If that's the case, you probably want to [add a line](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/conf/distro/nilrt.inc#L130) to `nilrt.inc` like
```
PACKAGE_FEED_URIS_ADDITIONAL[ni-main] = "${NILRT_FEEDS_URI_RELEASE}/ni-main"
```

recipes-core/images/include/licenses.inc
```diff
@@ -10,5 +10,6 @@ tar_licenses() {
 	[ ! -e "${IMAGE_ROOTFS}/usr/share/common-licenses" ]
 }
 
-ROOTFS_POSTPROCESS_COMMAND += "tar_licenses; "
+# Fix this function for the BUILD_IMAGE_FROM_FEEDS case
```
amstewart: Could you reword this comment? I was confused the first time reading it. Something like
`# tar_licenses() doesn't work when using BUILD_IMAGE_FROM_FEEDS, because licenses are not available in the rootfs.`

amstewart (changes requested):

conf/distro/nilrt.conf
shruthi-ravi: I've moved all of these lines to nilrt.inc and reused the NILRT_FEEDS_URI variable.

conf/distro/nilrt.conf
shruthi-ravi: Yes, the image does not have that config line. This is currently added by the conf [files](https://dev.azure.com/ni/DevCentral/_git/ni-central?version=GCea6985a18c9614ca87ba416cb4a71da550f2014e&path=%2Fsrc%2Frtos%2FSystemImaged%2Fnilrt_onert_system_image%2Fsource%2Frss%2Fcdf%2Fni-software-cg.conf) in SystemImage. Do you want to add it to the base image instead?

conf/distro/nilrt.conf
amstewart: That works too.

conf/distro/nilrt.conf
amstewart: Yeah; any feed from which we install IPKs during image creation needs to be represented in the opkg `.conf` files.

conf/distro/nilrt.inc
```diff
@@ -104,6 +104,9 @@ NILRT_FEEDS_URI_RELEASE            ?= "${NILRT_FEEDS_URI}/${NILRT_RELEASE_NAME}"
 NILRT_MACHINE_FEED_URI_x64         ?= "${NILRT_FEEDS_URI_RELEASE}/x64"
 NILRT_MACHINE_FEED_URI_xilinx-zynq ?= "${NILRT_FEEDS_URI_RELEASE}/arm"
 
+NILRT_MAIN_FEED_VERSION            ?= "2020.6"
```
shruthi-ravi, line 108: Need to have this variable because the 2021 ni feeds haven't been created yet.

amstewart (approved):

PR #49, opened 2020-11-02
Apply patches to make 'opkg install --noaction <package>' faster

zraul123:
Signed-off-by: Raul Zavaczki <raul.zavaczki@ni.com>

SystemsManagement added two new features that heavily rely on 'opkg install --noaction' to get a summary of packages that are going to be installed. Unfortunately, this operation takes extremely long (~90 seconds with 'opkg install --noaction ni-crio' on a freshly formatted target) and the expected behavior is that these operations take at most a few seconds.
After some investigations, we found that if the 'download_first' option is supplied, 'opkg install --noaction' will download the packages, and that a feed is verified for each package - even if it was verified before.
The patches optimize these two use-cases.

The setup I've been developing this under is:
 - I cloned https://github.com/ni/nilrt on an Ubuntu VM and followed the instructions there
 - Built sources with 'bitbake opkg' and deployed on real targets with 'devtool deploy_target'

Question: I saw in previous PR that we need to test the patches against nilrt/master/dunfell branch too - is it correct into the same workspace to "git checkout nilrt/master/dunfell -> apply patches -> bitbake opkg -> devtool deploy_taget -> test everything works"? This is the first time I use bitbake/devtool

### DIFF
--- a/recipes-core/base-files/base-files_3.0.14.bbappend
+++ b/recipes-core/base-files/base-files_3.0.14.bbappend
@@ -35,6 +35,10 @@ do_install_append () {
 	install ${WORKDIR}/safemode-ps1.sh ${D}${sysconfdir}/profile.d/
 
 	install -d ${D}${sysconfdir}/default/volatiles/
+	# 10_var_vol_cache is only needed for post-8.5 safemodes to support
+	# chrooted opkg operations on pre-8.5 runmodes.
+	echo "d root root 0755 /var/volatile/cache none" \
+		>> ${D}${sysconfdir}/default/volatiles/10_var_vol_cache
 	echo "d ${LVRT_USER} ${LVRT_GROUP} 0775 /run/natinst none" \
 		>> ${D}${sysconfdir}/default/volatiles/20_run_natinst
 }


recipes-devtools/opkg/files/0001-opkg_solver_libsolv-Pass-opkg_config-noaction-as-noa.patch
```diff
@@ -0,0 +1,30 @@
+From 852b31b13de3ab739f889345b982a82970614cf7 Mon Sep 17 00:00:00 2001
+From: Zavaczki Raul <raul.zavaczki@ni.com>
+Date: Mon, 26 Oct 2020 05:57:41 -0700
+Subject: [PATCH] opkg_solver_libsolv: Pass opkg_config->noaction as 'noaction'
+ parameter to libsolv_solver_transaction_preamble
+
+Currently, if the option for 'download first' is true, opkg will download packages even when noaction is set to false
+Changing the function call to pass opkg_config->noaction instead of 0 will fix this behaviour
+
+Signed-off-by: Zavaczki Raul <raul.zavaczki@ni.com>
```
amstewart: Could you add an `Upstream Status: Accepted...` line to the commit message trailer in the `.patch`es, similar to [this](https://github.com/ni/meta-nilrt/blob/0a38a5b5a87c4dd315c6ffedbd75f927624c6f5d/recipes-devtools/opkg/files/0002-libopkg-clear-curl-properties-on-download-error-to-p.patch#L36). It helps us to correctly handle the .patch during a rebase.

recipes-devtools/opkg/files/0001-pkg_src_verify-Verify-feeds-signatures-only-once.patch
```diff
@@ -0,0 +1,104 @@
+From a00ab183da1559a7f6e411af4134b4ec1d911af4 Mon Sep 17 00:00:00 2001
+From: Zavaczki Raul <raul.zavaczki@ni.com>
+Date: Mon, 26 Oct 2020 04:36:23 -0700
+Subject: [PATCH] pkg_src_verify: Verify feeds signatures only once
+
+Rename pkg_src_t disable_sig_check to signature_verified to better
+reflect the field purpose
+
+Signed-off-by: Zavaczki Raul <raul.zavaczki@ni.com>
+---
```
amstewart: Same comment about an upstream status line.

recipes-devtools/opkg/opkg_%.bbappend
```diff
@@ -9,6 +9,8 @@ SRC_URI += " \
             file://run-ptest \
             file://0001-opkg-key-add-keys-even-if-creation-date-is-in-the-fu.patch \
             file://0002-libopkg-clear-curl-properties-on-download-error-to-p.patch \
+            file://0001-opkg_solver_libsolv-Pass-opkg_config-noaction-as-noa.patch \
+            file://0001-pkg_src_verify-Verify-feeds-signatures-only-once.patch \
```
amstewart: `.patch`es are applied to recipe sources in alphanumeric-order. In this case, it doesn't matter because your patches are functionally independent and so don't require any particular ordering. But it's more common that you would renumber the `.patch` file name to explicitly define their sequence. eg.

```bash
file://0003-opkg_solver_libsolv-Pass-opkg_config-noaction-as-noa.patch \
file://0004-pkg_src_verify-Verify-feeds-signatures-only-once.patch \
```

I'll leave it up to you to decide if you want to re-number the patches, but I do ask that you keep the batch of `.patch` files in the bbappend in sorted order, rather than adding your new patches at the end. Keeping them sorted helps other devs reviewing the bbappend to understand the order in which they are going to be applied by bitbake.

amstewart (changes requested):
Thanks for putting in this changelist, Raul! The opkg content all looks good and I appreciate you getting it upstreamed to opkg before backporting it.

I have a couple minor nits below, which are mostly process related. RTOS is intending cut beta branches for the 20.6 release pretty soon, so you got this changelist in at a good time. :)

amstewart:
>Question: I saw in previous PR that we need to test the patches against nilrt/master/dunfell branch too - is it correct into the same workspace to "git checkout nilrt/master/dunfell -> apply patches -> bitbake opkg -> devtool deploy_taget -> test everything works"? 

The Dunfell master branch corresponds to the 21Q2 release. It was rebased upon the OE dunfell upstream during the summer and so is an independent development trunk.


**RE: how to build changes for dunfell**
You're broadly correct. You can just checkout the `nilrt/master/dunfell` ref of the nilrt.git repo and the same ref of each of the submodule repos. Then you can use the same build and test workflows that you used to create this patchset. However, after checking out the new refs, you will want to use a different build workspace from the sumo branch. The sstate cache and other bitbake contextual information are different between the two bases. You can pass a different build directory name to the [initialization script](https://github.com/ni/nilrt/blob/nilrt/master/dunfell/nilrt-build-init.env) like

```bash
. ./nilrt-build-init.env build-dunfell
```

Note that the initialization script changed contents between the sumo and dunfell master branches. 

For those reasons, and others, I find its easiest to clone different rebases into entirely separate directories and maintain them independently.


**RE: whether we need to port these .patches to dunfell**
Since these `.patch`es are already upstreamed with opkg, they will be included in the 0.4.4 release in December. I'm inclined to upgrade the entire `opkg` recipe in dunfell to 0.4.4 at that time.

So, if SL wants this opkg bug fixed in the dunfell development images between now and 21Q1, then you can submit a similar PR to the dunfell/master branch and we can pull it now. If you can wait, RTOS will pickup the change from upstream when we upgrade the base recipe.

recipes-devtools/opkg/opkg_%.bbappend
zraul123: I renumbered the patches so - it looks cleaner this way :-)

zraul123:
Hey @amstewart,
Thank you for the quick reply and for the answer, now I understand what's the link between everything! I don't know whether we need to port these changes to dunfell - most probably yes - but I'll post a PR on the dunfell branch aswell if we do! Just out of curiosity: sumo is for 21Q1 release? Or the one in december (20Q4 I guess)?
I don't know why it closed my PR, but I posted an updated one here: https://github.com/ni/meta-nilrt/pull/50

Thanks again! :-)

amstewart:
> Just out of curiosity: sumo is for 21Q1 release? Or the one in december (20Q4 I guess)?

Yes; the 21Q1 release will use the Sumo mainline; I think that release will be versioned 20.7. We aren't planning to do a Cycle X (20Q4) release.

To clarify: when I said "21Q1" above, I was talking about "the 21Q1 timeframe", not the release. So the 21Q1 release (20.7) will inherit your changes from this PR. The first release which ships the dunfell master will be 21Q2 (I don't know what its version will be yet). It's all very confusing.

PR #51, opened 2020-11-02
rauc: Fix safemode to rauc provisioning failures

chaitu236:
On small memory targets, initrd containing payload doesn't
fit in RAM. These changes fix the issue.

There are now 2 iso images and 2 ipk images
  1. nilrt-restore-mode-image.iso - contains payload (rauc bundle),
     kernel, initrd, etc. Payload is outside initrd
     This is used for USB provisioning
  2. nilrt-restore-mode-boot-image.iso - same as (1) but without payload
     and is used in (4) below for IPK provisioning
  3. dist-nilrt-efi-ab.ipk - contains payload
  4. dist-nilrt-efi-ab-gateway.ipk - contains (2) and a scripts to
     set up target to boot into provisioning in next boot.

When provisioning from a USB formatted with nilrt-restore-mode-image.iso:
  1. Kernel, initrd are loaded from USB
  2. UUID of USB is passed from grub and is mounted
  3. Since root of USB contains payload, this is used for provisioning

For gateway ipk
  1. When gateway ipk is installed and its nilrt-install script is run,
     dist-nilrt-efi-ab.ipk is downloaded to nirootfs/payload
  2. If (1) is successful, nibootfs is overwritten with contents of
     nilrt-restore-mode-boot-image.iso
  3. After reboot, kernel and initrd are loaded from nibootfs
  4. UUID of nibootfs is passed down from grub and is mounted
  5. Since root of this partition does not contain payload, it is
     determined that this is ipk provisioning and payload is obtained
     from nirootfs partition

Testing:
  Verified USB, IPK provisioning works on a VM with same amount of memory
   as our smallest memory targets and on a cRIO-9030.

@ni/rtos

### DIFF
--- a/recipes-core/dist-nilrt/files/nilrt-gateway-install
+++ b/recipes-core/dist-nilrt/files/nilrt-gateway-install
@@ -26,6 +26,7 @@ function cleanup() {
     # Restore boot partitions to previous states
     [ ! -z "${EFI_MOUNT_A-}" ] && mount -o remount,ro "$EFI_MOUNT_A" &>/dev/null
     [ ! -z "${EFI_MOUNT_B-}" ] && umount "$EFI_MOUNT_B" &>/dev/null
+    [ ! -z "${NIROOTFS_MOUNT-}" ] && umount "$NIROOTFS_MOUNT" &>/dev/null
 
     exit "$exitCode"
 }
@@ -36,6 +37,19 @@ function die() {
     echo "Error: $1"
     exit 1
 }
+
+# Download dist-nilrt-efi-ab.ipk into /payload of nirootfs partition
+NIROOTFS_MOUNT=$(mktemp -d "/var/volatile/niroot-XXXXXXX")
+mount -L nirootfs $NIROOTFS_MOUNT || die "Unable to mount nirootfs."
+
+rm -rf $NIROOTFS_MOUNT/payload
+mkdir $NIROOTFS_MOUNT/payload || die "Unable to create payload directory."
+
+pushd $NIROOTFS_MOUNT/payload > /dev/null
+opkg update > /dev/null
+opkg download dist-nilrt-efi-ab > /dev/null || (popd > /dev/null; die "Unable to download dist-nilrt-efi-ab. Check if disk has sufficient space.")
+test -e dist-nilrt-efi-ab*.ipk || (popd > /dev/null; die "Unable to download dist-nilrt-efi-ab. Check if feeds contain dist-nilrt-efi-ab package.")
+popd > /dev/null
 
 # Mount the ISO/wic file as a loopback device
 info "Mounting the ISO"
--- a/recipes-core/initrdscripts/files/init-restore-mode.sh
+++ b/recipes-core/initrdscripts/files/init-restore-mode.sh
@@ -8,6 +8,17 @@ umask 0022
 
 NIRECOVERY_MOUNTPOINT=/mnt/NIRECOVERY
 MOUNT_NIRECOVERY_USB_TIME=10
+BOOTMEDIA_UUID=
+
+parse_args() {
+   for x in $(cat /proc/cmdline); do
+       case $x in
+       bootmedia_uuid=*)
+          BOOTMEDIA_UUID=${x#bootmedia_uuid=}
+          ;;
+       esac
+   done
+}
 
 early_setup() {
     mkdir -p /proc
@@ -18,6 +29,8 @@ early_setup() {
     mount -t efivarfs efivarfs /sys/firmware/efi/efivars
     mount -t devtmpfs none /dev
 
+    parse_args
+
     COUNT=0
     while [ $COUNT -le "$MOUNT_NIRECOVERY_USB_TIME" ]; do
         mount_nirecovery_usb
@@ -27,6 +40,7 @@ early_setup() {
         COUNT=$(expr $COUNT + 1)
         sleep 1
     done
+    mount_payload
 
     # Set hostname
     echo "recovery" | tee /etc/hostname > /proc/sys/kernel/hostname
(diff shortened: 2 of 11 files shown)


recipes-core/dist-nilrt/files/nilrt-gateway-install
```diff
@@ -13,6 +13,11 @@ function cleanup() {
 
     set +e
 
+    if [ -e "${NIROOTFS_MOUNT:-}" ]; then
+        umount "$NIROOTFS_MOUNT"
+        rmdir "$NIROOTFS_MOUNT"
+    fi
+
```
amstewart: Why do you need both this block and line 34? Aren't they doing the same things?

recipes-core/dist-nilrt/files/nilrt-gateway-install
```diff
@@ -37,6 +43,19 @@ function die() {
...
+# Download dist-nilrt-efi-ab.ipk into /payload of nirootfs partition
+NIROOTFS_MOUNT=$(mktemp -d "/var/volatile/niroot-XXXXXXX")
+mount -L nirootfs $NIROOTFS_MOUNT || die "Unable to mount nirootfs."
+
+rm -rf $NIROOTFS_MOUNT/payload
+mkdir $NIROOTFS_MOUNT/payload || die "Unable to create payload directory."
+
+pushd $NIROOTFS_MOUNT/payload
+opkg update > /dev/null
+opkg download dist-nilrt-efi-ab > /dev/null || (popd; die "Unable to download dist-nilrt-efi-ab. Check if disk has sufficient space.")
```
amstewart: I don't think you need the `popd` calls here. Your `pushd` changes should only be within the context of this script, so there is no lasting effect after you `die()`.

recipes-core/dist-nilrt/files/nilrt-gateway-install
```diff
@@ -37,6 +43,19 @@ function die() {
...
+ls dist-nilrt-efi-ab*.ipk &> /dev/null || (popd; die "Unable to download dist-nilrt-efi-ab. Check if feeds contain dist-nilrt-efi-ab package.")
```
amstewart: Is there a reason to use `ls` here instead of `test -e`?

Also the fact that `opkg` returns 0 when it can't satisfy a download request is a bug. I'll fix it upstream when I get some time. But this is fine until then.

recipes-core/dist-nilrt/files/nilrt-gateway-install
```diff
@@ -37,6 +43,19 @@ function die() {
     exit 1
 }
 
+# Download dist-nilrt-efi-ab.ipk into /payload of nirootfs partition
+NIROOTFS_MOUNT=$(mktemp -d "/var/volatile/niroot-XXXXXXX")
+mount -L nirootfs $NIROOTFS_MOUNT || die "Unable to mount nirootfs."
```
amstewart, line 43: Does this mount as read-write or read-only?

recipes-core/initrdscripts/files/init-restore-mode.sh
```diff
@@ -86,10 +100,38 @@ mount_nirecovery_usb()
...
+
+      ar x $ROOTFS_MOUNTPOINT/payload/*.ipk
+      mkdir data
+      tar -xpf data.tar* -C data
+
+      # Ramdisk size is limited, so copy payload to a tmpfs
+      mount -t tmpfs -o size=420m tmpfs /payload
+      cp data/usr/share/nilrt/* /payload
+      mv /payload/*.raucb /payload/niboot.raucb
+      popd
```
amstewart: I think you're still working on non-volatile storage for the this IPK extraction. The contents of `$ROOTFS_MOUNTPOINT/payload/data` are still going to be on nirootfs and contain a copy of the RAUC bundle - which means that we could run out of storage when extracting it.

I was thinking of something more like.
```bash
mount -t tmpfs -o size=420m tmpfs /payload

mkdir /mnt/nirootfs
mount -o ro -L nirootfs /mnt/nirootfs

ar p /mnt/nirootfs/payload/*.ipk data data.tar.xz data.tar.gz | \
    tar -xJp --wildcards '**/*.raucb' -C /payload
find /payload -type f -path '**/*.raucb' -exec mv {} /payload \;
```

That way, we're never writing more to the rootfs partition in the provisioning context. If we do, and we run out of space, it's a shitty customer experience.

amstewart (changes requested):

amstewart:
> There are now 2 iso images and 2 ipk images
> 
>     1. nilrt-restore-mode-image.iso - contains payload (rauc bundle),
>        kernel, initrd, etc. Payload is outside initrd
>        This is used for USB provisioning
> 
>     2. nilrt-restore-mode-boot-image.iso - same as (1) but without payload
>        and is used in (4) below for IPK provisioning
> 
>     3. dist-nilrt-efi-ab.ipk - contains payload
> 
>     4. dist-nilrt-efi-ab-gateway.ipk - contains (2) and a scripts to
>        set up target to boot into provisioning in next boot.

These restore mode names are getting very confusing. Could we do this:

1. `nilrt-restore-mode.iso` : the kernel and initrd ISO; used by `dist-nilrt-efi-ab-gateway`.
2. `nilrt-recovery-media.iso`: the kernel, initrd, and RAUC bundle ISO; used only by recovery media.

3. `nilrt-restore-mode-image.bb` : the bitbake image target which builds (1)
4. `nilrt-recovery-media-image.bb`: the bitbake image target which builds (2)

recipes-core/dist-nilrt/files/nilrt-gateway-install
chaitu236: Without popd, die wasn't able to unmount NIROOTFS_MOUNT

recipes-core/dist-nilrt/files/nilrt-gateway-install
chaitu236: > Is there a reason to use ls here instead of test -e?

Wasn't aware of test -e. Will change to it.

> Also the fact that opkg returns 0 when it can't satisfy a download request is a bug. I'll fix it upstream when I get some time. But this is fine until then.

Thanks

recipes-core/dist-nilrt/files/nilrt-gateway-install
chaitu236: read/write. We want to to extract payload to the same partition so it needs to be writable

recipes-core/initrdscripts/files/init-restore-mode.sh
chaitu236: I intended to remove content from nirootfs after reboot, but looks like I missed it.

> I was thinking of something more like.

I can test it out

recipes-core/dist-nilrt/files/nilrt-gateway-install
amstewart: Ah; understood.

recipes-core/initrdscripts/files/init-restore-mode.sh
chaitu236: > I was thinking of something more like.

Ran out of memory doing this.

Added code to remove everything except payload from nirootfs

recipes-core/dist-nilrt/files/nilrt-gateway-install
chaitu236: You're right. Removed this

chaitu236:
> These restore mode names are getting very confusing. Could we do this:

Renaming isos, recipes created some complications - working through them.

chaitu236:
> > These restore mode names are getting very confusing. Could we do this:
> 
> Renaming isos, recipes created some complications - working through them.

Done

recipes-core/initrdscripts/files/init-restore-mode.sh
```diff
@@ -86,10 +100,46 @@ mount_nirecovery_usb()
...
+   else
+      # We used migration ipk, so payload would be in nirootfs
+      ROOTFS_MOUNTPOINT=$(mktemp -d "/var/volatile/rootfs-XXXXXXX")
+      mount -o rw -L nirootfs $ROOTFS_MOUNTPOINT;
+
+      # Delete everything except payload directory to ensure device
+      # has enough space to extract payload
+      for dir in $ROOTFS_MOUNTPOINT/*; do
+          if [ ${dir##*/} != "payload" ]; then
+              rm -rf $dir;
```
amstewart: If `mktemp` fails for whatever reason (like `/var/volatile` being mounted read-only), `ROOTFS_MOUNTPOINT` will have a null value, and line 121 will `rm -rf` everything from root which isn't `/payload`.

It's unlikely that `mktemp` will fail like that, but can you add a check to keep it from being so destructive?

amstewart (changes requested):
Everything looks good, apart from the one comment about the `mktemp` failure case. Thanks for refactoring the image names.

recipes-core/initrdscripts/files/init-restore-mode.sh
chaitu236: Done. Thanks for reviewing

amstewart (approved):

PR #53, opened 2020-11-04
openvpn: start initscript in rl 4 and 5

amstewart:
The NILRT-specific openvpn-wrapping initscript: `vpn` is configured to start at `rc4.d/S42`. In a RAUC system, this means that the openvpn daemon will only start in the "safemode" equivalent runlevel.

Start it in both runlevels 4 and 5, so that users can configure openvpn to automatically start in both safemode and runmode.

Natinst-AZDO-ID: 1165918


amstewart:
@ni/rtos

gratian (approved):
LGTM. Ship it!

shruthi-ravi (approved):

PR #54, opened 2020-11-04
niwatchdogpet: Ignore PXI targets

jeminor:
Signed-off-by: James Minor <james.minor@ni.com>

### DIFF
--- a/recipes-ni/niwatchdogpet/files/niwatchdogpet.sh
+++ b/recipes-ni/niwatchdogpet/files/niwatchdogpet.sh
@@ -3,6 +3,11 @@
 
 PETTER=/usr/sbin/niwatchdogpet
 
+class="`/sbin/fw_printenv -n TargetClass`"
+
+# PXI targets use the built-in watchdog rather than CPLD
+[ "$class" = "PXI" ] && exit 0 
+
 # Make sure the petting tool is available.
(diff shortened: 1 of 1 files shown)


shruthi-ravi:
Have you tested this change?
And this fix is for a bug right? Can you add the bug ID to the commit message like "Natinst-AZDO-ID: xxx"?

jeminor:
> Have you tested this change?

I tested "by hand" by modifying the file on an emulated cRIO and emulated PXIe controller, but no I haven't built it into an image to insure it will build.  I was really just trying to make sure that I didn't forget what needed to be done.

> And this fix is for a bug right? Can you add the bug ID to the commit message like "Natinst-AZDO-ID: xxx"?

Happy to do so, yes.  I'm curious if we're comfortable with this solution?

shruthi-ravi:
>I tested "by hand" by modifying the file on an emulated cRIO and emulated PXIe controller, but no I haven't built it into an image to insure it will build. I was really just trying to make sure that I didn't forget what needed to be done.

Ok, that should be good enough.

>Happy to do so, yes. I'm curious if we're comfortable with this solution?

Why do you feel otherwise? :) I have found it easy to track the "why" for the change. The commit message is currently pretty barebones.

amstewart (approved):

PR #59, opened 2020-11-13
packagegroup-ni-runmode: Add gdb

shruthi-ravi:
Fix for this odd build [failure](https://dev.azure.com/ni/DevCentral/_build/results?buildId=565007&view=logs&j=7b573d1f-6776-53c2-0b0a-ff05672f5008&t=0a7242b1-32ca-59b5-a501-bcb76228df87&l=32484). When I added this dependency, it built alright locally, again, oddly.

Signed-off-by: Shruthi Ravichandran <shruthi.ravichandran@ni.com>

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-runmode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-runmode.bb
@@ -15,6 +15,7 @@ RDEPENDS_${PN} = "\
 	e2fsprogs-e2fsck \
 	e2fsprogs-mke2fs \
 	e2fsprogs-tune2fs \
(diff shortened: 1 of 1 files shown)


amstewart (approved):

gratian:
@shruthi-ravi  Do we know how much this adds to the image size? The reason we've only included gdbserver in the base image was for space savings with the idea being that people doing debugging can install the full gdb suite from the package feed if they need it (and most customers are not expected to need it). If this increases the base image enough to matter it might be better to take gdbsever out of the packagegroup-ni-runmode instead of adding gdb in.

PR #61, opened 2020-11-19
Replace globbing in file:// with list of files.

shruthi-ravi:
The upstream bitbake parser no longer supports globbing in file://
urls. Replace those instances with the complete list of files to be
included.

Built locally.


recipes-ni/sysconfig-settings-ui/sysconfig-settings-ui_1.0.bb
```diff
@@ -4,7 +4,23 @@ LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 SECTION = "base"
 
-SRC_URI = "file://uixml/* \
+SRC_URI = " \
```
amstewart, line 7: Sort pls. :)

amstewart (approved):
A couple of these file lists need to be sorted; but otherwise looks good!

PR #63, opened 2020-11-25
kernel-devsrc-next: Add recipe

chaitu236:
Backported commits from dunfell for kernel-dev and kernel-devsrc

Add recipe to build kernel-devsrc for kernel-next

@ni/rtos

### DIFF
--- a/recipes-core/packagegroups/packagegroup-kernel-module-build.bb
+++ b/recipes-core/packagegroups/packagegroup-kernel-module-build.bb
@@ -8,5 +8,6 @@ RDEPENDS_${PN} = "\
 	gcc \
 	gcc-symlinks \
 	kernel-dev \
+	kernel-devsrc \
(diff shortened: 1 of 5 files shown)


amstewart (approved):
lgtm. For what it's worth, I don't think we use the `packagegroup-kernel-module-build` pkggrp in the OneRT builds. But it's harmless and will be removed eventually.

chaitu236:
> lgtm. For what it's worth, I don't think we use the `packagegroup-kernel-module-build` pkggrp in the OneRT builds. But it's harmless and will be removed eventually.

Oh, is there another package group or are the users expected to install/depend on these packages if they want to build modules?

amstewart:
I think all modules ought depend on the `dkms` package, so the `RDEPENDS_` you have added is sufficient to get devsrc installed.

PR #64, opened 2020-11-26
nilrt-restore-mode: Use different volume label

chaitu236:
Using NIRECOVERY-CD as volume label for both recovery iso
and gateway IPK causes conflict when recovery media is
used for provisioning right after gateway IPK was used (could
happen if for some reason gateway IPK migration failed) - in
this case both recovery media and one of the target's disk
partitions have same label and grub may use wrong volume
as root.

Using different volume label for nilrt-restore-mode iso
which ends up in gateway IPK.

Tested both provisioning methods on a VM with no provisioning issues.

@ni/rtos


amstewart (commented):
In what scenarios is `NIRECOVERY-HD` actually a partition on disk? Does it only exist in the case that the provisioning process is halted before it completes, or does it persist after a successful provision?

chaitu236:
> In what scenarios is `NIRECOVERY-HD` actually a partition on disk? Does it only exist in the case that the provisioning process is halted before it completes, or does it persist after a successful provision?

Yes, it only exists if provisioning is halted before provisioning script got to deleting disk partitions. But in case of successful provision, it'll cease to exist.

amstewart (approved):
Alright; I agree with this change.

PR #67, opened 2020-12-02
linux-nilrt: Bump the 'next' kernel version to 5.10

gratian:
Upgrade the kernel built by the linux-nilrt-next recipe from 5.6 to
5.10 to enable early testing.


gratian:
Testing done:

- [x] verified that 'bitbake packagegroup-ni-kernel-next' builds w/o error
- [x] verified that installing the resulting package via 'opkg install packagegroup-ni-next-kernel' works as expected on a cRIO-9035
- [x] verified that the system can boot the new kernel successfully

shruthi-ravi (approved):

PR #73, opened 2020-12-11
Fix driver build issue

chaitu236:
Fixes driver build issue on "current kernel".

Tested by installing kernel-devsrc on sumo and installing ni-kal without issues.
Installed kernel*next packages and installed ni-kal without issues.

@ni/rtos

### DIFF
--- a/recipes-kernel/linux/linux-nilrt.inc
+++ b/recipes-kernel/linux/linux-nilrt.inc
@@ -77,6 +77,10 @@ do_install_append() {
 	${WORKDIR}/export-kernel-headers.sh ${S} ${D}/kernel
 }
 
+do_shared_workdir_append() {
+	cp -rf ./* ${STAGING_KERNEL_BUILDDIR}
+}
+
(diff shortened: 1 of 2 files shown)


recipes-kernel/linux/linux-nilrt.inc
```diff
@@ -77,6 +77,10 @@ do_install_append() {
 	${WORKDIR}/export-kernel-headers.sh ${S} ${D}/kernel
 }
 
+do_shared_workdir_append() {
+   cp -rf * ${STAGING_KERNEL_BUILDDIR}
+}
+
```
amstewart: Nits:
1. It looks like you're using soft-tabs here, where the rest of the recipe uses hard-tabs. Could you make it consistent?
2. As long as you're touching up the whitespace, I would recommend using `./*` for the wildcard - only because it seems more clear to me.

amstewart (changes requested):
Hopefully a quick nit below. Otherwise, I agree with the changes' functionality.

amstewart (approved):

PR #76, opened 2020-12-11
kernel-tests: Add futex test for transient "ownerless" rtmutex bug

gratian:
Add reproducer/regression test for the BUG_ON(!newowner) in
fixup_pi_state_owner().

Upstream discussion:
  https://lore.kernel.org/lkml/87a6w6x7bb.fsf@ni.com/

NI-BUG-ID: 1168289

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-test-fbomb-files/fbomb.c
+++ b/recipes-kernel/kernel-tests/kernel-test-fbomb-files/fbomb.c
@@ -1,0 +1,407 @@
+// SPDX-License-Identifier: LGPL-2.1-only
+// Copyright (c) 2020 National Instruments.
+
+#define _GNU_SOURCE
+#include <stdlib.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdarg.h>
+#include <string.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <pthread.h>
+#include <err.h>
+#include <errno.h>
+#include <sched.h>
+#include <time.h>
+#include <sys/types.h>
+#include <sys/syscall.h>
+#include <sys/sysinfo.h>
+#include <linux/futex.h>
+#include <limits.h>
+#include <getopt.h>
+#include <ctype.h>
+
+#define TEST_CPU	1
+#define NSEC_PER_USEC	1000ULL
+#define NSEC_PER_SEC	1000000000ULL
+#define NOISE_THREADS	64
+
+#define FUTEX_TID_MASK	0x3fffffff
+
+typedef uint32_t futex_t;
+
+static futex_t cond;
+static futex_t lock1 = 0;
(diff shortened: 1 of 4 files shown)


gratian:
@ni/rtos

recipes-kernel/kernel-tests/kernel-test-fbomb-files/fbomb.c
```diff
@@ -0,0 +1,407 @@
...
+				goto err_exit;
+                }
+        }
+
+        return EXIT_SUCCESS;
+
+  err_exit:
+	usage(argv[0]);
+	return EXIT_FAILURE;
+}
```
amstewart, line 124: I see there is a `PASS: test_fbomb\n` line output directly from the test. Is this function an appropriate place to express a `FAIL: test_fbomb\n` line?

recipes-kernel/kernel-tests/kernel-test-fbomb-files/run-ptest
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+./fbomb -d 24h
```
amstewart: Recommend that you add an `exec 2>&1` call here to keep your stderr and stdout lines synchronized. [See here](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/3560/Yocto-PTests?anchor=gotchya-%23-1%3A).

recipes-kernel/kernel-tests/kernel-test-fbomb-files/run-ptest
```diff
@@ -0,0 +1,5 @@
...
+
+exit 0
```
amstewart: NIT: I don't think it _strictly_ matters for the ptest runner, but it would probably be better to let the return code from `fbomb` propagate as the `run-ptest` return code. [See here](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/3560/Yocto-PTests?anchor=return-codes)

amstewart (changes requested):

recipes-kernel/kernel-tests/kernel-test-fbomb-files/fbomb.c
gratian: I guess, but this is more of a failure of using the test correctly not an actual failure detected by the test (which we cannot detect within the test itself bc. the kernel will OOPS). Anyway it's not a big deal to add a print here so I'll do it.

gratian: OK, I've cleaned-up the error exit paths and added a FAIL print. See fix-up commit.

amstewart (approved):
ACK all V2 changes. Thanks.

PR #78, opened 2020-12-17
kernel-devsrc-next: Notify dkms of new kernel

chaitu236:
@ni/rtos

### DIFF
--- a/recipes-kernel/linux/kernel-devsrc-next.bb
+++ b/recipes-kernel/linux/kernel-devsrc-next.bb
@@ -14,4 +14,14 @@ pkg_postinst_ontarget_${PN} () {
    cd /lib/modules/${KERNEL_VERSION}/build
    make prepare
    make modules_prepare
+
+   if [ -e /etc/kernel/postinst.d ]; then
+      run-parts /etc/kernel/postinst.d -a ${KERNEL_VERSION}
+   fi
+}
+
+pkg_prerm_${PN} () {
+   if [ -e /etc/kernel/prerm.d ]; then
+      run-parts /etc/kernel/prerm.d -a ${KERNEL_VERSION}
+   fi
 }


amstewart:
I wasn't familiar with this runparts system and googled around for `/etc/kernel/postinst.d` and got to this [debian kernel handbook](https://kernel-team.pages.debian.net/kernel-handbook/ch-update-hooks.html#s-kernel-hooks).

>  The maintainer scripts in Linux kernel packages must use run-parts to invoke hook scripts in the corresponding subdirectory of /etc/kernel, e.g. the postinst script must invoke scripts in /etc/kernel/postinst.d. 

Does that line imply that:
1. since `kernel-devsrc` is a "kernel package" that it should invoke *all* of the contents of `/etc/kernel/postinst.d`, not just the dkms hook?
2. we should prefer to use the `run-parts` binary to do the invocation, rather than relying on the shell to interpret the runparts files?

chaitu236:
> I wasn't familiar with this runparts system and googled around for `/etc/kernel/postinst.d` and got to this [debian kernel handbook](https://kernel-team.pages.debian.net/kernel-handbook/ch-update-hooks.html#s-kernel-hooks).
> 
> > The maintainer scripts in Linux kernel packages must use run-parts to invoke hook scripts in the corresponding subdirectory of /etc/kernel, e.g. the postinst script must invoke scripts in /etc/kernel/postinst.d.
> 
> Does that line imply that:
> 
> 1. since `kernel-devsrc` is a "kernel package" that it should invoke _all_ of the contents of `/etc/kernel/postinst.d`, not just the dkms hook?
> 2. we should prefer to use the `run-parts` binary to do the invocation, rather than relying on the shell to interpret the runparts files?

Yes looks like it. Fixed

amstewart (approved):
lgtm

PR #81, opened 2021-01-07
Enable empty root password login in images.

shruthi-ravi:
Bill was not able to login to the images generated by the OE build. He validated that this change now allows him to login as root.

 @ni/rtos

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -139,6 +139,7 @@ PACKAGE_CLASSES ?= "package_ipk"
 #                     e.g. ssh root access has a blank password
 # There are other application targets that can be used here too, see
 # meta/classes/image.bbclass and meta/classes/core-image.bbclass for more details.
+EXTRA_IMAGE_FEATURES += "empty-root-password allow-empty-password"
 
 #
 # Additional image features


amstewart:
I don't think we were using those OE image features in the old CG build; right? I think there is some nuanced interaction between that feature and our NIAuth configuration. I don't know if it is the case that it just *isn't necessary* when you have NIAuth or if it conflicts.

I'm fine with having it enabled in the base/recovery images, if NIAuth properly replaces it. I just want to ensure that our decision here is intentional. So are you aware of how/if this conflicts with NIAuth?

shruthi-ravi:
You are right, Alex. I had confused the CG and NXG rules. Looks like this was enabled only in NXG images.
This feature does not conflict with NIAuth, nor does NIAuth replace this feature. I would say they work together harmoniously :).
So here's the relevant PAM config file:
image
When I log in, the credentials are first tried against the niauth library, and on authentication failure, the unix library. Since niauth doesn't have a root login, the credentials are tried against the unix library and that's where this shadow file kicks in. On seeing a no-password entry for root in shadow, the unix library approves the login request without requesting a password.

If you still feel wary of adding this, we could revert this commit at a later stage when the niauth is fixed up so it works in the image generated by the bitbake build.

amstewart:
> You are right, Alex. I had confused the CG and NXG rules. Looks like this was enabled only in NXG images.

Yeah; I don't think we ever had the NIAuth database in the LVComms images. You directly logged into `root` on those images.

> When I log in, the credentials are first tried against the niauth library, and on authentication failure, the unix library. Since niauth doesn't have a root login, the credentials are tried against the unix library and that's where this shadow file kicks in. On seeing a no-password entry for root in shadow, the unix library approves the login request without requesting a password.

What happens if a user changes the admin password via sysapi? Does it only apply to the NIAuth database? And will the shadow fallthrough allow you to still login with no password, even though you set one in NIAuth?

jpautler:
> If you still feel wary of adding this, we could revert this commit at a later stage when the niauth is fixed up so it works in the image generated by the bitbake build.

Like both you and Alex have already mentioned, the root/empty login was only for NXG images where NIAuth was never present. I would be okay with adding this temporarily if we just want to make the image usable until we have NIAuth support. However, I think we should definitely remove it afterwards. We don't want an image with two different insecure logins. The root/empty login would be new, currently undocumented, and would not be changeable through the normal MAX workflow.

shruthi-ravi:
Sorry, I had tried out a few experiments and missed following up here.

> The root/empty login would be new, currently undocumented, and would not be changeable through the normal MAX workflow.

That's right. I spun up the password management tool on the web server for a target and was not able to modify the root credentials in that context. I was able to modify the root credentials only on the command line, which is also documented as one of the methods to change password though.

> What happens if a user changes the admin password via sysapi? Does it only apply to the NIAuth database? And will the shadow fallthrough allow you to still login with no password, even though you set one in NIAuth?

The admin password change applies only to the NIAuth database. You can still login as root with no password.

Given our current focus on os-common and the progress we are making, we might have the changes enabling admin login moved to the ni-auth IPK even before this change gets checked in, and I can probably abandon this PR.

amstewart (approved):
@shruthi-ravi 
I'll defer to you as to whether it is worth pulling this PR now, or waiting for the NI-auth IPK to become functional.

I'm not super happy with decoupling the root/admin passwords. But as long as we are sure that we'll revert this change once NI-Auth works, I suppose it's fine for now.

shruthi-ravi:
Closing this PR until we find this change necessary, given the niauth changes going into the image soon.

PR #82, opened 2021-01-07
recipe-ni: add nisdbootconfig-safe and nisdbootconfig to configure boot from sd card

NIwkoe:
Artemis uses sd card as its main disk instead of NAND disk (ubifs). Some configurations are required for it to boot properly. These configurations are mount points and fw_printenv configuration. The implementation for both safemode and runmode is similar, which is to install the script and invoke it earlier than mountall.sh so that the SD card partitions can be mounted properly and fw_printenv can function correctly with the right path to u-boot's environmental variable.

We can assume the arm target uses an sd card by checking the absence of mtd partition. Since this configuration is only needed on first boot, the script will remove its link once it has been executed.
@ni/rtos

### DIFF
--- a/recipes-ni/nisdbootconfig/files/nisdbootconfig
+++ b/recipes-ni/nisdbootconfig/files/nisdbootconfig
@@ -1,0 +1,26 @@
+#!/bin/sh
+set -e
+
+# check if target is NOT using ubifs (mtd)
+if ! cat /proc/mtd | grep -qs mtd ; then
+
+    # configure the partitions to mount by label in fstab
+    sed -i '/ubifs/d' /etc/fstab
+    echo LABEL=nibootfs         /boot                ext4       sync           0  0 >>/etc/fstab
+    echo LABEL=niconfig         /etc/natinst/share   ext4       sync           0  0 >>/etc/fstab
+    mkdir -p /etc/natinst/share
+
+    # configure the nirootfs partition if we are in safemode
+    if [ -f /etc/natinst/safemode ]; then
+        echo LABEL=nirootfs         /mnt/userfs          ext4       defaults       0  0 >>/etc/fstab
+        mkdir -p /mnt/userfs
+    fi
+
+    # configure the path for fw_printenv to read environmental variables from u-boot
+    # replace /dev/mtd4 and /dev/mtd5 with /boot/uboot/uboot.env
+    sed -i '/mtd/d' /etc/fw_env.config
+    echo /boot/uboot/uboot.env         0x0000         0x20000 >>/etc/fw_env.config
+fi
+
+# we don't need this to run all the time, remove after first boot
+update-rc.d -f nisdbootconfig remove
--- a/recipes-ni/nisdbootconfig/nisdbootconfig.bb
+++ b/recipes-ni/nisdbootconfig/nisdbootconfig.bb
@@ -1,0 +1,26 @@
+SUMMARY = "NI SDboot Config"
+DESCRIPTION = "Configuration script for arm target booting from SD Card"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+SECTION = "base"
+
+inherit update-rc.d
+
+SRC_URI = "file://nisdbootconfig \
+"
+
+INITSCRIPT_NAME = "nisdbootconfig"
+INITSCRIPT_PARAMS = "start 00 S ."
+
+S = "${WORKDIR}"
+
+do_install () {
+	install -d ${D}${sysconfdir}/init.d/
+
+	# from artemis onwards, using sd card as main disk for arm target is introduced
+	# install nisdbootconfig to configure sd card booting requirement
+	if [ "${TARGET_ARCH}" = "arm" ]; then
+		install -m 0755 ${WORKDIR}/nisdbootconfig ${D}${sysconfdir}/init.d
+	fi
+
+}
--- a/recipes-core/packagegroups/packagegroup-ni-runmode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-runmode.bb
@@ -23,6 +23,7 @@ RDEPENDS_${PN} = "\
 	libfmi-dev \
 	librtpi \
 	lldpd \
+	nisdbootconfig \
 	niwatchdogpet \
 	parted \
 	rtctl \
--- a/recipes-core/packagegroups/packagegroup-ni-safemode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-safemode.bb
@@ -15,4 +15,5 @@ RDEPENDS_${PN} = " \
 	e2fsprogs-e2fsck \
 	e2fsprogs-mke2fs \
 	e2fsprogs-tune2fs \
+	nisdbootconfig \
 "


recipes-ni/nisdbootconfig/files/nisdbootconfig
```diff
@@ -0,0 +1,20 @@
...
+if ! cat /proc/mtd | grep -qs mtd ; then
+
+    # configure the partitions to mount by label in fstab
+    sed -i '/ubi/d' /etc/fstab
+    echo LABEL=nibootfs         /boot                ext4       sync           0  0 >>/etc/fstab
+    echo LABEL=niconfig         /etc/natinst/share   ext4       sync           0  0 >>/etc/fstab
+    mkdir -p /etc/natinst/share
+
+    # configure the path for fw_printenv to read environmental variables from u-boot
+    sed -i '/mtd/d' /etc/fw_env.config
```
amstewart, line 21: For my understanding, could you provide examples of what lines you expect to be deleted by this sed command?

recipes-ni/nisdbootconfig/files/nisdbootconfig
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+# check if target is NOT using ubifs (mtd)
+if ! cat /proc/mtd | grep -qs mtd ; then
+
+    # configure the partitions to mount by label in fstab
+    sed -i '/ubi/d' /etc/fstab
+    echo LABEL=nibootfs         /boot                ext4       sync           0  0 >>/etc/fstab
+    echo LABEL=niconfig         /etc/natinst/share   ext4       sync           0  0 >>/etc/fstab
```
amstewart: I assume the intention here is to remove entries from `/etc/fstab` which are of the type `ubi` and replace them with the `ext4` variants. Could you add some whitespace characters to the regex match to make it more precise.

eg. `sed -i '/\W\+ubi\W\+/d' /etc/fstab`

recipes-ni/nisdbootconfig/nisdbootconfig.bb
```diff
@@ -0,0 +1,24 @@
+SUMMARY = "NI SDboot Config"
+DESCRIPTION = "Bios helper support files"
```
amstewart: Could you make this package description more specific. This recipe isn't generalized to all BIOS support files.

Also NIT: `BIOS` should have every letter capitalized. :)

amstewart (changes requested):

recipes-ni/initscripts-nilrt/files/nisdbootconfig-safe
```diff
@@ -0,0 +1,22 @@
...
+
+    # configure the path for fw_printenv to read environmental variables from u-boot
+    sed -i '/mtd/d' /etc/fw_env.config
+    echo /boot/uboot/uboot.env         0x0000         0x20000 >>/etc/fw_env.config
+fi
+
+# we don't need this to run all the time, remove after first boot
+update-rc.d -f nisdbootconfig-safe remove
+
+exit 0
```
amstewart: I'm not sure there's any reason to squash the exit code here. Since this script is responsible for bootstrapping the disks, any failure in the script should at least be reported.

I would recommend removing the `exit` call  and adding `set -e` to the top of the script.

recipes-ni/initscripts-nilrt/initscripts-nilrt-safemode_1.0.bb
```diff
@@ -9,6 +9,7 @@ DEPENDS += "shadow-native pseudo-native update-rc.d-native niacctbase"
 RDEPENDS_${PN} += "bash niacctbase update-rc.d"
 
 SRC_URI = " \
+                      file://nisdbootconfig-safe \
```
amstewart: NIT: Did you mean to have this much whitespace?

recipes-ni/initscripts-nilrt/files/nisdbootconfig-safe
```diff
@@ -0,0 +1,22 @@
...
+# check if target is NOT using ubifs (mtd)
+if ! cat /proc/mtd | grep -qs mtd ; then
+
+    # configure the partitions to mount by label in fstab
+    sed -i '/ubi/d' /etc/fstab
+    echo LABEL=nibootfs         /boot                ext4       sync           0  0 >>/etc/fstab
+    echo LABEL=niconfig         /etc/natinst/share   ext4       sync           0  0 >>/etc/fstab
+    echo LABEL=nirootfs         /mnt/userfs          ext4       defaults       0  0 >>/etc/fstab
+    mkdir -p /etc/natinst/share
+    mkdir -p /mnt/userfs
```
amstewart: Is the only difference between the safemode and runmode version of this script: whether or not the `nirootfs` partition is switched to `ext4`?

If that's the case, can we just have one version of the script for both, which checks to see if its in safemode and, if it is, fixes the rootfs entry. Something like

```bash
if [ -r '/etc/natinst/safemode' ]; then
    echo LABEL=nirootfs...
    mkdir -p /mnt/userfs
fi
```

Then we only have to maintain a single package and source for the script.

amstewart (changes requested):
Seems like these scripts could be collapsed into a single package recipe. Do you think that's feasible?

recipes-ni/nisdbootconfig/files/nisdbootconfig
NIwkoe: Here's the default values for fw_env.config:
/dev/mtd4		0x0000		0x20000		0x20000
/dev/mtd5		0x0000		0x20000		0x20000

I'll add more description in the comments to note this down.

recipes-ni/nisdbootconfig/files/nisdbootconfig
NIwkoe: I tried /\W\+ubi\W\+/d and it did not work as expected. The original string looks like this:
ubi0:bootfs          /boot                ubifs      noatime,sync          0  0

I'll change it to '/ubifs/d' instead to be more specific

recipes-ni/initscripts-nilrt/files/nisdbootconfig-safe
NIwkoe: exit is removed and added set -e at top of script to capture non-zero exits.

recipes-ni/initscripts-nilrt/files/nisdbootconfig-safe
NIwkoe: Yeap, that's the only difference. I think this is a great idea. I will remove the one I dedicated for safe mode (nisdbootconfig-safe) and just maintain one version in the new recipe (nisdbootconfig), then use the safemode and runmode recipes to include this nisdbootconfig recipe.

recipes-ni/nisdbootconfig/nisdbootconfig.bb
NIwkoe: Added the correct description.

recipes-ni/initscripts-nilrt/initscripts-nilrt-safemode_1.0.bb
NIwkoe: I mimicked the format of the following lines that contains the same amount of whitespaces:
SRC_URI_append_x64 = " \
                      file://niselectnetnaming \
"

Seems like this is not the way. Anyway this change will be reverted as I will be removing the nisdbootconfig-safe. Refer to comment below on just using one script.

recipes-ni/nisdbootconfig/files/nisdbootconfig
amstewart: Well, at the least, you need to escape the `+` control character. But it's also possible that the busybox version of `sed` we have installed doesn't interpret the string the same way (or support the `\W` metacharacter).

I guess I'm fine with changing the match to `ubifs`, so long as that is reasonably unlikely to occur in the fstab in any other place.

amstewart (approved):
Looks good to me!

shruthi-ravi:
Hi Wilson, can you do a quick sanity build for x64 and make sure that the image doesn't have this file? Otherwise, looks good.

NIwkoe:
@shruthi-ravi I've built zynq, zynq_safe, x64 and x64_safe and checked them. Only zynq and zynq_safe contains nisdbootconfig in /etc/init.d and /etc/rcS.d. The x64 targets do not have the file in their outputs.

NIwkoe:
> Hi Wilson, can you do a quick sanity build for x64 and make sure that the image doesn't have this file? Otherwise, looks good.

@shruthi-ravi I've built zynq, zynq_safe, x64 and x64_safe and checked them. Only zynq and zynq_safe contains nisdbootconfig in /etc/init.d and /etc/rcS.d. The x64 targets do not have the file in their outputs.

shruthi-ravi (approved):

PR #83, opened 2021-01-08
xilinx-zynqhf.conf: add artemis device code (7AAE)

NIwkoe:
The artemis project uses the device code 7AAE. Adding the device code
to the list so that the device tree can be recognized and included.

This commit adds "7AAE" to the "NILRT_ARM_DEVICE_CODES" list in
xilinx-zynqhf.conf.

@ni/rtos 


amstewart (approved):
Looks good to me. I like that the lines are the same length now. :)

Do you ACK @shruthi-ravi ?

shruthi-ravi (approved):

PR #84, opened 2021-01-11
Move files from os-common

chaitu236:
Move nisetreboottype, xfce files from os-common

@ni/rtos

### DIFF
--- a/recipes-ni/initscripts-nilrt/files/nisetreboottype
+++ b/recipes-ni/initscripts-nilrt/files/nisetreboottype
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# The purpose of this script is to set the reboot type of the controller based
+# on the value of ni_boot_mode_input. Reboots to install mode require a cold
+# reboot. All other modes should use the default reboot type
+
+NI_REQUESTED_REBOOT_TYPE_FILE=/sys/kernel/ni_requested_reboot_type
+BOOT_MODE_INPUT=/tmp/ni_boot_mode_input
+
+# If we're on cRIO and booting to install mode, request a cold reboot
+if [ "`/sbin/fw_printenv -n TargetClass`" = "cRIO" ]; then
+	if [ -f $NI_REQUESTED_REBOOT_TYPE_FILE ]; then
+		if [ -f $BOOT_MODE_INPUT ]; then
+			read NEXT_BOOT_MODE < $BOOT_MODE_INPUT
+			if [ -n "$NEXT_BOOT_MODE" ]; then
+				case "$NEXT_BOOT_MODE" in
+				install)
+					echo 1 > $NI_REQUESTED_REBOOT_TYPE_FILE
+					;;
+				esac
+			fi
+		fi
+	fi
+fi
--- a/recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
+++ b/recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
@@ -23,7 +23,16 @@ do
   echo 1 > $x
 done 2> /dev/null
 
-# Set affinity for running IRQ threads to Core-0
-grep -l irq/ /proc/*/comm | cut -d/ -f3 | xargs -n1 taskset -p 1 >/dev/null
+# Set affinity for running IRQ threads to Core-0 immediately.
+# Setting /proc/irq/default_smp_affinity and /proc/irq/*/smp_affinity as done above will
+# eventually cause the kernel to change the affinity for these threads automatically, but
+# not until the next time those threads are scheduled. We want to remove the possibility
+# of that happening during jitter-sensitive code in a user application.
+PIDS=`grep -l irq/ /proc/*/comm 2>/dev/null | cut -d/ -f3`
+if [ -z "$PIDS" ] ; then
+  echo $0: Warning: No IRQ threads found.
+  exit 0
+fi
+echo $PIDS | xargs -n1 taskset -p 1 >/dev/null 2>&1
 
 [ "${VERBOSE}" != "no" ] && echo "done"
--- a/recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
+++ b/recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
@@ -3,6 +3,13 @@
 # All rights reserved.
 [ "${VERBOSE}" != "no" ] && echo -n "Starting nisetupkernelconfig:"
 
+#Create core dump files in the root directory of runmode with the
+#executable name as the suffix.
+echo "/var/local/natinst/log/core_dump.%E" > /proc/sys/kernel/core_pattern
+
+#Enable core dump for tainted binaries
+echo 1 > /proc/sys/fs/suid_dumpable
+
 #This disables RT scheduler's CPU throttling for National Instruments LabVIEW Real-Time
 echo -1 > /proc/sys/kernel/sched_rt_runtime_us
 
--- a/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
+++ b/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
@@ -40,6 +40,7 @@ SRC_URI_append_x64 = "file://nidisablecstates \
                       file://nicheckbiosconfig \
                       file://niopendisks \
                       file://niclosedisks \
+                      file://nisetreboottype \
                       file://test-niopendisks-init \
                       file://test-niclosedisks-init \
 "
@@ -71,6 +72,7 @@ do_install () {
 	install -m 0755 ${WORKDIR}/nipopulateconfigdir   ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/nisetupkernelconfig   ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/nisetcommitratio      ${D}${sysconfdir}/init.d
+	install -m 0755 ${WORKDIR}/nisetreboottype       ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/wirelesssetdomain     ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/cleanvarcache         ${D}${sysconfdir}/init.d
 	install -m 0755 ${S}/nisetupirqpriority          ${D}${sysconfdir}/init.d
@@ -89,6 +91,8 @@ do_install () {
 	update-rc.d -r ${D} nisetcommitratio      start 99 S .
 	update-rc.d -r ${D} wirelesssetdomain     start 36 S .
 	update-rc.d -r ${D} cleanvarcache         start 38 0 6 S .
+
+	update-rc.d -r ${D} nisetreboottype       stop  55 6 .
 
 	# only for nilrt-nxg, on older nilrt it's installed via p4
 	if ${@oe.utils.conditional('DISTRO', 'nilrt-nxg', 'true', 'false', d)}; then
--- a/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
+++ b/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
@@ -91,7 +91,7 @@ do_install () {
 	update-rc.d -r ${D} nicreatecpusets       start 1  4 5 .
 	update-rc.d -r ${D} nicreatecpuacctgroups start 2  4 5 .
 	update-rc.d -r ${D} nipopulateconfigdir   start 36 S .
-	update-rc.d -r ${D} nisetupkernelconfig   start 3  4 5 .
+	update-rc.d -r ${D} nisetupkernelconfig   start 3  5 .
 	update-rc.d -r ${D} nisetcommitratio      start 99 S .
 	update-rc.d -r ${D} wirelesssetdomain     start 36 S .
 	update-rc.d -r ${D} cleanvarcache         start 38 0 6 S .
--- a/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
+++ b/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
@@ -34,6 +34,7 @@ SRC_URI = "file://nisetbootmode \
            file://nipopulateconfigdir \
            file://run-ptest \
            file://cleanvarcache \
+           file://modules_autoload \
            file://niinitled \
 "
 
@@ -76,6 +77,7 @@ do_install () {
 	install -m 0755 ${WORKDIR}/nisetreboottype       ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/wirelesssetdomain     ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/cleanvarcache         ${D}${sysconfdir}/init.d
+	install -m 0755 ${WORKDIR}/modules_autoload      ${D}${sysconfdir}/init.d
 	install -m 0755 ${S}/nisetupirqpriority          ${D}${sysconfdir}/init.d
 	install -m 0755 ${S}/niinitled                   ${D}${sysconfdir}/init.d
 
@@ -93,6 +95,7 @@ do_install () {
 	update-rc.d -r ${D} nisetcommitratio      start 99 S .
 	update-rc.d -r ${D} wirelesssetdomain     start 36 S .
 	update-rc.d -r ${D} cleanvarcache         start 38 0 6 S .
+	update-rc.d -r ${D} modules_autoload      start 37 S .
 	update-rc.d -r ${D} niinitled             start 40 S .
 
 	update-rc.d -r ${D} nisetreboottype       stop  55 6 .
--- a/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
+++ b/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
@@ -34,6 +34,7 @@ SRC_URI = "file://nisetbootmode \
            file://nipopulateconfigdir \
            file://run-ptest \
            file://cleanvarcache \
+           file://niinitled \
 "
 
 SRC_URI_append_x64 = "file://nidisablecstates \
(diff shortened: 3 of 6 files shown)


recipes-xfce/xfce-nilrt-settings/xfce-nilrt-settings_1.0.bb
```diff
@@ -65,6 +66,7 @@ FILES_${PN} = "${confdir}/autostart/dpms.desktop \
 	    ${confdir}/xfce4/xfconf/xfce-perchannel-xml/xfwm4.xml \
 	    ${confdir}/xfce4/xfconf/xfce-perchannel-xml/xsettings.xml \
 	    /usr/local/natinst/bin/dual-monitor-setup.sh \
+	    /etc/default/xserver-xfce \
```
shruthi-ravi: Here's another [recipe](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-xfce/xserver-xfce-init/xserver-xfce-init_1.0.bb#L27) that adds this file. I think you can merge the contents of this default file to that file. I think that would be the part where we read from the nirtcfg settings to check if UI is enabled.

recipes-ni/initscripts-nilrt/files/mountconfig
```diff
@@ -0,0 +1,58 @@
...
+mount_config()
+{
+	# Don't mount if already mounted
+	if ! grep -q " $mount_point " /proc/self/mounts; then
+		if cat /sys/class/mtd/mtd*/name 2>/dev/null | grep -q ^$partition_label\$
+		then
+			# If an mtd partition has this partition label, use
+			# mount_ubi_volume
+			mount_ubi_volume "$partition_label" "$volume_label" \
+				"$volume_number" "$mount_point" "$mount_args"
```
amstewart: The `config` partition only exists on ARM targets, iirc. So I don't think this initscript is relevant to dunfell images and beyond. It can be dropped.

recipes-ni/initscripts-nilrt/files/nicreatelogdir
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (c) 2012-2013 National Instruments.
+# All rights reserved.
+#
+# Ensure the log directory is always present with the correct
+# ownership.
+
+log_dir="/var/local/natinst/log"
+
+mkdir -p -m 777 $log_dir
+chown lvuser:ni $log_dir
```
amstewart: Do we really need an initscript for this? Surely there is either (a) already another initscript which creates NI-owned directories or (b) we can do what every other distro does and expect that log-creating packages will make their own directories.

recipes-ni/initscripts-nilrt/files/nisetreboottype
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# The purpose of this script is to set the reboot type of the controller based
+# on the value of ni_boot_mode_input. Reboots to install mode require a cold
+# reboot. All other modes should use the default reboot type
```
amstewart, line 5: I don't know if this is a relevant workflow; it seems like the remnants of the obsolete "install mode" workflow. Can you check if SCS or anyone else even sets `/sys/kernel/ni_requested_reboot_type` to `install`? If not, we can drop this whole file.

recipes-ni/initscripts-nilrt/files/mountconfig
```diff
@@ -0,0 +1,58 @@
...
+# Bail out on first error
+set -o errexit
+
+mount_config()
+{
+	# Don't mount if already mounted
+	if ! grep -q " $mount_point " /proc/self/mounts; then
+		if cat /sys/class/mtd/mtd*/name 2>/dev/null | grep -q ^$partition_label\$
+		then
+			# If an mtd partition has this partition label, use
```
shruthi-ravi: The entries that refer to mtd and mount_ubi_volume are Zynq-specific and can be removed. The mount command on line 33 is enough. Do you think it's a good idea to edit these files as you move them or should it be done in the next update? Can you create a task for it or something if you pick the latter option?

recipes-ni/initscripts-nilrt/files/nisetupdump
```diff
@@ -0,0 +1,18 @@
...
+#Create core dump files in the root directory of runmode with the
+#executable name as the suffix.
+if [ -f $safemodeflag ]; then
+	echo "/mnt/userfs/var/local/natinst/log/core_dump.%E" > /proc/sys/kernel/core_pattern
+else
+	echo "/var/local/natinst/log/core_dump.%E" > /proc/sys/kernel/core_pattern
+fi
+
+#Enable core dump for tainted binaries
+echo 1 > /proc/sys/fs/suid_dumpable
```
amstewart: We need to keep these settings*, but the rest of this script is only relevant to safemode. Surely we already have a package in OE that sets `ulimit` or something? That's probably a better place to put sysctl options like this.

* Though the `core_pattern` should be set to a better location. Where should that be @gratian ?

recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
```diff
@@ -78,6 +79,8 @@ do_install () {
 	install -m 0755 ${WORKDIR}/mountconfig           ${D}${sysconfdir}/init.d
 	# script for auto-loading kernel modules (needed for modules that are not auto-loaded by udev)
 	install -m 0755 ${WORKDIR}/modules_autoload      ${D}${sysconfdir}/init.d
+	# nicreatelogdir script that ensures log directory is present with proper ownership
```
shruthi-ravi: Description feels redundant.

recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
```diff
@@ -85,10 +98,16 @@ do_install () {
 	update-rc.d -r ${D} nicreatecpusets       start 1  4 5 .
 	update-rc.d -r ${D} nicreatecpuacctgroups start 2  4 5 .
 	update-rc.d -r ${D} nipopulateconfigdir   start 36 S .
-	update-rc.d -r ${D} nisetupkernelconfig   start 3  4 5 .
+	update-rc.d -r ${D} nisetupkernelconfig   start 3  5 .
```
amstewart, line 94: This script will eventually need to be on RL 4 as well, but we can do that in the later RAUC story to configure RL 4 as the new "safemode".

recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
```diff
@@ -78,6 +79,8 @@ do_install () {
 	install -m 0755 ${WORKDIR}/mountconfig           ${D}${sysconfdir}/init.d
 	# script for auto-loading kernel modules (needed for modules that are not auto-loaded by udev)
 	install -m 0755 ${WORKDIR}/modules_autoload      ${D}${sysconfdir}/init.d
+	# nicreatelogdir script that ensures log directory is present with proper ownership
+	install -m 0755 ${WORKDIR}/nicreatelogdir        ${D}${sysconfdir}/init.d
```
shruthi-ravi: This [recipe](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-support/syslog-ng/syslog-ng_3.%25.bbappend#L25) seems to be creating the same directory.

recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
```diff
@@ -75,6 +76,8 @@ do_install () {
 	install -m 0755 ${WORKDIR}/wirelesssetdomain     ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/cleanvarcache         ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/mountconfig           ${D}${sysconfdir}/init.d
+	# script for auto-loading kernel modules (needed for modules that are not auto-loaded by udev)
```
shruthi-ravi: I think it would be more appropriate to move this description to the script.

recipes-xfce/xfce-nilrt-settings/files/xserver-xfce.default
```diff
@@ -0,0 +1,10 @@
+if [ "$1" = "start" ] && \
+	/usr/local/natinst/bin/nirtcfg --get section=SYSTEMSETTINGS,token=ui.enabled,value=False \
+	| grep -iq false
+then
+	echo "$0: Disabled in system configuration." >&2
+	exit 0
+fi
+
+XUSER=lvuser
+HOME=/home/lvuser
```
amstewart: Seems like this file already exists in [xserver-xfce-init-1.0](https://github.com/ni/meta-nilrt/blob/29fbbd49b52e84de6fb1b3c18702b743d2124420/recipes-xfce/xserver-xfce-init/xserver-xfce-init-1.0/xserver-xfce.conf) and is installed to the same place [here](https://github.com/ni/meta-nilrt/blob/29fbbd49b52e84de6fb1b3c18702b743d2124420/recipes-xfce/xserver-xfce-init/xserver-xfce-init_1.0.bb#L27).

amstewart (changes requested):

shruthi-ravi (changes requested):
Part 1..I still have the last 3 commits to look at.

shruthi-ravi:
Well, looks like I was a minute late :) @amstewart has similar concerns.

recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
```diff
@@ -16,7 +16,16 @@ do
...
 
-# Set affinity for running IRQ threads to Core-0
-grep -l irq/ /proc/*/comm | cut -d/ -f3 | xargs -n1 taskset -p 1 >/dev/null
+# Set affinity for running IRQ threads to Core-0 immediately.
+# Setting /proc/irq/default_smp_affinity and /proc/irq/*/smp_affinity as done above will
+# eventually cause the kernel to change the affinity for these threads automatically, but
+# not until the next time those threads are scheduled. We want to remove the possibility
+# of that happening during jitter-sensitive code in a user application.
+PIDS=`grep -l irq/ /proc/*/comm 2>/dev/null | cut -d/ -f3`
+if [ -z "$PIDS" ] ; then
```
shruthi-ravi, line 32: Should this really be an error? I don't understand why. This seems to be the only difference between existing code and new code, other than the comment itself?

recipes-ni/initscripts-nilrt/files/nicreatelogdir
chaitu236: It does seem excessive to have an initscript just for this. But I [haven't been able to find](https://codesearch.natinst.com/search/text?q=mkdir.*natin%20repo%3Ameta-nilrt&fold_case=false&regex=true&context=true) another initscript that creates NI dirs.

As for (b), I suspect the reason this was added was to support some old piece of software that didn't creates its own dir. Do you think we should remove this to find all such cases and have their devs fix them?

recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
amstewart: Yeah; I agree that this doesn't imply an error.

recipes-ni/initscripts-nilrt/files/nicreatelogdir
amstewart: >Do you think we should remove this to find all such cases and have their devs fix them?

It [seems like](https://codesearch.natinst.com/search/text?q=local%2Fnatinst%2Flog&fold_case=auto&regex=true&context=true) that path is used in a lot of different packages. In all cases, it seems like the intention is just to give the `lvuser` user a location to access core dumps, LV crash logs, ect. (which implies that it is where LV customers are intended to store their logs).

It's probably too risky to just remove it from the base image and require the NIFeeds package owners to fix all their packages. But I might like to move this code to something like the [`niacctbase`](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-core/niacctbase/niacctbase.bb) package, which does generic setup for the `lvuser` account.

shruthi-ravi: I found a [recipe](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-support/syslog-ng/syslog-ng_3.%25.bbappend#L25) that creates the directory too.

recipes-ni/initscripts-nilrt/files/nisetupdump
chaitu236: There is [this](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-ni/base-files-nilrt/files/ulimit.sh#L1) but it gets installed to /etc/profile.d, which is probably not where we want to add this.
How about [nisetupkernelconfig](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-ni/initscripts-nilrt/files/nisetupkernelconfig#L1)?

recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
chaitu236: Should I change the msg to "No IRQ threads found" and exit 0? [Ref](https://review-board.natinst.com/r/138441/)

chaitu236: @billpittman What do you think?

chaitu236: Ok I just tried the original and don't see an error when there are no irq threads - so not sure why we added it.

recipes-ni/initscripts-nilrt/files/mountconfig
chaitu236: Removed the file altogether as remaining changes aren't necessary. I can edit them as I go.

recipes-ni/initscripts-nilrt/files/nisetreboottype
chaitu236: Found [this](https://p4.natinst.com/browser/perforce/MAX/components/remote/softwareinstall/trunk/21.0/source/SoftwareActionShared/Imaging.cpp?revision_at=51501451#line_388)... so keep this around?

amstewart: I... don't know what that's supposed to mean? Does that codepath run when `reboot_type` is set to `install`?

chaitu236: The codepath runs when ni_boot_mode_input is set to install and [this](https://p4.natinst.com/browser/perforce/MAX/components/remote/softwareinstall/trunk/21.0/source/SoftwareActionShared/Imaging.cpp?revision_at=51501451#line_388) does that.

recipes-ni/initscripts-nilrt/files/nisetupdump
chaitu236: Adding core_pattern also to nisetupkernelconfig for now

amstewart:
@chaitu236 Can you post a V2 summary, so I know what you changed.

chaitu236:
> @chaitu236 Can you post a V2 summary, so I know what you changed.

Addressed all feedback except for the one in nisetupkernelconfig.

Removed mountconfig, nicreatelogdir, xserver-xfce.default
Moved nisetupdump  stuff into nisetupkernelconfig

recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
shruthi-ravi: Ok, I think we can keep the original file with the updated comment unless @gratian objects.

recipes-xfce/xserver-xfce-init/xserver-xfce-init-1.0/xserver-xfce.conf
```diff
@@ -1,2 +1,10 @@
+if [ "$1" = "start" ] && \
+	/usr/local/natinst/bin/nirtcfg --get section=SYSTEMSETTINGS,token=ui.enabled,value=False \
+	| grep -iq false
+then
+	echo "$0: Disabled in system configuration." >&2
+	exit 0
```
shruthi-ravi, line 6: Does this have the effect of exiting from the parent xserver-xfce process too? I assume that is the intent given the message.

shruthi-ravi (commented):
lgtm, pending the two outstanding issues.

recipes-xfce/xserver-xfce-init/xserver-xfce-init-1.0/xserver-xfce.conf
gratian: The intent of the original code was to not start the X stack (X server and Xfce desktop environment) if the "Enable Embedded UI" checkbox in MAX is not checked.

recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
gratian: I'll defer to @billpittman since he's the original author. My guess is the error is there to catch a case where this script is not working e.g. the kernel changed the names of irq threads or we misconfigured the kernel and interrupts are not threaded.

recipes-ni/initscripts-nilrt/files/nisetupdump
gratian: Yeah, I think moving the core_pattern  into nisetupkernelconfig is OK.

recipes-xfce/xserver-xfce-init/xserver-xfce-init-1.0/xserver-xfce.conf
shruthi-ravi: Right. Since this file is invoked from [here](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-xfce/xserver-xfce-init/xserver-xfce-init-1.0/xserver-xfce#L8), I was wondering if this exit would cause the parent process to also finish or would it continue on to line 26, which does not seem to be the intent.

recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
billpittman: Definitely does not need to be an error and I think Gratian instincts about the error being there in the first place are correct.  You could probably make a case for removing it all together seeing as it will be corrected eventually and with this running at startup (and the suggested practice of warm-up iterations), there are multiple opportunities for the threads/IRQs to be scheduled/re-scheduled.

chaitu236: Changed error to warning and exit 0

chaitu236:
v3 changes: Changed nisetupkernelconfig's error to warning and exit 0

amstewart (approved):

recipes-xfce/xserver-xfce-init/xserver-xfce-init-1.0/xserver-xfce.conf
shruthi-ravi: Do you have an update on this @chaitu236?

chaitu236: Oh sorry I misunderstood the comment. Yes the exit would cause the parent to also exit and it wouldn't continue to line 26.

shruthi-ravi (approved):

amstewart (approved):

PR #86, opened 2021-01-14
packagegroup-ni-extra: Add the kernel performance tests to the extra feed

gratian:
Build and publish the kernel performance tests in the extra feed for
wider consumption.


amstewart (commented):
Do you want to put these in the [desirables packagegroup](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-core/packagegroups/packagegroup-ni-desirable.bb)?

gratian:
Yeah, it's not a bad idea, since other groups might use it. I'll move it.

PR #87, opened 2021-01-14
packagegroup-ni-extra: Add the kernel performance tests to the extra feed

gratian:
Build and publish the kernel performance tests in the extra feed for
wider consumption.


gratian:
Closing in favor of including this package in the desirable package group instead (new pull request).

PR #90, opened 2021-01-15
initscripts-nilrt: Move nisetembeddeduixml into OE

chaitu236:
Move nisetembeddeduixml from os-common into OE

I wasn't sure if this should instead go here or into xserver-xfce-init.

The script was originally intended to work on both safemode and runmode so adding it here where it'd end up in rauc bundle makes sense if we want MAX experience to be same - xserver-xfce-init wouldn't presumably in rauc bundle so adding it there would change MAX experience.
But there are other scripts like this on os-common (e.g., niselectsystemsettings) which can have a better home than initscripts if MAX experience need not be exactly same. Thoughts?

@ni/rtos

### DIFF
--- a/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
+++ b/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
@@ -37,6 +37,7 @@ SRC_URI = "file://nisetbootmode \
 "
 
 SRC_URI_append_x64 = "file://nidisablecstates \
+                      file://nisetembeddeduixml \
                       file://nicheckbiosconfig \
                       file://niopendisks \
                       file://niclosedisks \
@@ -65,6 +66,7 @@ do_install () {
 	fi
 
 	install -m 0755 ${WORKDIR}/mountdebugfs          ${D}${sysconfdir}/init.d
+	install -m 0755 ${WORKDIR}/nisetembeddeduixml    ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/nicleanstalelinks     ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/nicreatecpusets       ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/nicreatecpuacctgroups ${D}${sysconfdir}/init.d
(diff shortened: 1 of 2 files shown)


recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
```diff
@@ -81,6 +83,7 @@ do_install () {
 	update-rc.d -r ${D} nisetbootmode         start 80 S . stop 0 0 6 .
 	update-rc.d -r ${D} firewall              start 39 S .
 	update-rc.d -r ${D} mountdebugfs          start 82 S .
+	update-rc.d -r ${D} nisetembeddeduixml    start 20 5 .
```
amstewart, line 86: Are you missing [the stop orders](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Frtos%2Frtoslegacyd%2FNIOE_Dunfell_x64_temp%2Fx64_dkms_dunfell_nilrt.mak&version=GC885fcbb1263c692781eb9e1f42250da94e6480a3&line=281&lineEnd=282&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) here?

amstewart (changes requested):
>But there are other scripts like this on os-common (e.g., niselectsystemsettings) which can have a better home than initscripts if MAX experience need not be exactly same. Thoughts?

I think, for now, we'd like to try and avoid changing the MAX behavior.

recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
chaitu236: The script does nothing in stop, so its unnecessary right?

amstewart: Hmm... yeah; you're right. Weird that we had a stop sequence set. Maybe the script did something on shutdown at some point in the past.

amstewart (approved):

recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
chaitu236: Looks like it was that way [from the beginning](https://review-board.natinst.com/r/227548/) - probably a copy paste issue.

PR #91, opened 2021-01-15
nisdbootconfig: Fixup commit 801d331 breaking x64 images.

shruthi-ravi:
Commit 801d331 added nisdbootconfig to x64 images too. The recipe
inherits from update-rc.d but does not install any file to
/etc/init.d on x64 targets. This causes the post-inst of the package
to fail on x64 targets when the following command is run:
update-rc.d $OPT nisdbootconfig start 00 S .
because the nisdbootconfig file does not exist on that image.

To fix this, add the package to only ARM images in
packagegroup-ni-base since the package is common to both safemode and
runmode.

Testing: Still building the packagegroup-ni-base recipe. Will update when it is done.

@ni/rtos, fyi @NIwkoe

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-base.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-base.bb
@@ -39,6 +39,7 @@ NILRT_NXG_x64_PACKAGES = "\
 NILRT_ARM_PACKAGES = "\
 	jitterentropy-rngd \
 	mtd-utils-ubifs \
+	nisdbootconfig \
 "
 
 NILRT_x64_PACKAGES = "\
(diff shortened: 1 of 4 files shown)


amstewart (approved):

shruthi-ravi:
The recipe as it is builds for x64 targets too and as a result, would be available in the feeds for download. Updated the recipe to be compatible only with ARM builds. Still working on that packagegroup-ni-base build for ARM..

shruthi-ravi:
Built packagegroup-ni-base for ARM and x64 locally. No issues.

amstewart (approved):

PR #92, opened 2021-01-16
packagegroup-ni-desirable: Fixup commit 8221f77, breaks ARM builds

shruthi-ravi:
The kernel-performance-tests package depends on the fw-printenv
package that is marked [compatible](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-ni/ni-utils/fw-printenv_1.0.bb#L14) only for x64 builds, thus breaking
ARM builds. As a quick workaround, set RDEPENDS on the tests package
only during x64 builds.
@gratian I'm not sure if this is the right change at all. I don't know if these
tests were intended to run on ARM targets. The alternative is to change the 
fw-printenv dependency for ARM builds. This is blocking a good SystemImage
build, so I thought I'll get a quick change out to start the conversation. Lmk
what you think.

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-desirable.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-desirable.bb
@@ -16,6 +16,7 @@ RDEPENDS_${PN}_append_x64 = "\
 	packagegroup-ni-next-kernel \
 	packagegroup-ni-mono-extra \
 	florence \
+	kernel-performance-tests \
 "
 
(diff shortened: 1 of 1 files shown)


gratian:
@shruthi-ravi sorry I had a fix for this in works but got distracted and just posted it now: https://github.com/ni/meta-nilrt/pull/93

This should fix it for both ARM and x64.

amstewart:
I merged PR #93 ; so I think this PR can be closed.

PR #93, opened 2021-01-16
kernel-performance-tests: Fix runtime dependency on fw-printenv

gratian:
The 'fw_printenv' binary is provided by different packages on
different architectures:

  * 'fw-printenv' package on x86_x64 targets.

  * 'u-boot-fw-utils' on 32-bit ARM targets.

Make this runtime dependency machine specific.

Fixes: c2e2ecbdcd91 ("kernel-tests: Add kernel performance ptests")


gratian:
- Build tested on both xilinx-zynq and x64 build targets.
- Unpacked the resulting .ipk and verified that the dependencies are correct.

amstewart (approved):

PR #95, opened 2021-01-20
Move from from os-common

chaitu236:
Moved ni-utils related changes from os-common

@ni/rtos


recipes-ni/ni-utils/ni-utils_1.0.bb
```diff
@@ -5,15 +5,16 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 SECTION = "base"
 
 SRC_URI = "file://status_led \
-	   file://nisetbootmode.functions \
-	   file://nisetbootmode \
+           file://nisetbootmode.functions \
+           file://nisetbootmode \
+           file://nisetled \
```
shruthi-ravi: Can you explain why you are moving nisetled to this package?

recipes-ni/ni-utils/ni-utils_1.0.bb
```diff
@@ -11,7 +11,10 @@ SRC_URI = "file://status_led \
 "
 
 FILES_${PN} += "\
-	/usr/lib/nisetbootmode.functions \
+	${bindir}/status_led \
```
shruthi-ravi, line 15: Shouldn't these LED changes be in the previous commit?

recipes-ni/ni-utils/ni-utils_1.0.bb
```diff
@@ -11,7 +11,10 @@ SRC_URI = "file://status_led \
...
+	${libdir}/nisetbootmode.functions \
+	${sysconfdir}/init.d/nisetbootmode \
```
shruthi-ravi, line 17: Again, why move nisetbootmode here? Cleanup?

recipes-ni/ni-utils/ni-utils_1.0.bb
chaitu236: nisetled is the init script for status_led which is provided by this package - so moved it here.

recipes-ni/ni-utils/ni-utils_1.0.bb
chaitu236: These changes aren't specifically for LED, I've replaced paths for all files in the package and it seemed better to do them together rather than breaking them up.

recipes-ni/ni-utils/ni-utils_1.0.bb
chaitu236: nisetbootmode is part of this package already (see below in do_install). I just added it to FILES section.
I'm not entirely sure why they're weren't already added to FILES section in first place, and why bitbake complains if nothing is added to FILES but doesn't when not all of them are.

recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
```diff
@@ -14,8 +14,7 @@ RDEPENDS_${PN}-ptest += "bash"
 DEPENDS_append_x64 = " nilrtdiskcrypt "
 RDEPENDS_${PN}_append_x64 = " nilrtdiskcrypt "
 
-SRC_URI = "file://nisetbootmode \
-           file://firewall \
+SRC_URI = "file://firewall \
```
amstewart: As long as you're touching this line, could you just refactor it like this, so that it stops showing up in the diff?

```
SRC_URI = " \
    file://firewall \
```

recipes-ni/ni-utils/ni-utils_1.0.bb
shruthi-ravi: I liked the initscripts only package separation from before. The status_led is called by other scripts too directly, so it makes sense that is installed by ni-utils. Can you make initscripts depend on ni-utils instead?

amstewart (approved):
RE: `nisetbootmode` & `ni-utils`

I think we might need to refresh `nisetbootmode` to support RAUC. It seems like that script is intended for user use (LV API?) to trigger mode shifts. Certainly, the safemode/runmode code that is in there now will be meaningless in RAUC, and we'll need to add support for changing runlevels.

Strictly migrating it for now is fine, but we should create a work item to refresh the API and move it into its own package. As long as we're pulling it out of `ni-utils`, I don't really like that package's name - it's another "ni-misc", and we already have too many of those. So I think that work item should also ask to move the LED-management scripts into their own recipe/package. 

What do you think @jpautler ?

recipes-ni/ni-utils/ni-utils_1.0.bb
chaitu236: I get your point, but in my view there is a fundamental difference between initscript and other scripts calling status_led. If this initscript were to do other things in addition to calling status_led, it'd make sense to put it in initscripts, but the sole purpose of this initscript is to make sure status_led runs at boot time, so I feel like it should go in same package. We always ship initscripts associated with other software in the same package even when they're called outside the initscript, wouldn't we be breaking this convention if we split them in this case?

shruthi-ravi: > We always ship initscripts associated with other software in the same package even when they're called outside the initscript..

I didn't realize this was the case. I'm good with this then.

recipes-ni/ni-utils/ni-utils_1.0.bb
chaitu236: Moved them to previous commit as per offline conversation.

recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
chaitu236: Done

chaitu236:
V2:
Addressed feedback.

Shruthi brought up the fact that nisetbootmode in meta-nilrt previously contained two files, one in bindir (that took auto-restore, restore, etc as params) and one in /etc/init.d (that has start and stop as params), so a user could invoke nisetbootmode with one of auto-restore, restore, etc without its full path as bindir is in PATH.
But I've now consolidated them into one file in /etc/init.d (like in os-common) so a user cannot invoke it without specifying the full path - but this isn't a concern because there doesn't seem to be any users that invoke the file directly (as per codesearch).

shruthi-ravi (approved):
Looks good, thanks for making the changes.

chaitu236:
> RE: `nisetbootmode` & `ni-utils`
> 
> I think we might need to refresh `nisetbootmode` to support RAUC. It seems like that script is intended for user use (LV API?) to trigger mode shifts. Certainly, the safemode/runmode code that is in there now will be meaningless in RAUC, and we'll need to add support for changing runlevels.
> 
> Strictly migrating it for now is fine, but we should create a work item to refresh the API and move it into its own package. As long as we're pulling it out of `ni-utils`, I don't really like that package's name - it's another "ni-misc", and we already have too many of those. So I think that work item should also ask to move the LED-management scripts into their own recipe/package.
> 
> What do you think @jpautler ?

Agreed we need to update many scripts to remove safemode/runmode assumptions and replace with RAUC. Does [this](https://ni.visualstudio.com/DevCentral/_workitems/edit/1200346/?workitem=1198511) seem like a good feature to put the stories under?

amstewart:
That's the RAUC epic, so yes? I'm not sure what Feature it should go under though.

PR #96, opened 2021-01-21
hostname: Update with os-common content.

shruthi-ravi:
The Host_Name token in ni-rt.ini is still commonly used in other
scripts. So, continue to preserve the hostname saved in ni-rt.ini.

Verified that when hostname is changed from MAX, the hostname
is also written to /etc/hostname.

### DIFF
--- a/recipes-core/initscripts/initscripts-1.0/hostname.sh
+++ b/recipes-core/initscripts/initscripts-1.0/hostname.sh
@@ -5,13 +5,10 @@
 # Version:      @(#)hostname.sh  1.10  26-Feb-2001  miquels@cistron.nl
 #
 
-if [ -f /etc/hostname ] && [ -r /etc/hostname ];then
-    HOSTNAME=$(cat /etc/hostname)
-else
-    HOSTNAME=
-fi
-
+# Check if there is a hostname in ni-rt.ini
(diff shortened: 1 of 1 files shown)


recipes-core/initscripts/initscripts-1.0/hostname.sh
```diff
@@ -5,13 +5,10 @@
...
-    HOSTNAME=$(cat /etc/hostname)
-else
-    HOSTNAME=
-fi
-
+# Check if there is a hostname in ni-rt.ini
+HOSTNAME=$(/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=Host_Name)
 if [ $? -eq 0 ] && [ -n "$HOSTNAME" ]; then
+    # Set the hostname based on ni-rt.ini
     hostname "$HOSTNAME"
```
amstewart, line 12: When we set the hostname here, does it automatically update the value of `/etc/hostname`? We need to ensure the value in that file is kept in-sync with `ni-rt.ini`.

shruthi-ravi: Checked on a target, the /etc/hostname file does not exist, even after updating the hostname. I can update this script to create the file.

amstewart (approved):

PR #97, opened 2021-01-21
wpa-supplicant: Migrate dbus conf and service files from os-common

gratian:
Add NILRT specific wpa-supplicant dbus configuration and service
files which used to live in os-common.

### DIFF
--- a/recipes-connectivity/wpa-supplicant/files/dbus-wpa_supplicant.conf
+++ b/recipes-connectivity/wpa-supplicant/files/dbus-wpa_supplicant.conf
@@ -1,0 +1,51 @@
+<!DOCTYPE busconfig PUBLIC
+ "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
+ "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
+<busconfig>
+        <policy user="webserv">
+                <allow own="fi.epitest.hostap.WPASupplicant"/>
+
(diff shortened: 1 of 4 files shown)


gratian:
Testing:
- built new wpa-supplicant package
- unpacked ipk and verified that the migrated files end up in the correct place

@ni/rtos

amstewart:
Do you know when the `wpa-supplicant` gets installed? ie. Are NI clients supposing that these files are a part of the base image, or do software stacks which enable wireless communication already know to depend on the `wpa-supplicant` IPK?

gratian:
No idea; @jeminor might know. I suspect wpa-supplicant needs to be part of the bundle since it was included in base before. You can't establish a WiFi connection w/o it. Also, these files are useless without wpa-supplicant so they need to go together.

amstewart:
OK. Well, I see that it is already included in the [wifi packagegroup](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-core/packagegroups/packagegroup-ni-wifi.bb#L19) and that is already included in the [`nilrt-base.inc`](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-core/images/nilrt-base.inc#L30). So it should already end up in the RAUC rootfs. That's really what I was driving toward.

amstewart (approved):
ACK

shruthi-ravi (approved):

PR #98, opened 2021-01-22
Move changes in nisetlocale.sh, sysctl.conf, start_getty from os-common

chaitu236:
@ni/rtos

### DIFF
--- a/recipes-core/sysvinit/sysvinit-inittab/start_getty
+++ b/recipes-core/sysvinit/sysvinit-inittab/start_getty
@@ -1,28 +1,24 @@
 #!/bin/sh
+# Copyright (c) 2012-2013 National Instruments.
+# All rights reserved.
+# Spawn getty on the enabled, controlling console device. There will be at most
+# one of these. If there aren't any, use /dev/tty0. (This mostly matches
+# sysvinit behavior when $CONSOLE is unset.)
 
-# $1 and $2 are are passed by inittab, its values are given by OE
-CONSOLE="$2"
-SPEED="$1"
+. /etc/natinst/networking/functions.common
+update_all_interface_info
 
-if [[ -z "$CONSOLE" || ! -e "/dev/$CONSOLE" || -z "$SPEED" ]]; then
-    echo Specified console "$CONSOLE" with baud rate "$SPEED" is invalid, autodetecting...
+[ "$1" != "" ] && DEFAULT_CONSOLE="$1" || DEFAULT_CONSOLE="tty2"
+CONSOLE=$DEFAULT_CONSOLE
+while read cons flags; do
+   case "$flags" in
+   *EC*) CONSOLE="$cons"
+         break
+         ;;
+   esac
+done < /proc/consoles
 
-    CONSOLE=""
-    while read cons flags; do
-	case "$flags" in
-	    *EC*) CONSOLE="$cons"
-		  SPEED=$(stty speed < /dev/$CONSOLE)
-		  echo Autodetected console "$CONSOLE" with baud rate "$SPEED"
-		  break
-		  ;;
-	esac
-    done < /proc/consoles
-
-    if [[ -z "$CONSOLE" ]]; then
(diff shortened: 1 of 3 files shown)


recipes-core/sysvinit/sysvinit-inittab/start_getty
```diff
@@ -1,28 +1,24 @@
...
+# All rights reserved.
+# Spawn getty on the enabled, controlling console device. There will be at most
+# one of these. If there aren't any, use /dev/tty0. (This mostly matches
+# sysvinit behavior when $CONSOLE is unset.)
 
-# $1 and $2 are are passed by inittab, its values are given by OE
-CONSOLE="$2"
-SPEED="$1"
+. /etc/natinst/networking/functions.common
+update_all_interface_info
```
amstewart, line 10: I recall us configuring the CONSOLE in the LVComms images (which is where `start_getty` was used) very intentionally. Can you comment on why the changes we made in this file are no longer necessary, and why `consolegetty` should be preferred in all cases?

recipes-ni/base-files-nilrt/files/sysctl.conf
```diff
@@ -1,12 +1,17 @@
...
+# Restrict virtual memory allocations to a percentage of the RAM that the
+# kernel makes available for virtual memory (the "CommitLimit" figure in
+# /proc/meminfo).  This setting was determined empirically.
+#
+# The maximum "safe" value of vm.overcommit_ratio (i.e., to avoid the OOM
+# killer) depends on the working set of the system, particularly for those
+# processes which are not mlockall()'d. Values above 98 are known to be unsafe.
+#
+# Note: The overcommit ratio is now set by the nisetcommitratio script using
+# memory reservation amounts defined in ni-rt.ini
```
amstewart: NIT: Could you strike "now" from this comment. The reader probably doesn't care that this value *used to* be set here.

amstewart (changes requested):
I'll defer to your expertise on `nisetlocale.sh`.

Looks good apart from the couple comments below.

recipes-core/sysvinit/sysvinit-inittab/start_getty
chaitu236: I wanted to follow what we have in os-common, so I basically replaced start_getty contents with consolegetty. From what I understand, consolegetty does everything start_getty does but also more stuff (printing interface info, replacing CONOLE with DEFAULT_CONSOLE in case it is set to tty0, tty1), so it seemed better to go with consolegetty contents.
Does anyone know why we chose to keep CG, NXG contents different in this case? If we deliberately chose start_getty contents, why didn't we also make the same changes to consolegetty?

chaitu236:
> I'll defer to your expertise on `nisetlocale.sh`.
> 
> Looks good apart from the couple comments below.

This is basically same content from os-common. I'm sure its contents will change when the issue is fixed.

recipes-core/sysvinit/sysvinit-inittab/start_getty
amstewart: @rtollert ; since you are probably the only one who understands the black magic we do to keep CONSOLE information sane, can you check out the changes in this PR and comment if anything jumps out at you as obviously wrong?

rtollert: I think @adelcast did all the comms stuff. It was rather deliberately meant to be much closer to upstream IIRC.
The primary value add of consolegetty is that it gives us a way to sub in the serial console if it's enabled. (Updating `/etc/issue` was always a stretch given hotplug.) That never existed for comms (and indeed its value remains a little questionable for PXI).

I don't have a problem with any of these changes.

rtollert (approved):

chaitu236:
V2: Removed "now" from sysctl.conf

amstewart (approved):
Alejandro has no objections, so this is fine with me.

PR #99, opened 2021-01-22
pkggrp-ni-coreimagerepo: add pkggrp-ni-transconf

amstewart:
Add the transconf packagegroup to the coreimagrepo so that we only have one necessary BB target to create the `core/` package feed. Previously, pkggrp-ni-transconf was built as an automatic dependency of one of the image builds done in the NIOE component.

-----

The transconf packagegroup is needed by all RAUC-enabled images to
provide package-specific hooks for core components. Add it to the
coreimagerepo to ensure that it is always available to RAUC image
builds.

* Trivial sort of the packagegroups in the RDEPENDS variable.


amstewart:
@ni/rtos

billpittman (approved):

PR #105, opened 2021-01-26
nisdbootconfig: remove from base package and only install on safemode

NIwkoe:
Commit ec1173c adds nisdbootconfig into base package with the condition to
only install on ARM images for both safemode and runmode. However, software
installation did not succeed because nisdbootconfig as a bootscript cannot
set up fw_env.config early enough for installation scripts that uses
fw_printenv. Hence, the nisdbootconfig script for runmode will be moved into
postinst of systemlink base image.

This commit includes removing the safemode check portion of the script,
removing nisdbootconfig from base recipe and adding it into safemode recipe
with append_armv7a.

@ni/rtos 

### DIFF
--- a/recipes-ni/nisdbootconfig/files/nisdbootconfig
+++ b/recipes-ni/nisdbootconfig/files/nisdbootconfig
@@ -2,19 +2,15 @@
 set -e
 
 # check if target is NOT using ubifs (mtd)
-if ! cat /proc/mtd | grep -qs mtd ; then
+if ! grep -qs mtd /proc/mtd ; then
 
     # configure the partitions to mount by label in fstab
     sed -i '/ubifs/d' /etc/fstab
     echo LABEL=nibootfs         /boot                ext4       sync           0  0 >>/etc/fstab
     echo LABEL=niconfig         /etc/natinst/share   ext4       sync           0  0 >>/etc/fstab
+    echo LABEL=nirootfs         /mnt/userfs          ext4       defaults       0  0 >>/etc/fstab
     mkdir -p /etc/natinst/share
-
-    # configure the nirootfs partition if we are in safemode
-    if [ -f /etc/natinst/safemode ]; then
-        echo LABEL=nirootfs         /mnt/userfs          ext4       defaults       0  0 >>/etc/fstab
-        mkdir -p /mnt/userfs
-    fi
+    mkdir -p /mnt/userfs
 
     # configure the path for fw_printenv to read environmental variables from u-boot
     # replace /dev/mtd4 and /dev/mtd5 with /boot/uboot/uboot.env
--- a/recipes-core/packagegroups/packagegroup-ni-base.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-base.bb
@@ -39,7 +39,6 @@ NILRT_NXG_x64_PACKAGES = "\
 NILRT_ARM_PACKAGES = "\
 	jitterentropy-rngd \
 	mtd-utils-ubifs \
-	nisdbootconfig \
 "
 
 NILRT_x64_PACKAGES = "\
--- a/recipes-core/packagegroups/packagegroup-ni-safemode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-safemode.bb
@@ -16,3 +16,7 @@ RDEPENDS_${PN} = " \
 	e2fsprogs-mke2fs \
 	e2fsprogs-tune2fs \
 "
+
+RDEPENDS_${PN}_append_armv7a = " \
+	nisdbootconfig \
+"
--- a/recipes-ni/nisdbootconfig/nisdbootconfig.bb
+++ b/recipes-ni/nisdbootconfig/nisdbootconfig.bb
@@ -1,5 +1,5 @@
 SUMMARY = "NI SDboot Config"
-DESCRIPTION = "Configuration script for arm target booting from SD Card"
+DESCRIPTION = "Configuration script for safemode to enable arm target to boot from SD Card"
 LICENSE = "MIT"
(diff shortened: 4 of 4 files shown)


NIwkoe:
I've verified that safemode image built using zynq from OE can be booted normally and software installation on the systemlink image built using zynq_dkms can be done successfully. I also checked that no nisdbootconfig.postinst and any other related staging files (including the script itself) exist in x64 and x64_dkms images, hopefully I won't break x64 builds again this round.

amstewart:
So, is the new intention that the safemode version of the script will remain as an initscript, but that the runmode version will be invoked only once, as a postinst to the SystemImage installation?

NIwkoe:
> So, is the new intention that the safemode version of the script will remain as an initscript, but that the runmode version will be invoked only once, as a postinst to the SystemImage installation?

Yup, that is the intention. Since safemode should pre-exist on the target, I think using the initscript might be the only way.

shruthi-ravi (approved):
Can you update the nisdbootconfig recipe description to say that it is intended for safemode only? Otherwise, looks good.

NIwkoe:
> Can you update the nisdbootconfig recipe description to say that it is intended for safemode only? Otherwise, looks good.

Ok, updated the description in the recipe.

recipes-ni/nisdbootconfig/files/nisdbootconfig
```diff
@@ -8,13 +8,9 @@ if ! cat /proc/mtd | grep -qs mtd ; then
...
     echo LABEL=niconfig         /etc/natinst/share   ext4       sync           0  0 >>/etc/fstab
+    echo LABEL=nirootfs         /mnt/userfs          ext4       defaults       0  0 >>/etc/fstab
     mkdir -p /etc/natinst/share
-
-    # configure the nirootfs partition if we are in safemode
-    if [ -f /etc/natinst/safemode ]; then
-        echo LABEL=nirootfs         /mnt/userfs          ext4       defaults       0  0 >>/etc/fstab
-        mkdir -p /mnt/userfs
-    fi
+    mkdir -p /mnt/userfs
```
amstewart, line 13: Why is it appropriate to remove this safemode check? Does the SystemImage postinst script get executed from the safemode context, rather than from a chroot into runmode?

NIwkoe: On previous implementation, we run nisdbootconfig for both safemode and runmode. Since we do not want to mount /mnt/userfs in runmode, I added the safemode check to isolate this portion out when in runmode. Now that we only run nisdbootconfig in safemode, I thought the safemode check is redundant here.

I believe there's a mix of both in the postinst script. There are some lines where dir manipulation (mainly in /mnt/userfs) is done which I think is in the context of safemode, while others like:

`# Import NI signing key for subsequent opkg operations`
`/usr/sbin/chroot /mnt/userfs/ opkg-key populate`

make use of chroot in the SystemImage postinst script.

amstewart: Yeah, that distinction is what I'm asking about. If you run this script while chrooted into /mnt/userfs, then I *think* the `/etc/fstab` you're modifying will be the *runmode* fstab (assuming that we don't symlink that file between safemode and runmode). So I would guess that the script, as written, might write the `/mnt/userfs` fstab entry into the runmode fstab file - which isn't a sane entry for the runmode context.

I might be mistaken about how you're calling the script though. You can quickly check that my concerns are invalid by checking the value of `/mnt/userfs/etc/fstab` after the script is run. As long as it doesn't get the `/mnt/userfs` entry, I'm satisfied.

NIwkoe: I've checked and confirm that the scripts only writes to /mnt/userfs/etc/fstab (and fw_env.config) and that after the script was executed, no /mnt/userfs/mnt/userfs exists.

recipes-ni/nisdbootconfig/files/nisdbootconfig
```diff
@@ -2,19 +2,15 @@
 set -e
 
 # check if target is NOT using ubifs (mtd)
-if ! cat /proc/mtd | grep -qs mtd ; then
+if ! grep -qs mtd /proc/mtd ; then
```
NIwkoe, line 5: I've also amended this line as per @gratian 's comment over at the review-board that discussed about the redundancy of using "cat" in this case. Refer to https://review-board.natinst.com/r/356932/.

NIwkoe:
I've also amended the condition check as per @gratian 's comment over at the review-board that discussed about the redundancy of using "cat" in this case. Refer to https://review-board.natinst.com/r/356932/.

recipes-ni/nisdbootconfig/files/nisdbootconfig
amstewart: OK; but does `/mnt/userfs/etc/fstab` contain an entry for `/mnt/userfs`?

NIwkoe: I realized I got confused by myself with the other commit in review-board. This script runs as S00nisdbootconfig startup script instead of by postinst. It will not be executed by chroot. Only the /etc/fstab in safemode gets modified from this script.

Nope, /mnt/userfs/etc/fstab does not contain entry for /mnt/userfs.

NIwkoe:
Btw, I'm trying out a new condition check which makes use of device tree model compatibility instead of detecting mtd partition at the systemlink side of things (more info at https://review-board.natinst.com/r/356932/). If that gets a ship it, I shall port it into nisdbootconfig.

amstewart (approved):

PR #106, opened 2021-01-26
pkggrp-ni-coreimagrepo: add oe-core packagegroups

amstewart:
Packagesgroups-(base, core-boot) are normally included in all images
which inherit the OE base image class. Include them in the coreimagerepo
to ensure that they are available in the feeds at image-creation-time.

@ni/rtos


amstewart:
PATCH V2:
* Fixed grammer in commit message to match with this PR.

shruthi-ravi (approved):

PR #107, opened 2021-01-26
nilrt.inc: Add the lv2020 feed to the feed URIs list

shruthi-ravi:
The lv2020 feed hosts the ni-rtlog package that is required to be
installed in the base image. The previous solution moved the ni-rtlog
to the ni-main feed, but that broke LabVIEW installations for a
number of downstream clients who were replacing the ni-main feed with a
local feed.
To workaround the breakage, install ni-rtlog from the lv2020 feed
instead. The previous solution can be adopted again (reverting this
change) when downstream clients are able to test with the new base
image that includes the ni-rtlog package.

Related bug [1244148](https://ni.visualstudio.com/DevCentral/_workitems/edit/1244148).
Tested that a package from the lv2020 feed could be installed into the base
image.

@ni/rtos


amstewart:
As a temporary fix, I'm OK with this.

We need an ACK from someone on HW config though. The source name field for the NIFeeds is important to the MAX feed configuration workflow (iirc.). I don't know if changing them from `NI-software` to something else is going to break them.

shruthi-ravi:
Are you talking about this: PACKAGE_FEED_URIS_ADDITIONAL[**lv2020**]? If that's the case, I can change lv2020 and ni-main to NI-software.
The SystemImage component also adds the NI-software.conf file with the lv feed paths, right? Isn't that enough for now?

conf/distro/nilrt.inc
```diff
@@ -147,5 +151,6 @@ IPK_FEED_URIS += "\
     NIOE-all##${NILRT_LOCAL_FEED_URI}/all \
     NIOE-${MACHINE}##${NILRT_LOCAL_FEED_URI}/${MACHINE} \
     NIOE-${TUNE_PKGARCH}##${NILRT_LOCAL_FEED_URI}/${TUNE_PKGARCH} \
-    NI-software##${NILRT_MACHINE_FEED_URI_nimain} \
+    NI-main-software##${NILRT_MACHINE_FEED_URI_nimain} \
+    NI-lv-software##${NILRT_MACHINE_FEED_URI_lv2020} \
```
amstewart, line 149: I was talking about these lines actually. I can't remember now whether or not these sources will only be used during do_rootfs, or if OE will actually write them into the permanent `/etc/opkg/base-feeds.conf` (or whatever the file is called.)

If it does write them, then I think they will interfere with the source files which the SystemImage installer adds after us. And also, I'm not sure how the MAX workflows are going to react, given that we [intentionally chose](https://review-board.natinst.com/r/327590/diff/1#index_header) the names of those source feeds, with approval from HW config.

amstewart (approved):

shruthi-ravi:
I verified that even when the PACKAGE_FEED_URIS_ADDITIONAL entries for ni-main and lv2020 are removed, bitbake is still able to download the package from the feeds correctly and doesn't create the lv2020.conf and ni-main.conf files in the image.

amstewart (approved):

PR #112, opened 2021-01-28
Move files from os-common. Part 5

chaitu236:
Move ifplugd.conf, nisetcommitratio

@ni/rtos

### DIFF
--- a/recipes-core/busybox/busybox_1.31.%.bbappend
+++ b/recipes-core/busybox/busybox_1.31.%.bbappend
@@ -3,6 +3,7 @@ FILESEXTRAPATHS_prepend := "${THISDIR}:${THISDIR}/files:${THISDIR}/${PN}:"
 SRC_URI =+ " \
             file://busybox-ifplugd \
             file://ifplugd.action \
+            file://ifplugd.conf \
             file://busybox-acpid \
             file://acpid.conf \
             file://acpid_poweroff.sh \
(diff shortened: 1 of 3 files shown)


amstewart (approved):

shruthi-ravi:
The files here also seem to be part of the ever-expanding list of files that were updated to not use ni-rt.ini in NXG (I presume). [Changes](https://github.com/ni/meta-nilrt/commit/2473a6c69bbe0f78efbd2555be6de7dd85e6b8d5#diff-b281fabbe8cd8d698aa1dd70551ef6e4fc4f56b1fb2cdd8fe0f5593f7ee69e55) that we are reverting now, going back to the old behavior. 
Just an observation :).

Btw, there seem to be a couple of tests for nisetcommitratio based on the changes made in the above PR. What should we do about them? Delete?

chaitu236:
> The files here also seem to be part of the ever-expanding list of files that were updated to not use ni-rt.ini in NXG (I presume). [Changes](https://github.com/ni/meta-nilrt/commit/2473a6c69bbe0f78efbd2555be6de7dd85e6b8d5#diff-b281fabbe8cd8d698aa1dd70551ef6e4fc4f56b1fb2cdd8fe0f5593f7ee69e55) that we are reverting now, going back to the old behavior.
> Just an observation :).
> 
> Btw, there seem to be a couple of tests for nisetcommitratio based on the changes made in the above PR. What should we do about them? Delete?

One of them passes but there are some failures in the other one. How about we create a US to fix the failing ones?

PR #114, opened 2021-02-01
initscripts-nilrt: Add populateconfig from os-common

shruthi-ravi:
populateconfig references in [os-common](https://codesearch.natinst.com/search/text?q=populateconfig%20file%3A%2Fos-common&fold_case=auto&regex=false&context=true).
Built the initscripts-nilrt IPK, unpacked and verified that the symlinks looked as expected.

@ni/rtos

### DIFF
--- a/recipes-ni/initscripts-nilrt/files/populateconfig
+++ b/recipes-ni/initscripts-nilrt/files/populateconfig
@@ -1,0 +1,100 @@
+#!/bin/sh
+# Copyright (c) 2012-2013 National Instruments.
+# All rights reserved.
+
+if [ "$1" = "start" ]
+then
+	#check if the system is in reset state
+	reset_state=`/sbin/fw_printenv -n sys_reset 2>&1`
+	if [ "$reset_state" = "true" -o "$sys_reset" = "true" ]; then
+		#restore default environmnet vars
+		cp -fp /boot/grub/grubenv.bak /boot/grub/grubenv
+		chown root:ni /boot/grub/grubenv
+		chmod 664 /boot/grub/grubenv
+		/usr/local/natinst/bin/niresetip
+	fi
+
+	if [ -x /usr/local/natinst/bin/nirtcfg ]; then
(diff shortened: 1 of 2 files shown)


recipes-ni/initscripts-nilrt/files/populateconfig
```diff
@@ -0,0 +1,100 @@
...
+	reset_state=`/sbin/fw_printenv -n sys_reset 2>&1`
+	if [ "$reset_state" = "true" -o "$sys_reset" = "true" ]; then
+		#restore default environmnet vars
+		cp -fp /boot/grub/grubenv.bak /boot/grub/grubenv
+		chown root:ni /boot/grub/grubenv
+		chmod 664 /boot/grub/grubenv
+		/usr/local/natinst/bin/niresetip
+	fi
+
+	if [ -x /usr/local/natinst/bin/nirtcfg ]; then
```
shruthi-ravi, line 17: This file is run in run levels 0 and 6 too apparently, and only this part of the script makes sense to run at that time. So I left those runlevels as is.

amstewart (approved):
(NONBLOCKING)

I suppose we probably intended to replace this file (at some point) with [nipopulateconfigdir](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-ni/initscripts-nilrt/files/nipopulateconfigdir) instead?

Does that mean that this is a legacy script that we should start deprecating on its own schedule? Or do they interact in some other way?

shruthi-ravi:
The logic in this file looks relevant. I think that the intent when nipopulateconfigdir was created was to split up populateconfig into several smaller shell scripts located in /etc/init.d/populateconfig.d/ or have clients install scripts to that directory. codesearch reveals only one client that actually installs a shell script to that directory to be run by nipopulateconfigdir. Given the contents of the populateconfig script, I don't know if we can push ownership of the contents to clients outside RTOS. And a few of our scripts also invoke /etc/init.d/populateconfig directly.
All that to say, I think we should leave populateconfig as is for now, and if we ever get time, split it up, and update the caller scripts.

PR #115, opened 2021-02-01
Move os-common to OE. Part 6

chaitu236:
Moved safemode utils into a new recipe ni-safemode-utils
Moved networking scripts from rtcfg into different recipes in meta-nilrt

@ni/rtos

### DIFF
--- a/recipes-core/busybox/files/simple.script
+++ b/recipes-core/busybox/files/simple.script
@@ -1,0 +1,95 @@
+#!/bin/sh
+
+# udhcpc script edited by Tim Riker <Tim@Rikers.org>
+# further modified by National Instruments
+
+[ -z "$1" ] && echo "Error: should be called from udhcpc" && exit 1
+
+. /etc/natinst/networking/functions.common
+
+case "$1" in
+   deconfig)
+      # Get rid of any global-scope IPv4 addresses on this interface.
+      remove_addresses $interface global;
+
+      # If interface isn't up at all, bring it up.
+      if ! ( is_interface_up $interface ) ; then
+         /sbin/ifconfig $interface up 0.0.0.0
+      fi
+      update_interface_info $interface
+      ;;
+
+   leasefail)
+      if ( is_linklocal_enabled $interface ); then
+         start_linklocal_client $interface;
+      fi
+      update_interface_info $interface
+      ;;
+
+   renew|bound)
+      RESOLV_CONF="/etc/resolv.conf.${interface}"
+      BROADCAST="broadcast +"
+      [ -n "$broadcast" ] && BROADCAST="broadcast $broadcast"
+      if [ -n "$subnet" ]; then
+         prefix_length_from_netmask $subnet
+         PREFIX=$?
+      fi
+
+      if ( is_primary_interface $interface ); then
+         METRIC=0
+      else
+         METRIC=5
+      fi
+
+      # Save the IP address that we get so we can request it next time.
+      set_var $(uboot_var_for_interface dhcpipaddr $interface) $ip
+
+      # if there's a link-local IP get rid of it.
+      if ( is_linklocal_enabled $interface ); then
+         stop_linklocal_client $interface;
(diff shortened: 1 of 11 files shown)


recipes-core/busybox/files/simple.script
```diff
@@ -0,0 +1,95 @@
+#!/bin/sh
+
+# udhcpc script edited by Tim Riker <Tim@Rikers.org>
+# further modified by National Instruments
```
amstewart, line 4: Where is this script installed and used?

chaitu236: Its installed as /etc/udhcpc.d/50default.
If the `-s` argument isn't given, udhcpc defaults to using [/usr/share/udhcpc/default.script](https://github.com/ni/openembedded-core/blob/nilrt/master/dunfell/meta/recipes-core/busybox/files/default.script#L1) which ends up executing /etc/udhcpc.d/50default

amstewart: Oh; I see. We're overwriting the upstream script implementation directly. OK.

amstewart (approved):

recipes-ni/ni-safemode-utils/ni-safemode-utils_1.0.bb
```diff
@@ -0,0 +1,36 @@
+SUMMARY = "nilrt safemode utilities"
+DESCRIPTION = "nilrt distro-specific safemode utilities that provide basic system functionality."
```
shruthi-ravi, line 2: I have my reservations to all these references to "safemode". Maybe they will go away/be updated when we figure out the role of these files in a RAUC system?

recipes-ni/ni-safemode-utils/ni-safemode-utils_1.0.bb
```diff
@@ -0,0 +1,36 @@
...
+	${natinstbin}/nisafemodeversion \
+	${natinstbin}/niinstallsafemode \
+	${natinstbin}/nicompareversion \
+"
+
+DEPENDS += "shadow-native pseudo-native niacctbase update-rc.d-native"
+
+RDEPENDS_${PN} += "niacctbase bash"
+
+RDEPENDS_${PN}_append_x64 += "fw-printenv"
```
shruthi-ravi, line 25: Is it reasonable to not have the _append_x64 since these recipes are now exclusively for x64 systems?

recipes-ni/ni-safemode-utils/ni-safemode-utils_1.0.bb
```diff
@@ -0,0 +1,36 @@
...
+
+S = "${WORKDIR}"
```
shruthi-ravi, line 27: What is this used for?

recipes-ni/ni-safemode-utils/ni-safemode-utils_1.0.bb
```diff
@@ -0,0 +1,36 @@
...
+
+do_install () {
+	install -d ${D}/usr/local/natinst/bin
```
shruthi-ravi, line 30: Can use natinstbin here too.

recipes-ni/ni-safemode-utils/ni-safemode-utils_1.0.bb
chaitu236: My hope is that the recipe itself will go away completely.

recipes-ni/ni-safemode-utils/ni-safemode-utils_1.0.bb
chaitu236: Sure, I'll remove it

recipes-ni/ni-safemode-utils/ni-safemode-utils_1.0.bb
chaitu236: Based this recipe off of something else that had this. Seems like its not required. Will remove

PR #116, opened 2021-02-01
fw_printenv: Add fw_env.config from os-common

jpautler:
Move fw_env.config from os-common.

Built the recipe and confirmed the ipk contains the new contents.

@ni/rtos

### DIFF
--- a/recipes-ni/ni-utils/files/x64/fw_env.config
+++ b/recipes-ni/ni-utils/files/x64/fw_env.config
@@ -1,0 +1,7 @@
+# Configuration file for fw_(printenv/saveenv) utility.
+# Up to two entries are valid, in this case the redundand
+# environment sector is assumed present.
+
+# MTD device name	Device offset	Env. size	Flash sector size
+/dev/mtd4		0x0000		0x20000		0x20000
+/dev/mtd5		0x0000		0x20000		0x20000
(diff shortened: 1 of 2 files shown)


recipes-ni/ni-utils/files/x64/fw_env.config
```diff
@@ -0,0 +1,7 @@
+# Configuration file for fw_(printenv/saveenv) utility.
+# Up to two entries are valid, in this case the redundand
+# environment sector is assumed present.
+
+# MTD device name	Device offset	Env. size	Flash sector size
+/dev/mtd4		0x0000		0x20000		0x20000
+/dev/mtd5		0x0000		0x20000		0x20000
```
amstewart, line 7: Do we use these flash storage sectors across all product lines or just on ARM? I though it was the latter.

shruthi-ravi: Yes, these entries are for ARM. I don't think the fw_env.config is used in an x64 context. Assuming this is the reimplemented [fw_printenv](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-ni/ni-utils/files/x64/fw_printenv#L1) for x64 targets.

jpautler: I don't know. I was basing this change on the fact that the file is already included in x64 builds today, but perhaps it's currently unused on x64. Who would know? @jeminor?

amstewart: :thinking:

jpautler: Looks like we don't need this in x64. I'm going to close this pull request.

PR #117, opened 2021-02-02
ni-utils: Add nisetprimarymac from os-common

jpautler:
Move nisetprimarymac from os-common.

Built the recipe and confirmed the ipk contains the new contents.

@ni/rtos

### DIFF
--- a/recipes-ni/ni-utils/ni-utils_1.0.bb
+++ b/recipes-ni/ni-utils/ni-utils_1.0.bb
@@ -9,6 +9,7 @@ SRC_URI = "\
 	file://nisetbootmode.functions \
 	file://nisetbootmode \
 	file://nisetled \
+	file://nisetprimarymac \
 	file://functions.common \
 "
 
@@ -17,6 +18,7 @@ FILES_${PN} += "\
 	${libdir}/nisetbootmode.functions \
 	${sysconfdir}/init.d/nisetbootmode \
 	${sysconfdir}/init.d/nisetled \
+	${sysconfdir}/init.d/nisetprimarymac \
 	${sysconfdir}/natinst/networking/functions.common \
 "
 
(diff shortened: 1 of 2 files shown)


recipes-ni/ni-utils/ni-utils_1.0.bb
```diff
@@ -38,11 +40,14 @@ do_install () {
...
+	install -m 0550   ${WORKDIR}/nisetprimarymac             ${D}${sysconfdir}/init.d
 	install -m 0755   ${WORKDIR}/functions.common            ${D}${sysconfdir}/natinst/networking
 
 	update-rc.d -r ${D} nisetled              start 40 S .
 	update-rc.d -r ${D} nisetbootmode         start 80 S . stop 0 0 6 .
+	update-rc.d -r ${D} nisetprimarymac       start 4 5 .
 
 	chown 0:${LVRT_GROUP} ${D}${bindir}/status_led
 	chown 0:${LVRT_GROUP} ${D}${sysconfdir}/init.d/nisetbootmode
+	chown 0:${LVRT_GROUP} ${D}${sysconfdir}/init.d/nisetprimarymac
```
shruthi-ravi: Is this needed for this script? I don't see an equivalent entry in the makefiles. These scripts all make sense to be owned by the ni group, but just want to make sure it's an intentional change.

amstewart (approved):

recipes-ni/ni-utils/ni-utils_1.0.bb
jpautler: Good eye! Looks like I got overzealous in adding an nisetprimarymac reference everywhere I saw an nisetbootmode reference. On an existing system, it looks like everything in /etc/init.d is owned by the administrators group with the exception of nisetbootmode and nisetconsoleout which are owned by the ni group.

I'll update to remove this line.

jpautler: Done.

shruthi-ravi (approved):

PR #118, opened 2021-02-02
Move os-common to OE. Part 7

chaitu236:
Bunch of small changes to move stuff from os-common.

@ni/rtos

### DIFF
--- a/recipes-kernel/linux/linux-nilrt.inc
+++ b/recipes-kernel/linux/linux-nilrt.inc
@@ -45,6 +45,9 @@ KBUILD_DEFCONFIG_x64 = "nati_x86_64_defconfig"
 KBUILD_DEFCONFIG_armv7a = "nati_zynq_defconfig"
 KCONFIG_MODE="--alldefconfig"
 
+# autoload if installed
+KERNEL_MODULE_AUTOLOAD += "ehci-hcd g_ether xilinx-xadcps"
+
 # i915 video driver module options to disable render and display power saving
 # disabling them improves real-time determinism
 KERNEL_MODULE_PROBECONF += "i915"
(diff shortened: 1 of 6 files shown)


recipes-core/base-files/base-files/x64/fstab
```diff
@@ -6,6 +6,9 @@ devpts               /dev/pts             devpts     mode=0620,gid=5       0  0
 tmpfs                /run                 tmpfs      mode=0755,nodev,nosuid,strictatime 0  0
 tmpfs                /var/volatile        tmpfs      size=26%              0  0
 
+LABEL=nibootfs       /boot                ext4       sync                  0  0
+LABEL=niconfig       /etc/natinst/share   ext4       sync                  0  0
+
```
amstewart: I don't think either of these entries are relevant for RAUC images; right? So maybe we should just not bother migrating them, since all the images we care about going-forward will need it reverted or overwritten.

recipes-kernel/linux/linux-nilrt.inc
```diff
@@ -45,6 +45,9 @@ KBUILD_DEFCONFIG_x64 = "nati_x86_64_defconfig"
 KBUILD_DEFCONFIG_armv7a = "nati_zynq_defconfig"
 KCONFIG_MODE="--alldefconfig"
 
+# autoload if installed
+KERNEL_MODULE_AUTOLOAD += "ehci-hcd g_ether xilinx-xadcps"
+
```
amstewart, line 50: >xilinx-xadcps

Are some of these modules ARM only? Is that going to be a problem if you define them here?

amstewart (changes requested):

recipes-kernel/linux/linux-nilrt.inc
chaitu236: Won't be a problem if the modules aren't built for x64. Only when a module's ipk is built will this setting make a difference.

recipes-core/base-files/base-files/x64/fstab
chaitu236: Sure. Removed

PR #119, opened 2021-02-02
ni-utils: Move wireless udhcpd configuration file from os-common

gratian:

### DIFF
--- a/recipes-ni/ni-utils/files/udhcpd.wlan0.conf
+++ b/recipes-ni/ni-utils/files/udhcpd.wlan0.conf
@@ -1,0 +1,11 @@
+start 172.16.0.10
+end 172.16.0.254
+
+interface wlan0
+
+lease_file /var/volatile/udhcpd.wlan0.leases
+pidfile /var/run/udhcpd.wlan0.pid
+
+option subnet 255.255.0.0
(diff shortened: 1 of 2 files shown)


gratian:
Attn: @ni/rtos

recipes-ni/ni-utils/files/udhcpd.wlan0.conf
```diff
@@ -0,0 +1,11 @@
+start 172.16.0.10
```
shruthi-ravi, line 1: Does it make sense to add this file to the busybox recipe in a udhcpd package instead, like the ifplugd.conf file [here](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-core/busybox/files/ifplugd.conf)?

gratian: Yeah, I'll move it. I really don't like tangling more things into busybox - we've been slowly replacing busybox utilities with the full functionality stand alone ones but I guess for now that makes more sense.

gratian: Closing this PR in favor of a new one against the busybox recipe.

PR #120, opened 2021-02-03
busybox-udhcpd: Move wireless udhcpd configuration file from os-common

gratian:

### DIFF
--- a/recipes-core/busybox/busybox_1.31.%.bbappend
+++ b/recipes-core/busybox/busybox_1.31.%.bbappend
@@ -10,7 +10,8 @@ SRC_URI =+ " \
             file://acpid_poweroff.sh \
             file://acpid-logrotate.conf \
             file://zcip-allow-action-script-to-reject-chosen-IP.patch \
-            file://login-utilities.cfg"
+            file://login-utilities.cfg \
+            file://udhcpd.wlan0.conf"
 
 SRC_URI_append_x64 += "file://enable_ar_create_fragment.cfg"
 
@@ -27,6 +28,7 @@ DEPENDS =+ " libselinux"
 FILES_${PN}-ifplugd = "${sysconfdir}/init.d/busybox-ifplugd ${sysconfdir}/ifplugd/ifplugd.action ${sysconfdir}/ifplugd/ifplugd.conf"
 FILES_${PN}-acpid = "${sysconfdir}/init.d/busybox-acpid ${sysconfdir}/acpid.conf ${sysconfdir}/acpi ${sysconfdir}/acpi/poweroff.sh"
 FILES_${PN}-zcip = "${sysconfdir}/natinst/networking/zcip.script"
+FILES_${PN}-udhcpd =+ "${sysconfdir}/udhcpd.wlan0.conf"
 
 INITSCRIPT_PACKAGES =+ " ${PN}-acpid"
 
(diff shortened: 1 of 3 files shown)


gratian:
Attn: @ni/rtos

recipes-core/busybox/busybox_1.31.%.bbappend
```diff
@@ -27,6 +28,7 @@ DEPENDS =+ " libselinux"
 FILES_${PN}-ifplugd = "${sysconfdir}/init.d/busybox-ifplugd ${sysconfdir}/ifplugd/ifplugd.action ${sysconfdir}/ifplugd/ifplugd.conf"
 FILES_${PN}-acpid = "${sysconfdir}/init.d/busybox-acpid ${sysconfdir}/acpid.conf ${sysconfdir}/acpi ${sysconfdir}/acpi/poweroff.sh"
 FILES_${PN}-zcip = "${sysconfdir}/natinst/networking/zcip.script"
+FILES_${PN}-udhcpd =+ "${sysconfdir}/udhcpd.wlan0.conf"
```
chaitu236, line 31: Should busybox-udhcpd be added to packagegroup-ni-base.bb?

amstewart (approved):

jeminor:
Looks good to me after chaitu236's suggestion.

recipes-core/busybox/busybox_1.31.%.bbappend
gratian: I did not add the busybox-udhcpd package (I'm just adding a file to it) so I was assuming is already in the image. Not sure what pulls it in. I'll see if I can track down the dependency chain.

billpittman (approved):

recipes-core/busybox/busybox_1.31.%.bbappend
gratian: Btw. busybox-udhcpc is in the same boat i.e. AFAICT not listed explicitly in packagegroup-ni-base.bb but used in meta-nilrt.

jeminor (approved):

recipes-core/busybox/busybox_1.31.%.bbappend
chaitu236: There is a busybox-udhcpd in feeds (which only has an initscript) but its not in the image. The udhcpd binary itself is part of busybox package. So we have to explicitly add it to packagegroup for this conf file to get installed. But I don't think we'd want to install the initscript that comes with the package as we run udhcp ourself.

chaitu236: It looks like busybox-udhcpc ends up getting installed in base image as busybox has recommends dependency on it, but thats not the case with busybox-udhcpd

gratian: > So we have to explicitly add it to packagegroup for this conf file to get installed. But I don't think we'd want to install the initscript that comes with the package as we run udhcp ourself.

Well that's unfortunate. I guess I'll have to figure out how to remove the init script in our bbappend.

chaitu236: Adding following seems to atleast prevent creation of postinst (that runs update-rc.d).

```
unset INITSCRIPT_NAME_${PN}-udhcpd
INITSCRIPT_PACKAGES_remove = "${PN}-udhcpd"
```

Don't know how to prevent the file from getting installed though, but maybe this is sufficient?

chaitu236 (approved):

recipes-core/busybox/busybox_1.31.%.bbappend
gratian: @chaitu236 nice catch and thanks for the suggestions. Rebased on latest nilrt/master/dunfell and pushed a new commit that adds busybox-udhcpd to packagegroup-ni-base.bb with the modifications discussed here.

PR #121, opened 2021-02-03
recipes-ni: add ni-hw-support packages

amstewart:
Create a new family of packages - ni-hw-support* - which deliver support
scripts (initscripts) and tools needed specifically to support NI
product line hardware.

The first installment is the ni-rename-ifaces initscript, which has been
migrated (without modification) from os-common. It will be delivered via
the ni-hw-support-common package, since it is common to all product
lines.

Add the family's meta-package - ni-hw-support - to the base
packagegroup, so that it is installed on all base system images.

---

* The script-contents of `ni-rename-ifaces` is untouched.
* I expect that we can move more of the scripts which enable hardware support into this package family. Especially stuff like the [crio-support-scripts](https://github.com/ni/meta-nilrt/tree/nilrt/master/sumo/recipes-ni/crio-support-scripts/files). If everyone is OK with that, I will submit a follow-on commit where I move those scripts into `ni-hw-support-crio`.

---

## Testing
* built the `x64_dkms_image`
* Verified that the `ni-hw-support` and `ni-hw-support-common` packages are installed to the base.tar.
* Verified the contents of the `ni-hw-support-common` postinst and postrm scripts manually.
* Verified that the `ni-rename-ifaces` script and its update-rc link are in the correct place and have the correct permissions in the base.tar.

----

@ni/rtos


recipes-ni/ni-hw-support/ni-hw-support-common_1.0.bb
```diff
@@ -0,0 +1,54 @@
...
+}
+
+pkg_postinst_${PN} () {
+	if [ -n "$D" ]; then
+		OPT="-r $D"
+	else
+		OPT="-s"
+	fi
+
+	update-rc.d $OPT ni-rename-ifaces start 38 S .
```
gratian: Can we use the INITSCRIPT_NAME and INITSCRIPT_PARAMS provided by the update-rc.d.bbclass instead of open coding the update-rc.d call here?

amstewart: I wish. But unless I'm blind and stupid, the [`update-rc.d.bbclass`](https://github.com/ni/openembedded-core/blob/nilrt/master/sumo/meta/classes/update-rc.d.bbclass) only supports a single initscript per (sub)package. Technically that would work for this PR, but I assume that this package (and others in the same family) will install multiple initscripts - where the only solution is to call update-rc.d yourself.

shruthi-ravi: Why is this not part of the install stage? Asking to learn the difference, because other recipes like [initscripts-nilrt](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb) call update-rc.d in install, and do not call `update-rc.d remove` (bug?).

recipes-ni/ni-hw-support/ni-hw-support-common_1.0.bb
```diff
@@ -0,0 +1,54 @@
+SUMMARY = "Support scripts and utilities common to all NI hardware products"
```
shruthi-ravi: I'm concerned about the package name and the direction it takes when you add a hardware family-specific recipe like ni-hw-support-crio, which sounds like a driver package to me and can be confused for one. There are other hwsupport packages in the feed that provide driver support.

recipes-ni/ni-hw-support/ni-hw-support-common_1.0.bb
amstewart: >Why is this not part of the install stage? 

Execing `update-rc` in the `do_install` means that the `data.tar.gz` in the IPK will contain the `/etc/rc*.d` symlink. When you install the IPK, okpg will drop the symlink in the correct place; so in the simplest case, it is functionally equivalent. But putting the rc*.d symlink in the data section isn't *correct* in general package management, since you normally can't be sure that the user hasn't reconfigured how their init system works. It's more portable to let their native `update-rc.d` tooling do its job.

>because other recipes like initscripts-nilrt call update-rc.d in install, and do not call update-rc.d remove (bug?).

I would say `initscripts-nilrt` is incorrect, for the same reasons above. But if the RC-symlink is in the data section of the IPKS, then opkg should remove it at uninstall-time. So, explicitly calling `update-rc` in the postrm is unnecessary (but still wrong!) in that case.

recipes-ni/ni-hw-support/ni-hw-support-common_1.0.bb
amstewart: Granted. I'm trying to use this family to isolate files which enable NI hardware *as opposed to* generic x64 PC hardware. Do you have an alternative namespace in mind?

shruthi-ravi: ni-hwutils? ni-hw-utils? ni-hw-support-utils? Do these names support my case at all? :)

amstewart:
After discussing with the team, we have decided to refactor this package family's namespace to `ni-hw-scripts` to avoid namespace conflation with the CRIO driver packages. Refactor will be submitted in V2.

amstewart:
## Patch V2
* renamed `ni-hw-support` packages to `ni-hw-scripts`
* re-verified that the new recipes create the correct IPKs and checked that the contents of `ni-hw-scripts-common.ipk` still installs the `ni-rename-ifaces` initscript.

recipes-ni/ni-hw-support/ni-hw-support-common_1.0.bb
jpautler: Just occurred to me that ni-hw-integration could be another alternative if you weren't crazy about ni-hw-scripts. Very similar to ni-hw-support, but maybe without quite the same implication of driver support?

recipes-ni/ni-hw-support/ni-hw-support-common_1.0.bb
```diff
@@ -0,0 +1,54 @@
...
+	install -d ${D}${sysconfdir}/init.d/
+
+	install -m 0755 ${S}/init.d/ni-rename-ifaces ${D}${sysconfdir}/init.d/ni-rename-ifaces
+}
+
+pkg_postinst_${PN} () {
+	if [ -n "$D" ]; then
+		OPT="-r $D"
+	else
+		OPT="-s"
```
jpautler: What does -s do? Searching through other update-rc.d calls I don't see us using that option. Is there a reason we only use -s in the else case here?

amstewart: `-s` invokes the initscript at install-time, if the current runlevel is appropriate for it to have been executed. eg. If the script is supposed to be installed at runlevel 5, and you're installing it from RL 5, then the script will get invoked when the postinst is executed.

http://www.embeddedlinux.org.cn/OEManual/update-rc-d_class.html

recipes-ni/ni-hw-support/ni-hw-support-common_1.0.bb
amstewart: I would prefer `-scripts` to `-integration`.

shruthi-ravi (approved):

PR #122, opened 2021-02-04
Move from os-common to OE. Part 8

chaitu236:
Move file/dir permissions, links, stuff from os-common

@ni/rtos

### DIFF
--- a/recipes-extended/timezone/tzdata.bbappend
+++ b/recipes-extended/timezone/tzdata.bbappend
@@ -1,1 +1,6 @@
 INSTALL_TIMEZONE_FILE = "0"
+
+pkg_postinst_${PN}_append() {
+	# create symlink to shared default timezone settings (UTC)
+	ln -sf /etc/natinst/share/localtime $(D)/etc/localtime
+}
--- a/recipes-core/niacctbase/niacctbase.bb
+++ b/recipes-core/niacctbase/niacctbase.bb
@@ -24,7 +24,7 @@ USERADD_PARAM_${PN} = " -m -N -g ${LVRT_GROUP} -G network,niwscerts,plugdev,tty,
 		${@oe.utils.conditional('DISTRO', 'nilrt-nxg', '', d.getVar('USERADD_WEBSERVICES', True), d)}"
 
 useradd_preinst_append () {
-    eval ${PSEUDO} chmod g+s ${SYSROOT}/home/${LVRT_USER} || true
+    eval ${PSEUDO} chmod g+sw ${SYSROOT}/home/${LVRT_USER} || true
     # this logic is only for nilrt and nilrt-xfce
     if ${@oe.utils.conditional('DISTRO', 'nilrt-nxg', 'false', 'true', d)}; then
 	eval ${PSEUDO} chmod g+s ${SYSROOT}/home/webserv || true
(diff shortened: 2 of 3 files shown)


amstewart (approved):

recipes-extended/timezone/tzdata.bbappend
```diff
@@ -1 +1,7 @@
 INSTALL_TIMEZONE_FILE = "0"
+
+pkg_postinst_${PN}_append() {
```
shruthi-ravi, line 3: I just submitted populateconfig to OE and haven't had a chance to remove the file from Azdo. Can you make this change to the [OE version](https://github.com/ni/meta-nilrt/commit/045a9031ad0a48855ebdcb7b01b46da2608efbec#diff-5743703e7eae8822c18ab0c8a47929d9a776c75dc553d0683887ca55ac4b2954) too?

shruthi-ravi: Also, I think there was a reason the ssh, tpm folders were created in populateconfig. Let me investigate and get back to you.

shruthi-ravi: This is not the right place for this comment, but since I've already started here instead of in your Azdo PR, I will continue here. There's [ninetcfgutil](https://p4.natinst.com/browser/penguin/nilinux/os-common/trunk/8.7/source/safemode/ninetcfgutil?revision_at=11428760#line_520) called by [nisystemformat](https://p4.natinst.com/browser/penguin/nilinux/os-common/trunk/8.7/source/nisystemformat?revision_at=11428760#line_385), which removes the files in /etc/natinst/share during a format, and calls populateconfig to recreate them. This applies to the [ssh directory](https://p4.natinst.com/browser/penguin/nilinux/os-common/trunk/8.7/source/safemode/ninetcfgutil?revision_at=11428760#line_323) for sure, but I can't tell if these timezone files are also collateral damage.
Please correct me if I've missed something and this is all defunct code now.

chaitu236: Ok, I can remove ssh dir creation from base-files-nilrt.

timezone stuff is introduced by [this cl](https://review-board.natinst.com/r/29265/diff/2#index_header) which was a fix [for this](https://dev.azure.com/ni/DevCentral/_workitems/edit/751901). The bug only says the file was missing during fresh install, so I'm not sure why they added it to an initscript instead of base image. Assuming that was by error, I'd like to keep the link creation here and remove from populateconfig unless there are objections.

shruthi-ravi: Isn't the /etc/natinst/share directory the "config" partition that is preserved otherwise during formats? If that is true, it makes sense that the timezone is reset in populateconfig vs in the base image. Otherwise, the target would remain in the timezone before the format, after the format. I don't know the future for the config partition in RAUC, probably the above doesn't apply there. @jpautler?

gratian: Yes, /etc/natinst/share is the config partition that is preserved across a format. Files in there should be considered for migration via transconf. I haven't looked at the details here but I would expect the time zone setting to persist the same way we're preserving the network configuration.

gratian: I'll amend my statement slightly - we should preserve the existing behavior i.e. if it is preserved across a format in the current safemode/runmode implementation we need to continue to do so.

chaitu236: Removed stuff thats done in populateconfig for now.

chaitu236:
V2: Removed stuff thats handled in populateconfig

shruthi-ravi (approved):

PR #125, opened 2021-02-09
Reorganize bitbake conf settings

amstewart:
To reduce the complexity of the local.conf, move settings which are
frequently set by build machines or which should not be set by the user
at all, into other configuration files.

In this commit, distro.conf takes ownership of:
* the packaging system (via PACKAGE_CLASSES)
* the appropriate image recipes to keep via (rm_work)
* the default MACHINE value (x64)
* the useradd-staticids configuration

Other local.conf settings have been moved to the `auto.conf` and
`ni-org.conf` files in nilrt.git.

Also, deprecate the meta-orgconf layer.

---

### Testing

* Rebuilt `rtos_oe_feeds` on my dev machine and verified that the variables I moved have their correct values and that all makefile targets complete.

### Associated PRs
* nilrt [PR 38](https://github.com/ni/nilrt/pull/38) for the nilrt.git angle on this change.
* [AZDO](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/123493)

@ni/rtos


conf/layer.conf
```diff
@@ -10,3 +10,15 @@ BBFILE_PATTERN_meta-nilrt = "^${LAYERDIR}/"
 BBFILE_PRIORITY_meta-nilrt = "15"
 
 LAYERSERIES_COMPAT_meta-nilrt = "dunfell"
+
+
+## NILRT GIT REPO CONFIGURATION
+NILRT_GIT_RO_PROTO ?= "git://"
+NILRT_GIT_RW_PROTO ?= "https://"
```
shruthi-ravi: I don't believe the RW URLs are used anymore. And are the remaining declarations needed given NILRT_GIT already exists in [nilrt.inc](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/conf/distro/nilrt.inc#L104)?

conf/local.conf.sample
```diff
@@ -219,59 +171,13 @@ BB_DISKMON_DIRS ??= "\
...
-#
-# Qemu configuration
-#
-# By default qemu will build with a builtin VNC server where graphical output can be
-# seen. The two lines below enable the SDL backend too. By default libsdl-native will
-# be built, if you want to use your host's libSDL instead of the minimal libsdl built
-# by libsdl-native then uncomment the ASSUME_PROVIDED line below.
-PACKAGECONFIG_append_pn-qemu-native = " sdl"
-PACKAGECONFIG_append_pn-nativesdk-qemu = " sdl"
-#ASSUME_PROVIDED += "libsdl-native"
```
shruthi-ravi: I don't know what changes when you remove these.

amstewart: I think the answer is that nothing changes, since we don't use the qemu-native tools that upstream poky does. But I haven't confirmed that, so I will double-check. If I'm unsure that we can remove those lines, I will retain them in the local.conf - since they primarily affect the host-tools.

conf/layer.conf
amstewart: Hmm... you're right. I'll drop these configs.

I think it's actually more correct that `NILRT_GIT` be defined in the layer.conf - since it is specific to the layer recipes, rather than the distribution - but that's not important enough to warrant moving it at this time.

amstewart:
PATCH V2:
* addressed sravicha's comments
    * removed the obsolete NILRT GIT configs from layer.conf
    * re-added the qemu-native configs, because I'm not 100% sure they don't impact how the recipes are built.
    NOTE: I copied the qemu-native configs from the OE-core/dunfell version of local.conf, which has changed slightly since we forked our local.conf from it years ago. I don't think the changes they've made are significant.

amstewart:
PATCH V3:
* While testing changes to the build env init script, I discovered that the MACHINE variable has to be set before the distro.conf. Otherwise, bitbake will complain that `TUNE_ARCH` is undefined. Asserting `MACHINE=x64` in the environment works - which is what the AZDO component do - but otherwise, MACHINE needs to be defined in local.conf; so I moved it back there.

shruthi-ravi (approved):

PR #127, opened 2021-02-09
niwatchdogpet: Move watchdogpet later in startup

billpittman:
Move niwatchdogpet later in startup, S03->S04.  The script
calls fw_printenv which needs to be after populate-volatile.sh
otherwise we get errors about /var/lock/fw_printenv.lock.

NI-BUG-ID: 1182311


billpittman:
@ni/rtos

gratian (approved):

amstewart (approved):

shruthi-ravi (approved):
Hey @billpittman, can you add a link to the Azdo bug in the description so we have something to look back to?

PR #128, opened 2021-02-11
Move from os-common to OE. Part 10

chaitu236:
@ni/rtos


recipes-ni/ni-safemode-utils/files/bootimage.ini
```diff
@@ -0,0 +1,3 @@
+[BootImageDetails]
+component_version=20.7.0d0
+device_code=0x7755 0x774B 0x7735 0x774D 0x773B 0x7773 0x76CE 0x77AA 0x77AD 0x77B9 0x77BC 0x77BD 0x77DB 0x77DC 0x77DD 0x77E1 0x787E 0x787F 0x7828 0x7833 0x7835 0x7840 0x7841 0x7875 0x7878 0x7880 0x78C8 0x78C9 0x78CA 0x78CB 0x78CC 0x78CD 0x78CE 0x78CF 0x78E3 0x78E4 0x78E5 0x78E6 0x78E7 0x78E8 0x78E9 0x78EA 0x78EB 0x78EC 0x78B7 0x79D0 0x79CE 0x7A38 0x7A39 0x7A3A 0x7A3B 0x7A41 0x7A43 0x79DD 0x79DE 0x79DF 0x79E0 0x7A2A 0x7A2B
```
amstewart, line 3: What uses this file? Is this a component that we need to replicate in RAUC, or is it going to be totally deprecated?

chaitu236: Its used by [two files](https://codesearch.natinst.com/search/text?q=bootimage.ini%20repo%3Adunfell&fold_case=auto&regex=true&context=true) from the same recipe. I think this will be deprecated in RAUC. For now I'm only porting it for the scripts to not fail outright.

amstewart: ACK. Makes sense.

amstewart (approved):
lgtrm

recipes-ni/ni-safemode-utils/files/bootimage.ini
billpittman: If I recall, its an attempt to make sure we DON'T install the wrong safemode to a target (for fear of bricking the target).  With OneRT, that would seem to be unnecessary since we'll install the same base image to all targets.

chaitu236: Makes sense. Once we fix up all scripts using the scripts in ni-safemode-utils, we could remove the recipe outright.

billpittman (approved):

amstewart:
Is this ready to merge @chaitu236 ?

chaitu236:
@amstewart Yes it is

PR #129, opened 2021-02-11
Don't delete bootimage.cfg.d

chaitu236:
@ni/rtos

### DIFF
--- a/recipes-bsp/grub/grub-bootconf_%.bbappend
+++ b/recipes-bsp/grub/grub-bootconf_%.bbappend
@@ -8,8 +8,8 @@ SRC_URI += " \
 
 PACKAGES_prepend = "${PN}-nilrt"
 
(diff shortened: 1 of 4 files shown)


chaitu236:
@gratian I think I understand why /bootimage.cfg.d is used instead of /boot/bootimage.cfg.d - the recipe [moves everything under /boot to /](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-core/images/nilrt-base-bundle-image.bb#L63), so even if we put the cfg files in /boot/bootimage.cfg.d, they'd end up in /bootimage.cfg.d and grub.cfg has to find them in /bootimage.cfg.d.

jpautler (approved):
Changes make sense to me.

amstewart (approved):

gratian (approved):

PR #130, opened 2021-02-12
rt-tests: Fix PREEMPT_RT presence test

gratian:
The kernel_test_preempt_rt_presence.sh ptest incorrectly greps for the presence
of 'PREEMPT RT' string instead of the expected 'PREEMPT_RT' in the 'uname -a'
output (note the missing underscore).

In the process also simplify the test to avoid spawning a subshell and a call
to 'wc'.


gratian:
@ni/rtos 
FYI: I've stumbled over this bug while validating kernel updates and it was easy to fix inline.

amstewart (approved):

PR #131, opened 2021-02-12
Add nisetserialnumber and misc

chaitu236:
@ni/rtos


recipes-ni/initscripts-nilrt/files/nisetserialnumber
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# If the serial number is undefined in hardware, create one from a MAC address.
+
+. /etc/natinst/networking/functions.common
+
+SERIAL=`/sbin/fw_printenv -n serial# 2>&1`
+if [ "$SERIAL" = "ABCDEFG" ]; then
```
gratian: I wonder if this is still needed. I can't find any evidence that fw_printenv will return this when the serial# is not set.
I've checked /sbin/fw_printenv and /usr/share/nisysinfo/smbios_helper. Also checked a qemu VM where the serial number is not set when running 'dmidecode -t 1' and in that case 'fw_printenv -n serial#' returns '00123456' not 'ABCDEFG'

I think  we can probably drop this hack.

gratian (approved):
LGTM once the state of 'nisetserialnumber' is resolved.

recipes-ni/initscripts-nilrt/files/nisetserialnumber
chaitu236: The ABCDEFG probably came from [here](https://github.com/ni/meta-orgconf/blob/nilrt/master/dunfell/recipes-core/images/safemode-image/grubenv_non_ni_target#L2). We don't use this file in current image but I wonder if we should.
Serial number seems to be used to set hostname. If we drop this, wouldn't all VMs on network having the same hostname create problems?

jpautler: I think we'd like to keep this behavior of generating a unique hostname. It looks like that mechanism is still working on an os-common 8.6 exported VM where I mounted the .qcow2 file and saw ABCDEFG in the grubenv.

amstewart (changes requested):
This seems NI-hardware specific. Could you move it to the new [`ni-hw-scripts-common`](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-ni/ni-hw-scripts/ni-hw-scripts-common_1.0.bb) package?

recipes-ni/initscripts-nilrt/files/nisetserialnumber
chaitu236: Ok. Keeping it then

chaitu236:
Moved nisetserialnumber into ni-hw-scripts-common

amstewart (approved):
Looks good now; thanks.

PR #132, opened 2021-02-15
Move from os-common to OE. Part 11

chaitu236:
Moved code from systemlink/postinst

@ni/rtos

### DIFF
--- a/recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
+++ b/recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
@@ -59,6 +59,9 @@ do_install () {
 }
 
 pkg_postinst_ontarget_${PN} () {
+	# Add the memory reservation for Base
+	/usr/local/natinst/bin/nirtcfg -s section=RtLinuxMemReserve,token=Base,value=24
+
 	# Enable ssh by default
 	/usr/local/natinst/bin/nirtcfg -s section=systemsettings,token=sshd.enabled,value=True
 
--- a/recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
+++ b/recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
@@ -59,5 +59,8 @@ do_install () {
 }
 
 pkg_postinst_ontarget_${PN} () {
+	# Enable ssh by default
+	/usr/local/natinst/bin/nirtcfg -s section=systemsettings,token=sshd.enabled,value=True
+
 	${sysconfdir}/natinst/niselectsystemsettings postinst
 }
--- a/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
+++ b/recipes-ni/initscripts-nilrt/initscripts-nilrt_1.0.bb
@@ -133,6 +133,10 @@ pkg_postinst_ontarget_${PN} () {
 		fi
 	fi
 
+
+	# Enable core dumps on PXI, not on any other targets
+	[ "$class" = "PXI" ] && echo "* soft core unlimited" > /etc/security/limits.d/allow-core-dumps.conf
+
 	# Restore the original state of /boot
 	[ $mountstate == 0 ] && umount /boot || true
 }


recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb
```diff
@@ -38,6 +38,12 @@ DEPENDS += "niacctbase"
 do_install () {
 	install -d -m 0755 -o ${LVRT_USER} -g ${LVRT_GROUP} ${D}${sysconfdir}/natinst/share/
 
+	# Something in base is installing /etc/natinst/share/uninstall that may not be
+	# getting installed for SystemLink. nisid will fail if this directory doesn't
+	# install, so this component ought to install /etc/natinst/share/uninstall
+	# itself, regardless of if some other component ultimately will too.
+	install -d -m 0755 ${D}${sysconfdir}/natinst/share/uninstall
```
amstewart: Can we file a bug to NISID to create this directory, if they depend on it?

What's supposed to go in that directory? Uninstall hooks for NI packages?

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -59,5 +59,17 @@ do_install () {
...
+	if ! grep -q 'RtLinuxMemReserve' /etc/natinst/share/ni-rt.ini; then
+		cat >> /etc/natinst/share/ni-rt.ini <<EOF
+[RtLinuxMemReserve]
+Base=24
+EOF
+	fi
+
+	# Enable ssh by default
+	/usr/local/natinst/bin/nirtcfg -s section=systemsettings,token=sshd.enabled,value=True
+
```
amstewart, line 67: I assume this was unique to the SystemLink DKMS image, because SL requires SSH to be enabled for target configuration. But non-SL targets - I think - have operated on the assumption that SSH will not be enabled by default. So we should discuss as a group whether or not it is acceptable to change that assumption. I think the answer is yes; but we might want to organize some communication to that effect.

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -59,5 +59,17 @@ do_install () {
 }
 
 pkg_postinst_ontarget_${PN} () {
+	# Add the memory reservation for Base. We assert that if the section exists,
+	# it's already been added.
+	if ! grep -q 'RtLinuxMemReserve' /etc/natinst/share/ni-rt.ini; then
+		cat >> /etc/natinst/share/ni-rt.ini <<EOF
+[RtLinuxMemReserve]
+Base=24
+EOF
```
amstewart: What workflow is this setting actually enabling? Is there a more specific package that it can go into?

The comment seems like a bad reason to not use `nirtcfg` to set this setting. Unless the original RB makes some compelling argument why we expect this value to exist prior to base image install, I think we can change this to `nirtcfg --set`.

amstewart (changes requested):

recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb
shruthi-ravi: From what I can see, the contents of that directory are postrm rules for the ni_systemlink CDFs that used to be installed as dependencies of the System Image. We have since moved all the IPKs installed by the CDFs to the base image. There doesn't seem to be anything else in that directory. I think it is best we file a bug to nisid or work with them to figure out what else goes in this directory and if it is still relevant in RAUC. Here's a snapshot of the directory contents:
image

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
shruthi-ravi: Hasn't that assumption already been changed, given the SystemLink DKMS image is installed as the OneRT image on all OneRT-supported targets?

chaitu236: Good point. I'm not sure if it'll be evident to clients/customers that ssh is enabled by default. Maybe something that needs to go in release notes for customers.. and a post on General channel for internal clients?

amstewart: >Hasn't that assumption already been changed, given the SystemLink DKMS image is installed as the OneRT image on all OneRT-supported targets?

I... don't know. And that's concerning.

@chaitu236 could you bring this up to James and Brian either sideband on Teams or otherwise once Austin has stabilized. We need to ensure that whatever changes we make here are intentional.

In the meantime, I think it's probably true that this is was the case for the x64 OneRT images, so this review shouldn't be blocked.

recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb
jpautler: I don't have much to add, except that I think we've been told that NISID would be gone in RAUC images.

chaitu236: Ok. It looks like we don't need this dir so I'm removing this commit for now. We can add back later if thats not the case.

chaitu236:
update 2: Removed /etc/natinst/share/uninstall commit, set RtLinuxMemReserve settings with nirtcfg.

shruthi-ravi (approved):

amstewart (approved):

PR #133, opened 2021-02-23
initrdscripts: add ptest for ni_provisioning script

amstewart:

## IPK info
```
root@NI-Unknown-525400123456:/etc/opkg# opkg info init-restore-mode-ptest
Package: init-restore-mode-ptest
Version: 1.0-r0.1584
Depends: bash, efibootmgr
Recommends: init-restore-mode-lic, ptest-runner
Status: install ok installed
Section: devel
Architecture: x64
Maintainer: NI Linux Real-Time Maintainers <nilrt@ni.com>
MD5Sum: 670856f8d7c2117324598ba3059d51e4
Size: 2584
Filename: init-restore-mode-ptest_1.0-r0.1584_x64.ipk
Source: init-restore-mode_1.0.bb
Description: Extremely basic live image init script - Package test files
Installed-Size: 4102
Installed-Time: 1614100987
OE: init-restore-mode
License: MIT
```

## SAMPLE OUTPUT

*From a recently provisioned QEMU image:*
```bash
root@NI-Unknown-525400123456:/etc/opkg# ptest-runner
START: ptest-runner
2021-02-23T17:23
BEGIN: /usr/lib/init-restore-mode/ptest
## Checking that RAUC partitons are present...
test_part_id() /dev/sda1 niboota C12A7328-F81F-11D2-BA4B-00A0C93EC93B
test_part_id() /dev/sda2 nibootb C12A7328-F81F-11D2-BA4B-00A0C93EC93B
test_part_id() /dev/sda3 niuser 4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709
## Checking that niboot* contents are similar...
----
## Checking EFI bootorder...
BootCurrent: 0004
Timeout: 0 seconds
BootOrder: 0000,0001,0003,0004,0005,0006
Boot0000* EFI Floppy	PciRoot(0x0)/Pci(0x1,0x0)/Floppy(0x0)
Boot0001* EFI Floppy 1	PciRoot(0x0)/Pci(0x1,0x0)/Floppy(0x1)
Boot0003* EFI DVD/CDROM	PciRoot(0x0)/Pci(0x1,0x1)/Ata(1,0,0)
Boot0004* EFI Hard Drive	PciRoot(0x0)/Pci(0x1,0x1)/Ata(0,0,0)
Boot0005* EFI Network	PciRoot(0x0)/Pci(0x3,0x0)/MAC(525400123456,1)
Boot0006* EFI Internal Shell	MemoryMapped(11,0x900000,0x11fffff)/FvFile(7c04a583-9e3e-4f1c-ad65-e05268d0b4d1)
----
ERROR: efi_bootorder[0] is (0000, "EFI Floppy"); not one of: niboota, nibootb
ERROR: efi_bootorder[1] is (0001, "EFI Floppy 1"); not one of: niboota, nibootb
FAIL: ni_provisioning

ERROR: Exit status is 256
END: /usr/lib/init-restore-mode/ptest
2021-02-23T17:23
STOP: ptest-runner
```
* The bootorder failures are known bugs for the provisioning image.

## Testing

* Built the recipe and installed the new `-ptest` IPK to a VM.
* Verified that the IPK installs correctly and also installs the `ptest-runner` along with it.
* Removed the package and verified that the ptest contents are removed.

## Notes

* I haven't decided if this test is appropriate for the `ptest-smoke` packagegroup. We might like to explicitly install and run it only after the provisioning tool has run.

### DIFF
--- a/recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
+++ b/recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
@@ -1,0 +1,151 @@
+#!/bin/bash
+#
+# Test that the booted storage disk matches the partition scheme which we
+# expect from the ni_provisioning script. This ptest only supports the efi-ab
+# bootscheme present on new x64, RAUC-based targets.
+exec 2>&1
+
+
+PTEST_TEST='ni_provisioning'
+
+print_error() {
+	echo ERROR: $@ >&2
+	found_error=true
+}
+
+ptest_fail() {
+	echo "FAIL: $PTEST_TEST"
+	exit 1
+}
+
+ptest_pass() {
+	echo "PASS: $PTEST_TEST"
+	exit 0
+}
+
+ptest_skip() {
+	echo "SKIP: $PTEST_TEST"
+	exit 77
+}
+
+
+test_part_id () {
+	local device=${1}
+	local label=${2}
+	local uuid=${3}
+
+	local correct_re_label=${4}
+	local correct_re_uuid=${5}
+
+	echo "test_part_id() ${@:1:3}"
+
+	if [[ ! "${label}" =~ $correct_re_label ]]; then
+		print_error "Partition ${label} @ ${device} does not match required label ${correct_re_label}"
+		found_error=true
+	fi
+	if [[ ! "${uuid}" =~ $correct_re_uuid ]]; then
+		print_error "Partition ${label} @ ${device} does not match required type UUID ${correct_re_uuid}"
+		found_error=true
+	fi
+}
+
+
+found_error=false
+
+if uname --machine | grep -q arm; then
+	echo "INFO: detected an ARM machine architecture." \
+	     "This ptest is only valid for non-ARM, RAUC systems."
+	ptest_skip
+fi
+
+
+## PARTITION TESTS ##
+echo "## Checking that RAUC partitons are present..."
+RE_ESP_UUID=C12A7328\-F81F\-11D2\-BA4B\-00A0C93EC93B
+
+# Determine the configured /boot storage device. Assert that this device should
+# be the root of the partitions we evaluate in the remainder of this test.
+# For RAUC systems, we expect this to be: /dev/sda, but it isn't required to be.
+boot_part=$(findmnt --kernel --first-only --noheadings --output SOURCE /boot)
+test -n "${boot_part}" || (print_error "No detected /boot mount; system state invalid."; ptest_fail)
+boot_dev=/dev/$(lsblk -no PKNAME ${boot_part})
+test -e "${boot_dev}" || (print_error "Boot device ${boot_dev} does not exist; system state invalid."; ptest_fail)
+
+
+readarray -t parts < <(fdisk -l ${boot_dev} -o Device,Name,Type-UUID | grep -o -E '^\/dev\/.*')
+
+# ASSERT: part 0 is: niboota and is an ESP partition
+test_part_id ${parts[0]} niboota $RE_ESP_UUID
+# ASSERT: part 1 is: nibootb and is an ESP partition
+test_part_id ${parts[1]} nibootb $RE_ESP_UUID
+# ASSERT: part 2 is: niuser and is any partition type
+test_part_id ${parts[2]} niuser '.*'
+
+# bail out if the partition tests failed
+$found_error && ptest_fail
+
+
+## PARTITION CONTENTS TESTING ##
+echo "## Checking that niboot* contents are similar..."
+set -eo pipefail
+
+# Unmount the read-only partitions and remove their empty directories
+unmount_ro_partitions () {
+	set +e
+	for part in $@; do
+		[ -d ${part} ] && umount -f ${part} && rmdir ${part}
+	done
+}
+
+# After provisioning, niboota and nibootb should contain exactly the same contents.
+mnt_a=$(mktemp -d --tmpdir $PTEST_TEST.niboota.XXXXX)
+mnt_b=$(mktemp -d --tmpdir $PTEST_TEST.nibootb.XXXXX)
+
+trap "unmount_ro_partitions ${mnt_a} ${mnt_b}; exit 99" EXIT
+
+mount -o ro -L niboota ${mnt_a}
+mount -o ro -L nibootb ${mnt_b}
+
+diff --recursive ${mnt_a} ${mnt_b} || \
+	print_error "niboot partition contents differ appreciably."
+echo "----"
+
+trap - EXIT
+unmount_ro_partitions ${mnt_a} ${mnt_b}
+
+
+## EFI VARIABLE TESTS ###
+echo "## Checking EFI bootorder..."
+found_niboota=false
+found_nibootb=false
+
+efibootmgr -v  # INFO
+IFS=',' read -a efi_bootorder < <(efibootmgr | grep '^BootOrder: ' | cut -d' ' -f2)
+
+echo "----"
+# ASSERT: bootpriorities 0 and 1 are niboota and nibootb, in any order.
+for bootnum in 0 1; do
+	bootorder=${efi_bootorder[${bootnum}]}
+	bootlabel=$(efibootmgr | grep "^Boot${bootorder}\*" | cut -d' ' -f2-)
+	case $bootlabel in
+		niboota)
+			found_niboota=true
+			echo 'INFO: Found niboota in bootorder.'
+			;;
+		nibootb)
+			found_nibootb=true
+			echo 'INFO: Found nibootb in bootorder.'
+			;;
+		*)
+			print_error "efi_bootorder[${bootnum}] is (${bootorder}, \"${bootlabel}\"); not one of: niboota, nibootb"
+			found_error=true
+			;;
+	esac
+done
+# fail if either a or b was not found
+($found_niboota && $found_nibootb) || found_error=true
+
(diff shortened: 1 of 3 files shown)


amstewart:
@ni/rtos

amstewart:
@jpautler

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
```diff
@@ -0,0 +1,151 @@
...
+if uname --machine | grep -q arm; then
+	echo "INFO: detected an ARM machine architecture." \
+	     "This ptest is only valid for non-ARM, RAUC systems."
+	ptest_skip
+fi
+
+
+## PARTITION TESTS ##
+echo "## Checking that RAUC partitons are present..."
+RE_ESP_UUID=C12A7328\-F81F\-11D2\-BA4B\-00A0C93EC93B
```
billpittman, line 66: Please put a comment in here about where/what this is for future reference.  Look at meta-nilrt/recipes-core/initrdscripts/files/disk_config_x64 for reference.

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
```diff
@@ -0,0 +1,151 @@
...
+
+
+readarray -t parts < <(fdisk -l ${boot_dev} -o Device,Name,Type-UUID | grep -o -E '^\/dev\/.*')
+
+# ASSERT: part 0 is: niboota and is an ESP partition
+test_part_id ${parts[0]} niboota $RE_ESP_UUID
+# ASSERT: part 1 is: nibootb and is an ESP partition
+test_part_id ${parts[1]} nibootb $RE_ESP_UUID
+# ASSERT: part 2 is: niuser and is any partition type
+test_part_id ${parts[2]} niuser '.*'
```
billpittman: Aren't we using a "Root Partition here 4F68BCE3-...." from the same reference as above?

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
```diff
@@ -0,0 +1,151 @@
...
+mnt_a=$(mktemp -d --tmpdir $PTEST_TEST.niboota.XXXXX)
+mnt_b=$(mktemp -d --tmpdir $PTEST_TEST.nibootb.XXXXX)
+
+trap "unmount_ro_partitions ${mnt_a} ${mnt_b}; exit 99" EXIT
+
+mount -o ro -L niboota ${mnt_a}
+mount -o ro -L nibootb ${mnt_b}
+
+diff --recursive ${mnt_a} ${mnt_b} || \
+	print_error "niboot partition contents differ appreciably."
```
billpittman, line 135: They don't contain exactly the same contents.  The booted partition has a RAUC text file that has some partition information RAUC creates.  I don't see that being handled here.

billpittman (changes requested):
kudos - Nice clear test, easy to follow.

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
amstewart: I didn't see that file on the test system I was using. Do you have a reference for what that file is supposed to contain, or when it gets made?

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
amstewart: Hmm... I suppose that since it's statically defined, we should test for it. Will fix.

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
amstewart: Will fix.

billpittman:
Another test we should probably add to this is to test the existence and assignments of /dev/niboot/*.  Some of the rauc provisioning scripts use those links, so we should check them.

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
billpittman: Strange, just booted a VM to find the exact name of the file, but it wasn't there either.  There used to be a file called slot.raucs that RAUC puts on the boot partition.  I'm not sure when/how/why the file exists, but it always gets created.  HOWEVER, its not there anymore.  I guess leave it off for now, the test will fail if/when the file returns, so we'll know about it.  In case you doubt me, you can look at rauc-mark-good where we make the backup and see that we explicitly remove that file from the backup.

amstewart:
## Patch V2
* Addressed feedback from @billpittman.
The new RAUC device link test output looks like:
```bash
## Checking for RAUC partition device links in /dev...
/dev/niboot/niboot.current [OK]
/dev/niboot/niboot.other [OK]
/dev/niboot/niboota [OK]
/dev/niboot/nibootb [OK]
/dev/niboot/niuser [OK]
/dev/niboot/does-not-exist [FAIL]
ERROR: Device link "/dev/niboot/does-not-exist" does not exist.
## Checking that niboot* contents are similar...
```

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
```diff
@@ -0,0 +1,176 @@
...
+)
+for link in ${niboot_links[@]}; do
+	if [ -L "$link" ]; then
+		echo "${link} [OK]"
+	else
+		echo "${link} [FAIL]"
+		print_error "Device link \"$link\" does not exist."
+	fi
+done
+# the link subtests are not critical failures, so keep testing, even if they fail.
```
billpittman, line 110: Great, however they are critical - at least to RAUC (cat /etc/rauc/system.conf).  RAUC wouldn't find the slots if these aren't set properly.  rauc-mark-good also uses them.  The initramfs creates the links and uses them itself as part of the boot process, so one could argue that if you're booted and running a ptest, then they're at least good enough, but there's no harm in actually testing and caring about them.

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
```diff
@@ -0,0 +1,176 @@
...
+}
+
+# After provisioning, niboota and nibootb should contain exactly the same contents.
+mnt_a=$(mktemp -d --tmpdir $PTEST_TEST.niboota.XXXXX)
+mnt_b=$(mktemp -d --tmpdir $PTEST_TEST.nibootb.XXXXX)
+
+trap "unmount_ro_partitions ${mnt_a} ${mnt_b}; exit 99" EXIT
+
+mount -o ro -L niboota ${mnt_a}
+mount -o ro -L nibootb ${mnt_b}
```
billpittman, line 132: If you don't want to check the /dev/niboot/* links, then lets at least change this code to mount /dev/niboot/niboota and /dev/niboot/nibootb.

billpittman (approved):

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
shruthi-ravi: @amstewart do you have an upcoming fix for this? If the tests are all going to be automated, I don't see how a test that could error but not fail add value. Noone is probably going to pay attention to the stdout unless there is a failure. So, I'd say either keep and fail the test, or remove.

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
```diff
@@ -0,0 +1,176 @@
...
+#
+# Test that the booted storage disk matches the partition scheme which we
+# expect from the ni_provisioning script. This ptest only supports the efi-ab
+# bootscheme present on new x64, RAUC-based targets.
+exec 2>&1
+
+
+PTEST_TEST='ni_provisioning'
+
+print_error() {
```
shruthi-ravi, line 11: Non-blocking for this review: In the interest of parsing test output, can these functions be made available to future tests? This and the [OK], [FAIL], INFO logs.

shruthi-ravi (approved):
lgtm but for the issue raised by Bill that I've commented on previously.

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
billpittman: @amstewart and I discussed this, the comment is a little misleading.  The test WILL fail, however the test continues to run the remainder of the tests.  IE The test doesn't bail out just because of this part failing.

shruthi-ravi: Does the ptest runner pick up the [FAIL] and ERROR messages to report a failure, or do we need to parse that as a failure when we set up the test runs? I guess I don't understand how the runner reports failures. @amstewart, if you want I can follow up with you offline.

recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh
amstewart: Yeah; we should write a standard ptest library at some poiint. I kind-of wrote one for the [kernel-tests](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-kernel/kernel-tests/kernel-tests-files/ptest-format.sh) packages, but didn't have the opportunity to make it into a shared library for other packages.

I intend on addressing that work in the later API guarantee features.

PR #139, opened 2021-03-09
Dev/nilrt/master/dunfell/dist nilrt/grub

billpittman:
@ni/rtos Note to reviewers, the files in the safemode-restore-image commit were mostly rescued from the defunct meta-orgconf layer and simply moved to meta-nilrt and edited there.


recipes-core/dist-nilrt/files/nilrt-gateway-install
```diff
@@ -38,29 +38,36 @@ function die() {
...
-pushd $NIROOTFS_MOUNT/payload > /dev/null
-opkg update > /dev/null
-opkg download dist-nilrt-efi-ab > /dev/null || (popd > /dev/null; die "Unable to download dist-nilrt-efi-ab. Check if disk has sufficient space.")
-test -e dist-nilrt-efi-ab*.ipk || (popd > /dev/null; die "Unable to download dist-nilrt-efi-ab. Check if feeds contain dist-nilrt-efi-ab package.")
-popd > /dev/null
-
 # Mount the ISO/wic file as a loopback device
 info "Mounting the ISO"
 if [ -f "$SCRIPT_DIR/restore-mode-image-x64.iso" ] ; then
+    # safemode->efi-ab migration
```
amstewart: NIT: let's prefer `efi-grub` to `safemode` in these comments; it will make more sense to the future maintainers and community, I hope.

recipes-core/dist-nilrt/files/nilrt-gateway-install
```diff
@@ -38,29 +38,36 @@ function die() {
...
+    EXTERNAL_PAYLOAD="dist-nilrt-efi-ab"
     losetup -f "$SCRIPT_DIR/restore-mode-image-x64.iso" -P -r
     LOOPDEV=$(losetup -j "$SCRIPT_DIR/restore-mode-image-x64.iso" |cut -f 1 -d:)
 else
+    # efi-ab->safemode migration
     losetup -f "$SCRIPT_DIR/safemode-restore-image-x64.iso" -P -r
     LOOPDEV=$(losetup -j "$SCRIPT_DIR/safemode-restore-image-x64.iso" |cut -f 1 -d:)
 fi
 
+if [ ! -z "${EXTERNAL_PAYLOAD:-}" ] ; then
```
amstewart: NIT: I think `[ ! -z "" ]` and `[ -n "" ]` are always equivalent, but the latter version is probably more clear.

amstewart (approved):
Looks pretty good to me. A few nits below.

chaitu236 (approved):

recipes-core/dist-nilrt/files/nilrt-gateway-install
billpittman: Good point, thanks, it's changed.

recipes-core/dist-nilrt/files/nilrt-gateway-install
billpittman: Done.

recipes-core/initrdscripts/files/init-restore-mode.sh
```diff
@@ -106,50 +106,59 @@ mount_nirecovery_usb()
...
-      mv /payload/*.raucb /payload/niboot.raucb
-      popd > /dev/null
-
-      umount $ROOTFS_MOUNTPOINT;
+   # Locate and mount the payload at /payload (in the case of initrd having the payload, do nothing)
+   # 1) Check in the current directory (ie built into the initrd as with rauc->safemode migration)
+   # 2) Check in the boot media (as with USB provisioning)
+   # 3) Check on the "big" partition (as with safemode->rauc migrations)
+
+   if [ -z "$(ls -A /payload 2>dev/null)" ] ; then
```
shruthi-ravi: You're missing the forward slash in 'dev/null' here and in the line below.

gratian (approved):
LGTM minus the issues already flagged by @shruthi-ravi

recipes-core/initrdscripts/files/init-restore-mode.sh
billpittman: Nice catch - both fixed!

shruthi-ravi (approved):

jpautler (changes requested):
I don't see a way to comment on a collapsed line (that wasn't changed by the commit), so I'll just comment here...

On line 178 of init-restore-mode.sh, it checks $ARCH and allows x64 and ARM. Since we've decided to lock ARM/os-common at Sumo, could we change this to be x64 only. It would make the code more explicit about what we actually support.

billpittman:
Removed the reference to ARM as Jeff suggested.

jpautler (approved):

PR #140, opened 2021-03-10
Add ptest for safemode migration

chaitu236:
@ni/rtos 

Output of run-ptest

```
(safemode) admin@NI-cRIO-903x-VM-29af4652:/usr/lib/dist-nilrt-grub-gateway/ptest# ./run-ptest 
## Checking that RAUC partitons are present...
test_part_id() /dev/sda1 nigrub C12A7328-F81F-11D2-BA4B-00A0C93EC93B
test_part_id() /dev/sda2 nibootfs 0FC63DAF-8483-4772-8E79-3D69D8477DE4
test_part_id() /dev/sda3 niconfig 0FC63DAF-8483-4772-8E79-3D69D8477DE4
test_part_id() /dev/sda4 nirootfs 0FC63DAF-8483-4772-8E79-3D69D8477DE4
## Checking that nibootfs contents are as expected...
/boot/bootmode [OK]
/boot/.safe/bzImage [OK]
/boot/.safe/bootimage.ini [OK]
/boot/.safe/EFI_NI_vars [OK]
/boot/.safe/GRUB_NI_readonly_vars [OK]
/boot/.safe/bootimage.cfg [OK]
/boot/.safe/SMBIOS_NI_vars [OK]
/boot/.safe/ramdisk.gz [OK]
/boot/.safe/ramdisk.xz [OK]
/boot/grub/recoverytool-ni-version [OK]
/boot/grub/grub-ni-version [OK]
/boot/grub/grubenv [OK]
PASS: ni_provisioning
(safemode) admin@NI-cRIO-903x-VM-29af4652:/usr/lib/dist-nilrt-grub-gateway/ptest# 
```

NOTE: Depends on https://github.com/ni/meta-nilrt/pull/139 to build.

### DIFF
--- a/recipes-core/dist-nilrt/dist-nilrt-grub-gateway.bb
+++ b/recipes-core/dist-nilrt/dist-nilrt-grub-gateway.bb
@@ -7,6 +7,7 @@ ALLOW_EMPTY_${PN}-dev = "0"
 
 SRC_URI += "\
     file://nilrt-gateway-install \
+    file://ptest \
 "
 
 DEPENDS = "safemode-image"
@@ -33,6 +34,18 @@ python do_package_prepend() {
              line = fp.readline()
     if found != True:
         bb.fatal("Safemode version not found (check safemode-image recipe) !!!")
+}
+
+inherit ptest
+
+RDEPENDS_${PN}-ptest += "bash"
+# The ptests should be run on a system which has already been provisioned, so a
+# dependency on the migration IPK is not necessary.
+RDEPENDS_${PN}-ptest_remove += "${PN}"
+
+do_install_ptest_append_x64() {
+    install -m 0755 ${WORKDIR}/ptest/run-ptest ${D}${PTEST_PATH}
+    install -m 0755 ${WORKDIR}/ptest/test_safemode_provisioning.sh ${D}${PTEST_PATH}
 }
 
 FILES_${PN} = "\
--- a/recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
+++ b/recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+#
+# Test that the booted storage disk matches the partition scheme which we
+# expect from the ni_provisioning script. This ptest only supports the
+# bootscheme present on x64, safemode-based targets.
+exec 2>&1
+
+
+PTEST_TEST='ni_provisioning'
+
+print_error() {
+	echo ERROR: $@ >&2
+	found_error=true
+}
+
+ptest_fail() {
+	echo "FAIL: $PTEST_TEST"
+	exit 1
+}
+
+ptest_pass() {
+	echo "PASS: $PTEST_TEST"
+	exit 0
+}
+
+ptest_skip() {
+	echo "SKIP: $PTEST_TEST"
+	exit 77
+}
+
+
+test_part_id () {
+	local device=${1}
+	local label=${2}
+	local uuid=${3}
+
+	local correct_re_label=${4}
+	local correct_re_uuid=${5}
+
+	echo "test_part_id() ${@:1:3}"
+
+	if [[ ! "${label}" =~ $correct_re_label ]]; then
+		print_error "Partition ${label} @ ${device} does not match required label ${correct_re_label}"
+	fi
+	if [[ ! "${uuid}" =~ $correct_re_uuid ]]; then
+		print_error "Partition ${label} @ ${device} does not match required type UUID ${correct_re_uuid}"
+	fi
+}
+
+
+found_error=false
+
+if uname --machine | grep -q arm; then
+	echo "INFO: detected an ARM machine architecture." \
+		  "This ptest is only valid for non-ARM, safemode systems."
+	ptest_skip
+fi
+
+
+## PARTITION TESTS ##
+echo "## Checking that safemode partitons are present..."
+# The UUIDs here should match those set by the disk_config_x64 file.
+# https://github.com/ni/meta-nilrt/blob/983d50796302b394aa428b992dae48b4da32ff08/recipes-core/initrdscripts/files/disk_config_x64#L28
+RE_ESP_UUID=C12A7328\-F81F\-11D2\-BA4B\-00A0C93EC93B
+RE_X64_OTHER_UUID=0FC63DAF\-8483\-4772\-8E79\-3D69D8477DE4
+
+# Determine the configured /boot storage device. Assert that this device should
+# be the root of the partitions we evaluate in the remainder of this test.
+boot_part=$(findmnt --kernel --first-only --noheadings --output SOURCE /boot)
+test -n "${boot_part}" || { print_error "No detected /boot mount; system state invalid."; ptest_fail; }
+boot_dev=/dev/$(lsblk -no PKNAME ${boot_part})
+test -e "${boot_dev}" || { print_error "Boot device ${boot_dev} does not exist; system state invalid."; ptest_fail; }
+
+
+readarray -t parts < <(fdisk -l ${boot_dev} -o Device,Name,Type-UUID | grep -o -E '^\/dev\/.*')
+
+# ASSERT: Only 4 partitions exist
+expected_partitions=4
+test ${#parts[@]} = $expected_partitions || { print_error "Expected $expected_partitions partitions, found ${#parts[@]}"; ptest_fail; }
+
+# ASSERT: part 0 is: nigrub and is an ESP partition
+test_part_id ${parts[0]} nigrub $RE_ESP_UUID
+# ASSERT: part 1 is: nibootfs and is a data partition type
+test_part_id ${parts[1]} nibootfs $RE_X64_OTHER_UUID
+# ASSERT: part 2 is: niconfig and is a data partition type
+test_part_id ${parts[2]} niconfig $RE_X64_OTHER_UUID
+# ASSERT: part 3 is: nirootfs and is a data partition type
+test_part_id ${parts[3]} nirootfs $RE_X64_OTHER_UUID
+
+part_1=( ${parts[1]} )
+test $boot_part = ${part_1[0]} || { print_error "booted from ${boot_part} partition instead of ${part_1[0]} (nibootfs)"; ptest_fail; }
+
+# bail out if the partition tests failed
+$found_error && ptest_fail
+
+
+## PARTITION CONTENTS TESTING ##
+echo "## Checking that nibootfs contents are as expected..."
+safemode_files=(                       \
+	/boot/bootmode                      \
+	/boot/.safe/bzImage                 \
+	/boot/.safe/bootimage.ini           \
+	/boot/.safe/EFI_NI_vars             \
+	/boot/.safe/GRUB_NI_readonly_vars   \
+	/boot/.safe/bootimage.cfg           \
+	/boot/.safe/SMBIOS_NI_vars          \
+	/boot/.safe/ramdisk.gz              \
+	/boot/.safe/ramdisk.xz              \
+	/boot/grub/recoverytool-ni-version  \
+	/boot/grub/grub-ni-version          \
+	/boot/grub/grubenv                  \
+)
+
+for file in ${safemode_files[@]}; do
+	if [ -e "$file" ]; then
+		echo "${file} [OK]"
+	else
+		echo "${file} [FAIL]"
+		print_error "File \"$file\" does not exist."
+	fi
+done
+
+$found_error && ptest_fail
+
+ptest_pass
--- a/recipes-core/dist-nilrt/files/ptest/run-ptest
+++ b/recipes-core/dist-nilrt/files/ptest/run-ptest
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+SCRIPT_ROOT=$(dirname $0)
+
+${SCRIPT_ROOT}/test_safemode_provisioning.sh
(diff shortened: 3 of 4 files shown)


recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+#
+# Test that the booted storage disk matches the partition scheme which we
+# expect from the ni_provisioning script. This ptest only supports the efi-ab
+# bootscheme present on new x64, RAUC-based targets.
```
billpittman: I think some of the comments are backwards and still refer to RAUC.

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
...
+		found_error=true
+	fi
+}
+
+
+found_error=false
+
+if uname --machine | grep -q arm; then
+	echo "INFO: detected an ARM machine architecture." \
+		  "This ptest is only valid for non-ARM, RAUC systems."
```
billpittman: Grub systems ... right?

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
...
+	ptest_skip
+fi
+
+
+## PARTITION TESTS ##
+echo "## Checking that RAUC partitons are present..."
```
billpittman: "Checking that expected partitions are present ..."  ???

billpittman (changes requested):

billpittman:
Can you add some test output to the description please ?

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
...
+test_part_id ${parts[0]} nigrub $RE_ESP_UUID
+# ASSERT: part 1 is: nibootfs and is a data partition type
+test_part_id ${parts[1]} nibootfs $RE_X64_OTHER_UUID
+# ASSERT: part 2 is: niconfig and is a data partition type
+test_part_id ${parts[2]} niconfig $RE_X64_OTHER_UUID
+# ASSERT: part 3 is: nirootfs and is a data partition type
+test_part_id ${parts[3]} nirootfs $RE_X64_OTHER_UUID
+
+part_1=( ${parts[1]} )
+test $boot_part = ${part_1[0]} || { print_error "booted from ${part_1[0]} partition instead of nibootfs"; ptest_fail; }
```
amstewart: This works; but is there a reason that `part_1` is storing a single value as an array? Why not just do
`part_1="${parts[1]}"`? Then you wouldn't have to index `${part_1}` every time you use it.

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
...
+	/boot/.safe/bootimage.ini           \
+	/boot/.safe/EFI_NI_vars             \
+	/boot/.safe/GRUB_NI_readonly_vars   \
+	/boot/.safe/bootimage.cfg           \
+	/boot/.safe/SMBIOS_NI_vars          \
+	/boot/.safe/ramdisk.gz              \
+	/boot/.safe/ramdisk.xz              \
+	/boot/grub/recoverytool-ni-version  \
+	/boot/grub/grub-ni-version          \
+	/boot/grub/grubenv                  \
```
amstewart, line 111: Style: I don't think it's worthwhile to try and align the line-escape characters like this.
1. It means that every time someone adds an entry which is even 1 character longer than the previous-longest entry, their diff will have to touch every entry to fixup the alignment.
2. The first line in this assignment block (`safemode_files=...`) is only ever going to align to the array elements coincidentally, because all the entry lines begin with a tab character - which is appropriate for bash styling.

amstewart (approved):
>Use {} instead of () so exit runs in same shell and exits from the script instead of a subshell created by ().

D'oh! Good catch.

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
chaitu236: Ugh sorry missed them. Fixed

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
chaitu236: Used "safemode" as both schemes use grub

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
chaitu236: part_1 is actually storing a space delineated string

```
(safemode) admin@NI-cRIO-903x-VM-29af4652:~# echo ${parts[1]}
/dev/sda2 nibootfs 0FC63DAF-8483-4772-8E79-3D69D8477DE4
```

Out of this we only want /dev/sda2

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
chaitu236: Good point. But it's easier to parse through the lines when the '\'s are not in the way. And we don't expect to add anymore lines to this list as safemode is frozen.
You're right about the first line but since we never change how many tab characters would go before the array element, and we don't expect to add new lines, so this wouldn't need to be realigned in future.

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
amstewart: Ah; I see. That makes sense.

gratian (approved):

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
...
+test_part_id ${parts[0]} nigrub $RE_ESP_UUID
+# ASSERT: part 1 is: nibootfs and is a data partition type
+test_part_id ${parts[1]} nibootfs $RE_X64_OTHER_UUID
+# ASSERT: part 2 is: niconfig and is a data partition type
+test_part_id ${parts[2]} niconfig $RE_X64_OTHER_UUID
+# ASSERT: part 3 is: nirootfs and is a data partition type
+test_part_id ${parts[3]} nirootfs $RE_X64_OTHER_UUID
+
+part_1=( ${parts[1]} )
+test $boot_part = ${part_1[0]} || { print_error "booted from ${part_1[0]} partition instead of nibootfs"; ptest_fail; }
```
shruthi-ravi: Hey @chaitu236, can you explain to me why this test is needed and when it would fail?

shruthi-ravi: Should the message have been "booted from $boot_part partition instead of ${part_1[0]}/nibootfs"? Then I think it makes sense to me.

recipes-core/dist-nilrt/dist-nilrt-grub-gateway.bb
```diff
@@ -35,6 +36,18 @@ python do_package_prepend() {
         bb.fatal("Safemode version not found (check safemode-image recipe) !!!")
 }
 
+inherit ptest
+
+RDEPENDS_${PN}-ptest += "bash"
+# The ptests should be run on a system which has already been provisioned, so a
+# dependency on the provisioning scripts is not necessary.
```
jpautler: Do you mean "migration IPK" (or similar) here instead of "provisioning scripts"?

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+#
+# Test that the booted storage disk matches the partition scheme which we
+# expect from the ni_provisioning script. This ptest only supports the efi-grub
```
jpautler: I would replace "efi-grub" with just "grub" or remove it entirely. I realize you are just fixing up the comment that came from test_ni_provisioning.sh. I'm not certain, but in that case I think "efi-ab" is referring to the "efi-ab" provisioning scheme defined by our [provisioning scripts](https://github.com/ni/meta-nilrt/blob/nilrt/master/dunfell/recipes-core/initrdscripts/files/ni_provisioning.answers.default#L12). The only other named provisioning scheme is "efi-auto" which we are no longer using.

On a related note, "efi-ab" is also the shorthand for "RAUC" that we use in naming our dist-nilrt-**efi-ab**-* IPKs. The corresponding shorthand for "safemode" is "grub" as in dist-nilrt-**grub**-*.ipk. I don't care for that grub=safemode association, but it is what it is and is unlikely to change at this point...

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
...
+
+	if [[ ! "${label}" =~ $correct_re_label ]]; then
+		print_error "Partition ${label} @ ${device} does not match required label ${correct_re_label}"
+		found_error=true
+	fi
+	if [[ ! "${uuid}" =~ $correct_re_uuid ]]; then
+		print_error "Partition ${label} @ ${device} does not match required type UUID ${correct_re_uuid}"
+		found_error=true
+	fi
+}
```
jpautler: This is the second test to make use of these functions. Do we think there will be enough reuse to justify moving some/all of this into a shared file?

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
...
+	local uuid=${3}
+
+	local correct_re_label=${4}
+	local correct_re_uuid=${5}
+
+	echo "test_part_id() ${@:1:3}"
+
+	if [[ ! "${label}" =~ $correct_re_label ]]; then
+		print_error "Partition ${label} @ ${device} does not match required label ${correct_re_label}"
+		found_error=true
```
jpautler: Do we need found_error=true here? Looks like it is already set by print_error. Same issue 4 lines below.

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
...
+	/boot/grub/grubenv                  \
+)
+
+for file in ${safemode_files[@]}; do
+	if [ -e "$file" ]; then
+		echo "${file} [OK]"
+	else
+		echo "${file} [FAIL]"
+		print_error "File \"$file\" does not exist."
+		found_error=true
```
jpautler: Same as above: is found_error=true needed here?

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
```diff
@@ -0,0 +1,130 @@
...
+	fi
+done
+
+set -eo pipefail
```
jpautler: I don't think you need this line. If found_error=true, then ptest_fail will run and "exit 1". If found_error=false, the && will short-circuit and ptest_pass will "exit 0".

jpautler (changes requested):

billpittman (approved):

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
chaitu236: ~~I'm not sure it'd be worth our effort to try to single source these when the files using them are in very different locations~~

chaitu236: test_part_id may not be useful for other ptests. The other functions may be useful but these are just four and really short, so may not be worth the effort to single-source them.

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
chaitu236: It seemed prudent to also test that we booted from the expected partition and not just that partition layout is as expected.
You're right, it should be $boot_part

recipes-core/dist-nilrt/dist-nilrt-grub-gateway.bb
chaitu236: Yes, thanks

chaitu236:
Update: Addressed feedback

shruthi-ravi (approved):

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
jpautler: Thanks for cleaning up this same code in test_ni_provisioning.sh as well!

jpautler (approved):

recipes-core/dist-nilrt/files/ptest/test_safemode_provisioning.sh
amstewart: @jpautler There is a growing desire to have a standard library for these kinds of functions (across this and other PRs). I'll probably get drunk and write one for bash and python3, but I think we have agreed that it shouldn't block this initial RAUC work.

PR #141, opened 2021-03-16
Revert PR 139

amstewart:
Something in [PR 139](https://github.com/ni/meta-nilrt/pull/139) is causing the latest VM builds to hang indefinitely during provisioning. I think it's the result of the [init-restore-mode.sh](https://github.com/ni/meta-nilrt/pull/139/files#diff-ebbcd813a09a31181820adb9e0c1d58eee4a8386874e68eafc9560c27df466f8) changes that went in, but I'm having general difficulty getting the VM to provide me a shell at the moment - so debugging is slow.

Revert commit [af1c19f6bbd59ef9a9bf56926906c038b3ad5481](https://github.com/ni/meta-nilrt/pull/139/commits/af1c19f6bbd59ef9a9bf56926906c038b3ad5481) to remove the init script change and the rest of the PR for cleanliness.

@ni/rtos  @billpittman


billpittman:
Keep me posted on this issue, I was able to migrate the VM several times, but honestly never _CREATED_ a new VM, so it could be something I changed, so feel free to dump this on me if/when you decide that reverting this change fixed it.

billpittman (approved):

PR #144, opened 2021-03-22
dist-nilrt-efi-ab-gateway: Disable -lic package creation

shruthi-ravi:
The recipe as is creates a separate accompanying -lic package and
adds a "RECOMMENDS" to it from the base dist package.
When we publish packages to ni.com, we do not want to publish a
separate -lic package for each dist package because that can grow out
of control very quickly and start to pollute the feed.
So, disable the -lic package creation and install the license file
as part of the base dist package.

@ni/rtos

### DIFF
--- a/recipes-core/dist-nilrt/dist-nilrt-efi-ab-gateway.bb
+++ b/recipes-core/dist-nilrt/dist-nilrt-efi-ab-gateway.bb
@@ -2,9 +2,13 @@ SUMMARY = "IPK to repartition a safemode based installation to an A/B EFI implem
 LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
+LICENSE_CREATE_PACKAGE = "0"
+
 SRC_URI = "\
     file://nilrt-gateway-install \
+    file://MIT \
 "
+FILESEXTRAPATHS_prepend := "${COMMON_LICENSE_DIR}:"
 
 PV = "${DISTRO_VERSION}"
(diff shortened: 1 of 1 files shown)


jpautler (approved):

recipes-core/dist-nilrt/dist-nilrt-efi-ab-gateway.bb
```diff
@@ -32,4 +38,5 @@ python do_package_ipk_append() {
 FILES_${PN}_x64 = "\
     /usr/share/nilrt/restore-mode-image-${MACHINE}.iso \
     /usr/share/nilrt/nilrt-install \
+    /usr/share/licenses/${PN}/MIT \
```
amstewart, line 41: Sorry to add another spanner to the pile, but I'm concerned that this might not be the correct way to distribute the license. I'm worried that declaring it here (as a generic owned-file of the package) is incorrect, since that would imply that other packages "own" the same file.

Is this how the other `-lic` subpackages work? What happens if you `opkg remove` a `-lic` package; does opkg remove the `MIT` license from the system, or is it intelligent enough to only remove it once *all owners* are uninstalled?

shruthi-ravi: I don't think I fully understand the concern. Which other packages are you referring to when you say _owners_? The installation location of the license is namespaced. Only dist-nilrt-efi-ab-gateway owns that license file.. What am I missing?

amstewart: You're not missing anything. I was missing the `${PN}` in the path.

You're right; that makes sense. Resolved.

amstewart (approved):

PR #148, opened 2021-03-24
nilrt.conf: bump NILRT version

amstewart:
Bump the mainline NILRT version to 8.8.

----

The NIOE and os-common /8.7 components have already been switched to the 20.7 product branch. So it is appropriate to upgrade the NIRLT version on the mainline branch to 8.8.

@ni/rtos

### DIFF
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -2,9 +2,9 @@ require nilrt.inc
 
 DISTRO_NAME = "NI Linux Real-Time"
 
-DISTRO_VERSION = "8.7"
+DISTRO_VERSION = "8.8"
 
-NILRT_RELEASE_NAME = "2020.7"
+NILRT_RELEASE_NAME = "2021"
 
 DISTRO_FEATURES_append_x64 = "\
         x11 \


conf/distro/nilrt.conf
```diff
@@ -2,9 +2,9 @@ require nilrt.inc
 
 DISTRO_NAME = "NI Linux Real-Time"
 
-DISTRO_VERSION = "8.7"
+DISTRO_VERSION = "8.8"
```
shruthi-ravi, line 5: Just a sanity check, this branch (sumo) and os-common are always going to be bumping only minor versions, correct?

shruthi-ravi (approved):

conf/distro/nilrt.conf
amstewart: Unless someone convinces me otherwise; yes.

amstewart:
## Update 2
After talking with @jeminor : the NILRT 8.8 release will align with the LabVIEW "2021" release, so use 2021 as the `NILRT_RELEASE_NAME` instead.

conf/distro/nilrt.conf
```diff
@@ -2,9 +2,9 @@ require nilrt.inc
...
 
-NILRT_RELEASE_NAME = "2020.7"
+NILRT_RELEASE_NAME = "2021"
```
shruthi-ravi, line 7: This is probably going to be problematic. We use this variable to form feed URIs. Dunfell is already set to 2021..

amstewart: Yeah; I'm working on a fix for the dunfell feed version. I won't pull this until the dunfell feeds are moved.

shruthi-ravi (approved):
Assuming open issue with the Dunfell branch is resolved.

jeminor (approved):

PR #149, opened 2021-03-24
nilrt.conf: rename the release to "unstable"

amstewart:
Ever-changing planning requirements mean that it is difficult to assert
a version number on a purely development branch like 'dunfell'. Instead
of trying to guess the next labview release version, use a static string
like 'unstable' as the release name.

As a NILRT branch ref falls onto the commitment schedule, its release
name can be changed to a version number.

----

@shruthi-ravi I *think* this is all that is necessary to change the feed locations. I'm not seeing anywhere else in the new (dunfell) components that we encoded the `2021` or `21.X` version numbers. I'm doing a build now to confirm.

@ni/rtos This is in response to [PR 148](https://github.com/ni/meta-nilrt/pull/148). Any objections to the new feed name? It's [based on the debian feed names](https://www.debian.org/releases).

### DIFF
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -4,7 +4,7 @@ DISTRO_NAME = "NI Linux Real-Time"
 
 DISTRO_VERSION = "9.0"
 
-NILRT_RELEASE_NAME = "2021"
+NILRT_RELEASE_NAME = "unstable"
 
 DISTRO_FEATURES_append_x64 = "\
         x11 \
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -4,7 +4,7 @@ DISTRO_NAME = "NI Linux Real-Time"
 
 DISTRO_VERSION = "9.0"
 
-NILRT_RELEASE_NAME = "unstable"
+NILRT_FEED_NAME = "unstable"
 
 DISTRO_FEATURES_append_x64 = "\
         x11 \
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -141,7 +141,7 @@ PACKAGE_ENABLE_FILELIST ?= "True"
 
 # Root URI of all NILRT feeds
 NILRT_FEEDS_URI                    ?= "http://download.ni.com/ni-linux-rt/feeds"
-NILRT_FEEDS_URI_RELEASE            ?= "${NILRT_FEEDS_URI}/${NILRT_RELEASE_NAME}"
+NILRT_FEEDS_URI_RELEASE            ?= "${NILRT_FEEDS_URI}/${NILRT_FEED_NAME}"
 NILRT_MACHINE_FEED_URI_x64         ?= "${NILRT_FEEDS_URI_RELEASE}/x64"
 NILRT_MACHINE_FEED_URI_xilinx-zynq ?= "${NILRT_FEEDS_URI_RELEASE}/arm"
 
--- a/recipes-extended/lsb/lsb-release_%.bbappend
+++ b/recipes-extended/lsb/lsb-release_%.bbappend
@@ -2,5 +2,5 @@
 # nilrt-xfce to be the same as nilrt
 do_install_append() {
 	sed -i 's/DISTRIB_ID=nilrt-xfce/DISTRIB_ID=nilrt/' ${D}${sysconfdir}/lsb-release
-	echo "DISTRIB_CODENAME=${NILRT_RELEASE_NAME}" >> ${D}${sysconfdir}/lsb-release
+	echo "DISTRIB_CODENAME=${NILRT_FEED_NAME}" >> ${D}${sysconfdir}/lsb-release
 }


shruthi-ravi (approved):
What job updates the nickdanger [feed](http://nickdanger/feeds/2021/x64/)? I was unable to locate it. Need an update there too?

jpautler (commented):
Having a variable named NILRT_RELEASE_NAME set to "unstable" seems a little odd to me (maybe it shouldn't?). But it looks like that variable is only used to set the feed name (for which "unstable" seems fine) and to write a DISTRIB_CODENAME field to /etc/lsb-release (I guess that's okay?).

Maybe NILRT_RELEASE_NAME is not an ideal name for that variable?

amstewart:
@shruthi-ravi 
> What job updates the nickdanger [feed](http://nickdanger/feeds/2021/x64/)? I was unable to locate it. Need an update there too?

I'm thinking maybe [this script](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Frtos%2Frtoslegacyd%2FNIOE_Dunfell_x64_temp%2Ftools%2Fupdate_feed.sh) in the NIOE_Dunfell_x64_temp component? I don't remember ever reviewing this and I can't find exactly what pipeline runs it.

I'm thinking maybe now is just the time to kill those temp components.

amstewart:
> Maybe NILRT_RELEASE_NAME is not an ideal name for that variable?

What would you prefer? @jpautler

amstewart:
## Testing
Hot off the `build.py`...
```bash
root@NI-Unknown-5254e9a55943:/etc/opkg# cat base-feeds.conf 
src/gz uri-all-0 http://nickdanger.amer.corp.natinst.com/feeds/unstable/x64/extra/all
src/gz uri-x64-0 http://nickdanger.amer.corp.natinst.com/feeds/unstable/x64/extra/x64
src/gz uri-core2-64-0 http://nickdanger.amer.corp.natinst.com/feeds/unstable/x64/extra/core2-64
src/gz uri-all-1 http://nickdanger.amer.corp.natinst.com/feeds/unstable/x64/main/all
src/gz uri-x64-1 http://nickdanger.amer.corp.natinst.com/feeds/unstable/x64/main/x64
src/gz uri-core2-64-1 http://nickdanger.amer.corp.natinst.com/feeds/unstable/x64/main/core2-64
```

jpautler:
> > Maybe NILRT_RELEASE_NAME is not an ideal name for that variable?
> 
> What would you prefer? @jpautler

I was hoping you would have a great idea.  :)

The only thing that comes to mind is NILRT_FEED_NAME. If we did that, we might want to rename NILRT_FEEDS_URI_**RELEASE** to NILRT_FEEDS_URI_**NAME**, but I think that would be the end of the rename chain.

At that point it still feels a little odd to set DISTRIB_CODENAME to NILRT_FEED_NAME. I guess fundamentally the issue is that "unstable" makes sense to me as a feed name but not as a release or codename. Right now we are tying the two together via this variable. The only solution to that is probably to invent codenames, which seems like more trouble than it's worth...

amstewart:
The only other terminology I'm aware of is "suite" - which is [technically what debian calls it](https://wiki.debian.org/DebianRepository/Format#Suite). So we could call it something like `NILRT_FEED_SUITE` or `NILRT_SUITE_NAME`. But I can't imagine any other dev understanding what that means at first blush.

shruthi-ravi:
> 
> 
> > What job updates the nickdanger [feed](http://nickdanger/feeds/2021/x64/)? I was unable to locate it. Need an update there too?
> 
> I'm thinking maybe [this script](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Frtos%2Frtoslegacyd%2FNIOE_Dunfell_x64_temp%2Ftools%2Fupdate_feed.sh) in the NIOE_Dunfell_x64_temp component? I don't remember ever reviewing this and I can't find exactly what pipeline runs it.
> 
> I'm thinking maybe now is just the time to kill those temp components.

Some detecting later, I found this work [item](https://dev.azure.com/ni/DevCentral/_workitems/edit/1253687) that referred to the Jenkins job. All those jobs in heisenberg have disappeared? I'm fine with killing those components. But before submitting this change, a new job needs to be created that publishes an unstable branch on nickdanger. Otherwise, the image is going to be broken when running an `opkg update` during the testing stage in the pipeline.

jpautler:
> The only other terminology I'm aware of is "suite" - which is [technically what debian calls it](https://wiki.debian.org/DebianRepository/Format#Suite). So we could call it something like `NILRT_FEED_SUITE` or `NILRT_SUITE_NAME`. But I can't imagine any other dev understanding what that means at first blush.

Right- it's not even exactly a feed. It's like a meta-feed. Unstable makes sense in the persistent suite/meta-feed context, but that's not how we're using it. We're using it as a placeholder name until we have a fixed name, which is really like a codename. But we probably don't want to deal with inventing codenames just for this narrow use-case.

I don't feel strongly enough and don't have any great suggestions, so I don't want to hold up the PR over it. I just wanted to test the waters and see if anyone else had a better idea.

amstewart:
## Update 2
* Refactored `NILRT_RELEASE_NAME` to `NILRT_FEED_NAME`, to be more explicit about its purpose.

shruthi-ravi (approved):

jeminor (approved):

jpautler (commented):
Looks good to me.

PR #157, opened 2021-04-02
dist-nilrt-*.bb: Include license file in pkg

chaitu236:
Testing: Built locally, verified license file is present in package and -lic wasn't generated

@ni/rtos


billpittman (approved):

shruthi-ravi (approved):
Thanks for taking care of the dist-nilrt-efi-ab package too.

PR #159, opened 2021-04-07
Deprecate meta-orgconf from sumo

amstewart:
To deprecate meta-orgconf, some configuration settings need to be migrated to meta-nilrt - specifically, the feed URI variable. As long as I was touching the configuration files, I took the opportunity to reconcile the .conf file differences between the sumo and dunfell mainlines.

The mainlines have diverged enough that cherry-picking the conf file changes requires rebasing, so this PR is mostly a manual merge of conf file changes. It also deprecates the `nilrt-nxg` conf files, since they are no-longer used.

## Testing
* Manually diffed the files her versus their dunfell cousins and verified that any differences are intended.
* Verified that the bitbake environment init script pulls the correct config files, that they are parsed by bitbake, and that they correctly set key variables (DISTRO, MACHINE, FEED_URI, etc).

----

Ref NILRT.git [PR #53](https://github.com/ni/nilrt/pull/53)

@ni/rtos


shruthi-ravi (approved):

billpittman (approved):

amstewart:
## Update 2
* I oopsied and used the `nilrt-nxg` version of the `bblayers.conf.sample` as the basis for the new sumo version. The `${TOPLEVEL}` variable that it uses to path to the submodules doesn't make sense in our sumo build environment. This update uses the correct bblayers.conf instead.

amstewart:
## Testing
I've completed testing these and the nilrt.git changes with the sumo pipeline and am ready to submit.

amstewart:
Rebased and squashed fixups.

PR #161, opened 2021-04-09
migrate org-specific recipes from meta-orgconf

amstewart:
The meta-orgconf layer is being deprecated. Unfortunately, there are
still some recipes which rely on org-specific information to build -
particularly the recovery media.

Move the required classes and recipes from meta-orgconf into meta-nilrt,
so that we can deprecate the former.

This commit does not modify the contents of the moved files.

----

I was under the mistaken impression that these files were only used by NXG; they are necessary for os-common to build the recovery media.

## Testing
* os-common/8.8 now builds on my dev machine (with necessary fixups to the p4 components).

@ni/rtos


recipes-org/zynq-safemode-cfgs/zynq-safemode-cfgs.bb
```diff
@@ -0,0 +1,41 @@
+SUMMARY = "Packages containing older nilrd ARM .cfg files + a script to install them"
```
shruthi-ravi: How's this recipe used?

amstewart: I'm not actually sure. [Here's the history](https://github.com/ni/meta-orgconf/commits/nilrt/20.7/sumo/recipes-core/zynq-safemode-cfgs). I'm not confident enough with why it was added to assert that it should be removed.

shruthi-ravi: Hmmm. Does os-common build without it? I didn't see any codesearch references for this recipe.

> zynq-safemode-cfgs: add new recipe
This recipe creates safemode packages for restoring all nilrt supported
devices on the older OS version. The contents of each package is a *.cfg
file containing the necessary files for installation (itb, bitfiles,
preinst, postinst) and a standard script to install a cfg file.
These ipks will be downloaded on target from feeds during backwards
migration on arm devices based on their device code.

Is the package even added to our feeds? If the answers are no, and no, let's remove it?

amstewart:
## Update 2
* After searching around more thoroughly for anything which uses either the `zynq-safemode-cfg` IPKS - or their contents - I cannot find any users. Added a commit to this PR which deprecates the cfg recipes.
* Rebased on `nilrt/master/sumo`.

recipes-org/images/safemode-image.bb
```diff
@@ -0,0 +1,59 @@
...
+inherit build-services
+# optional env var with path to local safemode tar.gz
+export SAFEMODE_PAYLOAD_PATH
+EXPORTS_TO_FETCH = "$EXPORTS_TO_FETCH"
+
+do_fetch_prepend() {
+        mkdir -p ${BS_EXPORT_DATA}
+
+        if [ -z "$SAFEMODE_PAYLOAD_PATH" ]; then
+                EXPORTS_TO_FETCH=$(bs_get_latest_export "nilinux/os-common/export/8.0")"/standard_x64_safemode.tar.gz"
```
shruthi-ravi, line 22: Does the version in this path have to be updated every release?

shruthi-ravi (approved):

PR #164, opened 2021-04-13
Add support for an extra set of kernel packages and group configured with NO_HZ_FULL

gratian:
In certain types of real-time applications it is desirable to reserve CPU cores for real-time threads and avoid interference from the kernel as much as possible. This includes things like the kernel scheduler tick which can create jitter spikes.

Provide packages for a kernel configured with NO_HZ_FULL. This allows users with strict RT requirements to install a kernel that will adaptively try to shutdown the scheduler thick whenever possible. This comes at the expense of some overhead in user <-> kernel transitions (i.e. syscalls, exceptions, and interrupts).

This necessitaded some refactoring of existing recipes which assumed only two possible kernel configurations: 'current' and 'next'


gratian:
Attn: @ni/rtos 
FYI: @jhershbe

jhershbe:
Looks good to me.

amstewart (changes requested):
`linux-nilrt-extra` kind of sounds like it installs non-essential kernel components. What about `linux-nilrt-alternate` instead?

recipes-kernel/linux/linux-nilrt-next_git.bb
```diff
@@ -1,6 +1,6 @@
 DESCRIPTION = "NILRT linux kernel next development build"
 
-require linux-nilrt-next.inc
+require linux-nilrt-extra.inc
```
amstewart: linux-nilrt-extra kind of sounds like it installs non-essential kernel components. What about linux-nilrt-alternate instead?

amstewart (changes requested):
Disregard the above comment. Let's try this again, github.

recipes-kernel/linux/linux-nilrt.inc
```diff
@@ -1,10 +1,18 @@
...
 machine_srcrev="${SRCREV}"
 
+# Subfolder of the same name will be added to FILESEXTRAPATHS and also
+# used for nilrt-specific config fragment manipulation during build.
+# Provide a unique name for each recipe saved in the same source folder
+KBUILD_FRAGMENTS_LOCATION ?= "nilrt"
+
 # Set the version for traceability purposes
-LINUX_VERSION_EXTENSION = "$([ -z "${BUILDNAME}" ] || echo '-')${BUILDNAME}"
+LINUX_VERSION_EXTENSION = "$([ "${LINUX_KERNEL_TYPE}" != "standard" ] && echo '-${LINUX_KERNEL_TYPE}')$([ -z "${BUILDNAME}" ] || echo '-')${BUILDNAME}"
```
amstewart, line 15: Does the `LINUX_KERNEL_TYPE` get set to `standard` in the "current" kernel recipe? I don't see that happening anywhere.

recipes-kernel/packagegroups/packagegroup-ni-debug-kernel.bb
```diff
@@ -5,12 +5,11 @@ inherit packagegroup
 
 RDEPENDS_${PN} = "\
 	kernel-debug \
-	kernel-debug-base \
-	kernel-debug-vmlinux \
-	kernel-debug-image \
```
amstewart: Why are we dropping all these from the packagegroup? Are they deprecated?

recipes-kernel/linux/linux-nilrt-next_git.bb
gratian: Sounds good. I'll  change the name on the next patch rev.

recipes-kernel/linux/linux-nilrt.inc
gratian: meta-nilrt/recipes-kernel/linux/linux-nilrt.inc (which is used by all nilrt kernel recipes) includes openembedded-core/meta/recipes-kernel/linux/linux-yocto.inc which sets:

```
# The distro or local.conf should set this, but if nobody cares...
LINUX_KERNEL_TYPE ??= "standard"
```
We've been using this default for virtual/kernel package builds. There might be an argument to be made here about setting the (weak defined) default in distro or local.conf as the comment in that .inc file implies but we've been operating using the linux-yocto.inc default for a while so I didn't mess with it.

amstewart: OK. That's sufficient, I think.

recipes-kernel/packagegroups/packagegroup-ni-debug-kernel.bb
gratian: The packagegroup is over-specifying the dependencies. The dependency chain for _kernel-debug_ already includes the _kernel-debug-base_ dependency which in turn depends on the package providing _kernel-debug-image_. Here's the dependency chain reconstructed from .ipks I have built to test this:

```
  kernel-debug_4.14+git0+db94e1d70a-r0_x64.ipk
  Depends: kernel-debug-4.14.146-rt67-debug-dev-nilrt

	kernel-debug-4.14.146-rt67-debug-dev-nilrt_4.14+git0+db94e1d70a-r0_x64.ipk
	Depends: kernel-debug-image-4.14.146-rt67-debug-dev-nilrt
	Provides: kernel-debug-4.14.146-rt67-debug-dev-nilrt, kernel-debug-base

		kernel-debug-image-4.14.146-rt67-debug-dev-nilrt_4.14+git0+db94e1d70a-r0_x64.ipk
		Depends: kernel-debug-image-bzimage-4.14.146-rt67-debug-dev-nilrt
		Provides: kernel-debug-image

			kernel-debug-image-bzimage-4.14.146-rt67-debug-dev-nilrt_4.14+git0+db94e1d70a-r0_x64.ipk
			Provides: kernel-debug-image-bzimage

```
The removal of _kernel-debug-vmlinux_ is accidental and caused by the fact I was reusing the nohz/next package groups (which don't need/want the vmlinux binary). For the debug kernel the vmlinux package is needed bc. it contains the un-stripped kernel image with debug symbols. Thanks for catching this; will fix in V2.

gratian:
Changes since V1:

  * rebase on latest nilrt/master/sumo

  * rename linux-nilrt-extra.inc -> linux-nilrt-alternate.inc and update corresponding recipes

  * re-add kernel-debug-vmlinux to packagegroup-ni-debug-kernel.bb

amstewart (approved):

amstewart:
@chaitu236 Any objections?

chaitu236 (approved):

chaitu236:
> @chaitu236 Any objections?

No looks good to me.

PR #165, opened 2021-04-14
curl: Reverts to stop calling MoveFileEx

rtollert:
Upstream made some file I/O atomicity fixes recently, which broke Phar Lap,
because they require MoveFileEx() which does not exist there. (The fixes appear
to be mostly cache-related and are probably not going to get hit the way we're
using curl right now.)

Natinst-AZDO-ID: 1385887

Signed-off-by: Richard Tollerton <rich.tollerton@ni.com>


amstewart (approved):
No objections from me.

billpittman (approved):
No objections here.

PR #166, opened 2021-04-16
niwatchdogpet: move watchdogpet later to fix fw_printenv read error

NIwkoe:
While developing safemode for artemis, we found out that fw_printenw call in
niwatchdogpet cannot function properly because the config lives in
/boot/uboot/uboot.env, which requires the /boot partition to be mounted first.
The mounting of /boot for artemis depends on /dev/mmcblk0p1 node to be ready,
which requires udev to be initialized first. Hence, by moving niwatchdogpet
from S04 to S05, we can resolve this issue.

This commit changes niwatchdogpet startup from S04 to S05.

@ni/rtos 


gratian (approved):

amstewart (approved):
Thanks for putting this in. When you have some time, could you also cherry-pick this onto the `nilrt/master/dunfell` branch and submit a PR for it.

PR #167, opened 2021-04-16
nilrt.inc: re-add USERADD-staticids

amstewart:
While rearchitecting the layer and distro configuration files for the
meta-orgconf deprecation effort [1], I accidently forgot to move the
USERADD block of local.conf into the distro.conf. As a result, the
uid/gid tables generated by OE - though internally valid - were not
consistent with what is expected by externally-built NI software.

Re-add the USERADD block of configuration to return to consistent
uid/gids.

[1] https://github.com/ni/meta-nilrt/pull/159

----

I have not modified the content of the configuration block from what we were using previously, or what we are currently using the the dunfell mainline.

## Testing

Rebuilding a minimal rootfs now, to verify that the passwd/group files look correct.

@ni/rtos

### DIFF
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -39,6 +39,30 @@ RM_WORK_EXCLUDE += " \
 	safemode-restore-image \
 "
 
+# Various packages dynamically add users and groups to the system at package
+# install time.  For programs that do not care what the uid/gid is of the
+# resulting users/groups, the order of the install will determine the final
+# uid/gid.  This can lead to non-deterministic uid/gid values from one build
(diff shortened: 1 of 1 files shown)


shruthi-ravi (approved):

gratian (approved):

jeminor (approved):
I don't know what all of the implications of this are: deferring to Shruthi and your testing to insure this fixes the issue.

amstewart:
## Testing
* Rebuilt `niconsole-image` and confirmed that the ni group values are set to the correct gids in `/etc/group` and that they match the (correct) values from the 8.8.0d1 export.

PR #170, opened 2021-04-19
packagegroups; disable building the debug-kernel pkggrp

amstewart:
The kernel-debug recipes have package names which are just long enough
to blow the 260 character limit on the nibuild export tooling.
Therefore, sumo builds cannot be exported.

Temporarily disable building the debug-kernel packagegroup, until we can
figure out a long-term solution.

## Testing
* I will let heisenberg test this.

@ni/rtos


gratian (approved):
I'm not convinced this is the only kernel flavor that will run into trouble and we should probably fix the export paths instead but in the interest of moving things forward. Ship it.

PR #174, opened 2021-05-06
ni-grpc-device: upgrade to 1.1

amstewart:
ni-grpc-device: upgrade 1.0.1 -> 1.1
    
    The grpc-device internal release has changed tags upstream. Rename the
    recipe into the 1.1.X range to follow the release tag.
    
    On the understanding that the repo will follow semantic versioning, use
    major.minor as the recipe filename, and add the patch rev within the
    recipe.

---

    ni-grpc-device: add correct legal files
    
    grpc-device upstream has added LICENSE and third party notice files to
    its release artifacts. Consume and package these licenses directly for
    legal compliance.
    
----

## Testing
* Rebuilt recipe on my dev machine, verified that the LICENSE and ThirdPartyNotice.txt files are in the `-lic` subpackage.

@ni/rtos


shruthi-ravi:
How often will the tags be changed upstream? Will they coordinate with you when they do?

billpittman (approved):

amstewart:
> How often will the tags be changed upstream? Will they coordinate with you when they do?

I spoke with them about this yesterday. For the next couple cycles, I expect regular development will happen on the grpc-device repo. They have agreed to follow semantic versioning, and to pushing a new GH Release object whenever some significant changes go in. They'll be tagged like `v1.2.3-rc4` for pre-release objects, and the release tags will be the same - minus the `-rc#`.

Once they conform to that standard, I can drop the `${RELEASE_TAG}` indirection in this version of the recipe and we can treat this as a more-normal recipe.

We have also levelset on the fact that the release object archives can't change contents once we have consumed them into a recipe (they weren't aware that we are checksumming the sources).

I've also been added as an admin on that repo and should have some reasonable insight into when they push releases.

shruthi-ravi (approved):

PR #175, opened 2021-05-07
ni-grpc-device: Add ptest

billpittman:
Add ptest smoke test for testing the server is installed and
communicating.


recipes-ni/ni-grpc-device/ni-grpc-device_1.1.bb
```diff
@@ -8,7 +8,13 @@ LIC_FILES_CHKSUM = "\
 	file://${PN}-server/ThirdPartyNotices.txt;md5=6def9ca42f278e76d89243699fae2b67 \
 "
 
+inherit ptest
+
 SRC_URI = "\
+	file://enumerate-device.py \
+	file://run-ptest \
+	file://session_pb2_grpc.py \
+	file://session_pb2.py \
```
amstewart, line 17: Are these files from upstream (without modification)? If so, can we add them as a remote source and package them.

amstewart, line 17: Move any `files/` content under the remote source entries. Bitbake will assume that the first source URI represents the *primary* source for the project and use it for things like package versions. I don't think it strictly matters in the case of our recipe, but it's standard practice.

recipes-ni/ni-grpc-device/files/run-ptest
```diff
@@ -0,0 +1,30 @@
...
+    echo "ERROR on line $lineno: $msg" 1>&2
+    echo "FAIL: $TEST_NAME"
+    exit 1
+}
+
+trap 'error_exit_handler ${LINENO} "$BASH_COMMAND"' ERR
+
+# Run the python code and expect the server to respond with a known error message
+# about the NI System Configuration API not being installed.  The API is not
+# required, the fact that the server responds is enough for the test to pass.
```
amstewart, line 25: Don't we expect a device stanza for the local device, per [this func](https://github.com/ni/grpc-device/blob/main/examples/session/enumerate-device.py#L28)?

recipes-ni/ni-grpc-device/ni-grpc-device_1.1.bb
```diff
@@ -8,7 +8,13 @@ LIC_FILES_CHKSUM = "\
 	file://${PN}-server/ThirdPartyNotices.txt;md5=6def9ca42f278e76d89243699fae2b67 \
 "
 
+inherit ptest
```
amstewart: STYLE: better to put this inherit line below the class-generic variables (see [this guide](https://www.openembedded.org/wiki/Styleguide#Ordering_and_grouping)). You could also group it with the `RDEPENDS_${PN}-ptest` and `do_install_ptest_append ()` lines.

amstewart (changes requested):

recipes-ni/ni-grpc-device/ni-grpc-device_1.1.bb
billpittman: No these are generated files from python3-grpcio-tools being run with session.proto as input.  I didn't see any reason to force a python compile for this simple test, so I've just included the output from the compile.

billpittman: As you wish.

recipes-ni/ni-grpc-device/files/run-ptest
billpittman: No we don't, that function call is the second one in the file, and I suspect that the first line in the try/catch fails and sends us through the catch code.  Either way, live running of the test outputs only a single line of text.

recipes-ni/ni-grpc-device/ni-grpc-device_1.1.bb
billpittman: As you wish.

recipes-ni/ni-grpc-device/ni-grpc-device_1.1.bb
billpittman: By the way, the test itself enumerate-devices.py is the same as the example, but its been "downgraded" to not use F-Strings since python 3.5.5 doesn't support them.  So the test file itself is semi-custom.

amstewart: OK. I'm fine pulling this for now then. Can you upstream the non-f-string version of `enumerate-devices.py` though?

amstewart (approved):

PR #177, opened 2021-05-11
run-ptest: Launch daemon as part of the test

billpittman:
Because the daemon is not started by an init script, the
test needs to start and daemon, and to be a good citizen, the
test also stops the daemon.


billpittman:
@amstewart

amstewart (approved):

PR #179, opened 2021-05-13
[Dunfell] add ni-grpc-device recipe

amstewart:
The [ni/grpc-device](https://github.com/ni/grpc-device) project is a server daemon which enables NI drivers to communicate with deployed LinuxRT systems using grpc. NILRT needs to provide a driver-agnostic package, which distributes the server-side grpc daemon, that NI drivers can depend upon for their specific implementations.

### Implementation
This patchset adds a new recipe: `recipes-ni/ni-grpc-device/ni-grpc-device_git.bb`, which compiles and packages the latest `main/` ref of the grpc-device repo. The package is architecturally similar to the recipe which has been accepted to the sumo mainline in PR #173, but the implementation deviates significantly: this version of the recipe compiles `ni_grpc_device_server` from source, using OE-native toolchain packages.

This version of the recipe also generates the `session_pb*.py` files (needed for ptesting) during the OE recipe build, rather than statically including them in the OE recipe source. The `enumerate-device.py` file is copied from the upstream source, and is modified to not use f-strings. Once it is upstreamed, we can drop this file as well.

### Patches
This patchset adds 5 patches to grpc-device, all of which apply to the `CMakeLists.txt` file from upstream. The first four patches are being upstreamed in [this grpc-device PR](https://github.com/ni/grpc-device/pull/195). Once they are upstreamed, I will remove them from the recipe. The final patch (`0005-CMakeLists.txt-remove-local-protobuf-includes.patch`) requires some rearchitecture to be upstreamable. I don't think its worthwhile at this time to bother doing so.

# Testing

The `ni-grpc-device` recipe builds on my local machine. I installed it (and its dependencies) to a recently built Dunfell QEMU vm and verified both manually and using the ptest, that the daemon starts correctly. Upstream's GH workflows verify that the CMakeList changes do not break compilation in non-OE environments.

@ni/rtos @SparkingSpork


amstewart:
## PATCH V2
* I remembered that we ship python3.8 with the dunfell pipeline. So it isn't necessary that we package a version of the `enumerate-devices.py` script that doesn't use f-strings. I've modified the recipe to use the version that is already upstreamed with grpc-device.

### Testing
Rebuilt the recipe and installed the new `-ptest` subpackage to a recent Dunfell QEMU image. Verified that the ptests still work.

amstewart:
## PATCH V3
* grpc-device [PR 195](https://github.com/ni/grpc-device/pull/195) has been accepted into the mainline. Removed patches 0001-0004, which are now in the mainline ref.

### Testing
Verified that the recipe still builds locally.

recipes-ni/ni-grpc-device/ni-grpc-device_git.bb
```diff
@@ -0,0 +1,90 @@
...
+	install -d ${D}${datadir}/${BPN}
+	install --mode=0644 ${B}/server_config.json ${D}${datadir}/${BPN}/server_config.json.example
+
+	# package .proto files for use by developers
+	install -d ${D}${includedir}/${BPN}
+	install --mode=0644 ${S}/generated/**/*.proto          ${D}${includedir}/${BPN}
+	install --mode=0644 ${S}/source/protobuf/session.proto ${D}${includedir}/${BPN}
+}
+
+RDEPENDS_${PN} += "\
```
billpittman, line 90: I think this RDEPENDS should be moved up in the recipe.  It looks odd to have an RDEPENDS so late.  How about moving it up near the ptest RDEPENDS and then either changing the format of this RDEPENDS to be one line, or change the ptest one to be multi-line ?

billpittman (changes requested):

billpittman:
Sorry wrong request.  Re-opened now.

recipes-ni/ni-grpc-device/ni-grpc-device_git.bb
amstewart: I have it at the end based on the [yocto recipe styleguide](https://www.openembedded.org/wiki/Styleguide#Ordering_and_grouping), which recommends having all the `RVARIABLE`s at the end of the recipe.

I'd guess that's to ensure that the owning recipe has the final say wrt. package metadata (ie. it can overwrite `RDEPENDS` from inheriting classes). Since I'm only appending to RDEPENDS here, it might not matter.

If you feel strongly about it; I can move it.

amstewart: >and then either changing the format of this RDEPENDS to be one line, or change the ptest one to be multi-line ?

I will change the `RDEPENDS_ptest` to be multiline.

amstewart:
## PATCH V4
* Broke `RDEPENDS_${PN}-ptest` into multiple lines.
* Other trivial whitespace and ordering.

billpittman (approved):

recipes-ni/ni-grpc-device/ni-grpc-device_git.bb
billpittman: Nah its fine, just wanted to make sure it was thought through.

PR #181, opened 2021-05-20
Shorten linux-kernel package names

amstewart:
Currently, the `BUILDNAME` variable string is appended to each kernel package and subpackage name, including the generated kernel-module packages. As a result, the package name (and therefore filepath) get very long. eg.

```
kernel-debug-module-nf-conntrack-broadcast-4.14.146-rt67-debug-cg-8.8.0d13-xilinx-zynq-27_4.14+git38+a0899816ee-r0.13_xilinx-zynq.ipk.

-cg-8.8.0d13-xilinx-zynq-27  # just the BUILDNAME
```

We add the `BUILDNAME` for traceability, but haven't actually used it for tracing very frequently (ever?). Further, these long filenames are starting to exceed the 260 character limit imposed by nibuild tooling. We previously had to disable the kernel-debug recipe entirely, because of that limit.

This patchset removes the `BUILDNAME` part of the package name - which should have no functional impact on the packages, but saves us many characters in the final filepaths. It then reverts the commit which disabled the debug kernel recipes (ie. enables them again.)

New packages look like:
```
kernel-debug-module-nf-conntrack-broadcast-4.14.146-rt67-debug_4.14+git0+a0899816ee-r0_xilinx-zynq.ipk
```

My simple calculations show that the longest IPK names have been reduced by ~27 characters.

## Testing

The kernel recipes build on my dev machine. I will let the build systems verify that it meets total path length requirements. But I will be very surprised if this doesn't resolve the issue (we should be ~25 characters under the limit, now).

## Future Work

The `LINUX_VERSION_EXTENSION` will be cherry-picked into the dunfell branch as well, for consistency and also because the same path length limit exists in the ni-central tooling.

There are better long-term solutions to our path-length woes (namely, archiving the built feeds in NIOE/rtos_nifeeds before export). But those solutions require rearchitecture of the components and will be easier to implement once we have migrated NIOE into AZDO.

@ni/rtos

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-desirable.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-desirable.bb
@@ -7,6 +7,7 @@ inherit packagegroup
 # essential packagegroups
 RDEPENDS_${PN} += "\
 	packagegroup-core-tools-debug \
+	packagegroup-ni-debug-kernel \
 	packagegroup-ni-ptest \
 	packagegroup-ni-selinux \
 "
--- a/recipes-core/packagegroups/packagegroup-ni-extra.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-extra.bb
@@ -71,6 +71,7 @@ RDEPENDS_${PN} = "\
 
 # meta-nilrt
 RDEPENDS_${PN} += "\
+	packagegroup-ni-debug-kernel \
 	packagegroup-ni-selinux \
 "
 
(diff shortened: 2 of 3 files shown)


amstewart:
Actually, hold off on this review. While cherry-picking into dunfell, I discovered that past-Alex [already removed](https://github.com/ni/meta-nilrt/commit/3154f2f82c4a4d9f300fdac407b0b8278fd3ec89) all the dynamic information from the dunfell kernel packages in PR #124 (for a different reason).

So I'm going to abandon this PR and pull the dunfell implementation into sumo - since it accomplishes the same thing.

amstewart:
Disregard previous; I think the sumo implementation is actually what we want. Dunfell's `-ni` extension is probably what needs to be reconciled.

gratian (approved):

PR #183, opened 2021-05-21
RFC: add PACKAGE_EXCLUDE to make sure we keep python out of base images

bstreiff:
This patch adds a `PACKAGE_EXCLUDE` to `nilrt-image-common.inc` for `python-core` and `python3-core` so that they aren't unknowingly added to our base images again, due to the size of the python runtimes.

### DIFF
--- a/recipes-core/images/nilrt-image-common.inc
+++ b/recipes-core/images/nilrt-image-common.inc
@@ -11,6 +11,9 @@ IMAGE_PREPROCESS_COMMAND += "rootfs_update_timestamp;"
 # Do not install license subpackages if they are only recommended.
 BAD_RECOMMENDATIONS_pn-${PN} += "*-lic"
 
+# Do not allow python to be installed into base images due to size
+PACKAGE_EXCLUDE += "python-core python3-core"
+
 # NILRT images expect the default kernel is an actual file
 # (e.g. bzImage) and not a symbolic link; fix it up here.
 ROOTFS_POSTPROCESS_COMMAND += "move_kernel;"


classes/qa-denydeps.bbclass
```diff
@@ -26,21 +26,24 @@ python package_qa_check_denydeps() {
...
+            return True
 
         rdepends = bb.utils.explode_deps(pkgdata['RDEPENDS_' + package] or "")
 
         for rdepend in rdepends:
-            check_deps(rdepend, denylist, visited, path + [rdepend])
+            if not check_deps(rdepend, denylist, visited, path + [rdepend]):
+                return False
+
+        return True
```
amstewart: I think the second commit has some fixup diffs for `qa-denydeps.bbclass` that should have been squashed into the previous commit.

amstewart (changes requested):

amstewart:
>I had also hoped that this could be potentially reused in other locations, but upon further reflection I don't know of where we would have put it to handle our other current snake infestation (dkms -> kernel-devsrc -> python3) since in that case dkms isn't being included via packagegroup.

Is it viable to get this QA check working for `IMAGE_INSTALL` entries in [image recipes](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-core/images/nilrt-dkms-image.bb#L6)? If so, that might be the ideal place to put it.

classes/qa-denydeps.bbclass
bstreiff: ugh, oops. fixed now.

bstreiff:
It looks like the `IMAGE_INSTALL` steps in the image recipes don't show up as `RDEPENDS` so `qa-denydeps` doesn't work. Digging into why that didn't work, I did find an existing `PACKAGE_EXCLUDE` feature that can be used:
```diff
diff --git a/recipes-core/images/nilrt-dkms-image.bb b/recipes-core/images/nilrt-dkms-image.bb
index a2bb7850..7b2b2b78 100644
--- a/recipes-core/images/nilrt-dkms-image.bb
+++ b/recipes-core/images/nilrt-dkms-image.bb
@@ -12,3 +12,6 @@ require nilrt-initramfs.inc
 require include/licenses.inc
 
 IMAGE_FSTYPES += "squashfs"
+
+# no sneks in image pls
+PACKAGE_EXCLUDE = "python-core python3-core"
```

```sh
(bb) bstreiff@starthinker:~/git/nilrt-sumo/build$ bitbake nilrt-dkms-image
[...]
NOTE: Executing RunQueue Tasks
ERROR: nilrt-dkms-image-1.0-r0 do_rootfs: Unable to install packages. Command '/home/bstreiff/git/nilrt-sumo/build/tmp-glibc/work/x64-nilrt-linux/nilrt-dkms-image/1.0-r0/recipe-sysroot-native/usr/bin/opkg --volatile-cache -f /home/bstreiff/git/nilrt-sumo/build/tmp-glibc/work/x64-nilrt-linux/nilrt-dkms-image/1.0-r0/opkg.conf -t /home/bstreiff/git/nilrt-sumo/build/tmp-glibc/work/x64-nilrt-linux/nilrt-dkms-image/1.0-r0/temp/ipktemp/ -o /home/bstreiff/git/nilrt-sumo/build/tmp-glibc/work/x64-nilrt-linux/nilrt-dkms-image/1.0-r0/rootfs  --force_postinstall --prefer-arch-to-version  --add-exclude python-core --add-exclude python3-core install dkms opkg packagegroup-core-x11 packagegroup-ni-base packagegroup-ni-runmode packagegroup-ni-tzdata packagegroup-ni-wifi packagegroup-ni-xfce run-postinsts' returned 2:
Collected errors:
 * Solver encountered 2 problem(s):
 * Problem 1/2:
 *   - nothing provides python3-core needed by python3-modules-3.5.5-r1.0.119.core2-64
 * 
 * Solution 1:
 *   - do not ask to install a package providing dkms

 * Problem 2/2:
 *   - package packagegroup-ni-runmode-1.0-r0.427.x64 requires opkg-utils, but none of the providers can be installed
 * 
 * Solution 1:
 *   - do not ask to install a package providing packagegroup-ni-runmode
```

The error reporting is a little less useful than I'd like (for instance, for problem 2, it doesn't say _why_ it cannot install opkg-utils, just that it can't), but maybe this is the route we should be taking instead of trying to handle things at the packagegroup level?

amstewart:
Oof. Yeah; that error message isn't too helpful. I definitely think that doing the exclude in the image recipe is the correct architecture, though.

I've thought for a bit about where we could add a qa check to provide more informative debugging information. But honestly, with us installing ni content from feeds now, there are an entire family of reasons why the IMAGE_INSTALL might fail in that same place. We're going to need to be comfortable debugging based on that error message one way or another. So adding an additional QA check somewhere is probably not worth it.

I think we should take the `PACKAGE_EXCLUDE` approach - assuming that it meets all of our requirements.

bstreiff:
Poking through this some more: I think the ideal situation here is that `opkg` should output a "solution" to "nothing provides python3-core" that is "do not pass `--add-exclude python3-core` to opkg"; that way when we build an image and a package is excluded via `PACKAGE_EXCLUDE` there is a hint that the request is unsolvable because of such an exclusion. Unfortunately it looks like today `opkg` just avoids passing in excluded packages to `libsolv`, so `libsolv` doesn't know to suggest it as a solution. `libsolv` did [learn how to have "blacklisted packages" in 2019](https://github.com/openSUSE/libsolv/commit/4120051be87d2cccc2146a30c5128edd5c78d76d), so maybe `opkg` should be updated to use that.

But that's an opkg feature request, not a meta-nilrt bug.

I think in the meantime setting `PACKAGE_EXCLUDE` in our images is the right thing to be doing as a QA check for "avoid python in the base images". Which places should that go? All three of `nilrt-image-common`, `minimal-nilrt-bundle-image`, and `nilrt-dkms-bundle-image`?

amstewart:
> But that's an opkg feature request, not a meta-nilrt bug.

Agreed.

> I think in the meantime setting `PACKAGE_EXCLUDE` in our images is the right thing to be doing as a QA check for "avoid python in the base images". Which places should that go? All three of `nilrt-image-common`, `minimal-nilrt-bundle-image`, and `nilrt-dkms-bundle-image`?

Yeah; I think `nilrt-image-common.inc` would be the correct place for sumo.

I'm not sure if we can say that the bundle images shouldn't have python. I think we need to have a separate conversation with @jpautler about whether it makes sense to apply this same restriction to the dunfell images.

bstreiff:
Pushed new implementation that is just using `PACKAGE_EXCLUDE`, since prior testing showed that this does the thing that we want.

With this change I have tested that the following image recipes build successfully. I didn't check for functionality.
- `niconsole-image` (arm)
- `niconsole-dkms-image` (arm)
- `nixfce-image` (x64)
- `nilrt-dkms-image` (x64)
- `nilrt-dkms-bundle-image` (x64)

`restore-mode-image` and `minimal-nilrt-bundle-image` fail in `do_image`, but they also fail without this change; since these have been deprecated it looks like they may have bitrotted.

amstewart (approved):
Looks good to me.

amstewart:
@jeminor I want to give you an opportunity to object to pulling this into the 21.0 release. I don't expect it will break anything, but there's a chance it could introduce some instability. If we want to avoid that, I can cut a 21.0 release branch without it.

amstewart:
Received an ACK from @jeminor .

PR #187, opened 2021-06-07
kernel-tests: Add ptest for dynamic tick configured kernels

gratian:
Add a ptest for testing NO_HZ_FULL configured kernels. The test sets
up the knobs required for maximum performance: core isolation, cpu
sets, cpu masks for kernel workqueue threads etc. and then uses a
clock_gettime() based polling loop to calculate the maximum latency
and a set of percentile values. The test success/failure is determined
based on several configuration options: maximum latency, 99.999% and
99.9999% thresholds.

The ptest should be run with an appropriate (large) timeout value set:
  e.g. ptest-runner -t 345600

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-desirable.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-desirable.bb
@@ -35,6 +35,7 @@ RDEPENDS_${PN} += "\
 	iperf2 \
 	iperf3 \
 	kernel-performance-tests \
+	kernel-test-nohz \
 	ldd \
 	ltrace \
 	mysql-python \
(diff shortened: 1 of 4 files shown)


gratian:
Attn: @ni/rtos

amstewart (changes requested):
Since the `ptest-runner` would have to be run with such a large timeout to satisfy this test, is it even appropriate to link it into the ptest-autodiscovery functionality in ptest-runner?

Maybe, instead of making a `-ptest`subpackage and installing a `run-ptest` hook, we should just make this a standalone test package - with a `run-test` script that the user is expected to call intentionally. What do you think?

gratian:
Personally I've found it convenient to use ptest-runner. It avoids having to hard-code paths to tests etc. The kernel performance tests have the same problem and as long as the person running the test or the ATS invokes ptest-runner with '-t' option I don't see a problem with it. I've been using the performance tests in this manner to validate new kernels and it works pretty well.

I do agree that making this test (and the performance tests) part of packagegroup-ni-ptest or packagegroup-ni-ptest-smoke which don't have such a long run-time would be a problem (hence why I have added them to the desirable group instead). I expect in the long run we'll encounter other long duration tests and we might want to create a separate package group for them (or just have them as loose tests) but I don't think that's a good reason not to use the ptest framework. The '-t' option exists for a reason.

amstewart (approved):

PR #188, opened 2021-06-09
fixup! kernel-tests: Add ptest for dynamic tick configured kernels

gratian:


gratian:
FYI: @ni/rtos

amstewart (approved):

PR #190, opened 2021-06-09
kernel-tests: Fix kernel-test-nohz x64 only dependency

gratian:
kernel-test-nohz ptest is expected to run on NO_HZ_FULL configured (5.10) kernels currently only available for x64 NILRT machines. Fix run-time dependencies to:

- only add it to the packagegroup-ni-desirable for x64
- only declare a run-time dependency on packagegroup-ni-nohz-kernel for x64. The test can still be built via bitbake for Zynq if desired but it is not  the expected use case.

Tests:

- bitbake kernel-test-nohz succeeds on both x64 and Zynq targets after these changes.


gratian:
FYI: @ni/rtos

amstewart (approved):
I'll also cherry-pick this into the new `nilrt/21.0/sumo` branch.

PR #192, opened 2021-06-10
kernel-devsrc: add elfutils-dev as dependency

bstreiff:
The dkms module rebuild machinery expects to have objtool if elfutils-dev
is on the system at the time of rebuild, but if elfutils-dev was not
present at the time that kernel-devsrc was installed, then objtool is not
generated. Resolve this by making elfutils-dev a dependency; thus it
should always be present if kernel-devsrc is.

Natinst-AzDO-ID: 1470108

### DIFF
--- a/recipes-kernel/linux/kernel-devsrc.bbappend
+++ b/recipes-kernel/linux/kernel-devsrc.bbappend
@@ -3,3 +3,9 @@ pkg_postinst_ontarget_kernel-devsrc () {
    make prepare
    make modules_prepare
 }
+
+# "make prepare" skips building objtool if elfutils-dev is not installed at
+# postinst time; however if you install elfutils-dev later, then when dkms
+# tries to build modules it expects objtool to be available, but it's not.
+# Depending on elfutils-dev forces it to be there during postinst.
+RDEPENDS_${PN}_append += "${@bb.utils.contains('ARCH', 'x86', 'elfutils-dev', '', d)}"


recipes-kernel/linux/kernel-devsrc.bbappend
```diff
@@ -3,3 +3,9 @@ pkg_postinst_ontarget_kernel-devsrc () {
    make prepare
    make modules_prepare
 }
+
+# "make prepare" skips building objtool if elfutils-dev is not installed as
```
amstewart: s/as/at/

amstewart (changes requested):
Looks good to me otherwise.

amstewart:
This needs to get into the 21.0 release; right?

bstreiff:
Since @jeminor asked offline about space usage: installing `elfutils-dev` also installs `libtool` and `libtool-dev` as things above-and-beyond what is normally installed on a VM image with the BSI installed plus `ni-crio` (which is what MAX wants to recommend that I put on this VM). Data from `opkg info`:

```
libtool: Installed-Size: 2104586
libtool-dev: Installed-Size: 17909
elfutils-dev: Installed-Size: 210577
```

This is space that's used only after we have dkms-related packages installed; we don't pull `kernel-devsrc` into the base image itself.

bstreiff:
Updated commit to fix typo in comment.

amstewart (approved):

PR #195, opened 2021-06-17
add pstore-save, replacing nieficleanvars fixup script

bstreiff:
As a better solution to dealing with the buildup of EFI variables created by `efi-pstore`, add a new script, `pstore-save`, to actually save off the dmesg data from prior boots. This works similarly (although not identically) to how [`systemd-pstore`](https://www.freedesktop.org/software/systemd/man/systemd-pstore.service.html) would if configured to save to local files and not the journal.

It may be possible to leverage this for ARM (using the `ramoops` backend for pstore) but our targets are not presently configured to use `ramoops`, so this is only being enabled for x64.

This obsoletes the `nieficleanvars` script.


recipes-ni/pstore-save/files/pstore-save
```diff
@@ -0,0 +1,139 @@
...
+            fi
+        fi
+    fi
+}
+
+function process_dmesg_files () {
+    declare -A LOGGED_TIMESTAMPS
+
+    # Any files that start with "dmesg" and end in ".enc.z" are bad;
+    # move them to the archive and don't try to process them otherwise.
```
amstewart: Could you clarify this comment? What does "are bad" mean, in this context?

amstewart (changes requested):
All looks good; apart from the minor clarification request above.

Also, could you add the ptest package to the [smoke packagegroup](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-core/packagegroups/packagegroup-ni-ptest-smoke.bb). It doesn't matter at this stage, but I'd like to have it enumerated there, so that we don't forget to enable it when we build out the ptest infrastructure.

recipes-ni/pstore-save/files/pstore-save
bstreiff: I explain it about [as well as systemd-pstore does](https://github.com/systemd/systemd/blob/df8fd03a7dcf2f06cd9e2ee3eed60c3a83b43ccc/src/pstore/pstore.c#L229). :) I can update the comment to clarify though.

In practice, it seems to mean "the underlying pstore backend says that these are compressed data elements, but an error occurred when the backend attempted to decompress them in order to present them as entries in /sys/fs/pstore (where from the point-of-view of the reader they are uncompressed text files)".

In the case of the "nicleanefivars is deleting the dump variables out from behind the back of pstore" case, the way that this would manifest is that the /sys/fs/pstore directory would be full of a bunch of 1024-bytes-of-zeroes files named `dmesg-efi-157349851401001.enc.z` (and also get some dmesg spew of `pstore: decompression failed: -5`)

recipes-ni/pstore-save/files/initscript
```diff
@@ -0,0 +1,46 @@
...
+
+# Read configuration variable file if it is present
+[ -r /etc/default/$NAME ] && . /etc/default/$NAME
+
+# no-op if ENABLE is 0
+if [ $ENABLE -eq 0 ]; then
+    exit 0
+fi
+
+test -x "$SCRIPT" || exit 0
```
shruthi-ravi, line 30: Shouldn't this be an error on exit, or atleast a debug message?

shruthi-ravi (approved):

recipes-ni/pstore-save/files/initscript
bstreiff: "exit 0 if the program the initscript is for isn't executable" is conventional. See [here](https://github.com/ni/meta-nilrt/blob/598ee2a3e03240f8876f4d88845253bcd50f35d6/recipes-ni/niwatchdogpet/files/niwatchdogpet.sh#L12) and [here](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-core/busybox/files/busybox-acpid#L6).

amstewart (approved):

amstewart:
Approved. Let me know when you're ready for merge, @bstreiff .

PR #196, opened 2021-06-17
initrdscripts: skip provisioning MMC block devices by default

amstewart:
Some supported NILRT hardware is equipped with onboard eMMC block
devices or a removable SD card slot. In most cases, it is not sane to
provision to an MMC block device.

Modify the ni_provisioning.common script to, by default, skip any MMC
block devices when considering candidates for provisioning.

To accomidate devices which do legitimately use MMC blocks for primary
storage, add a whitelist at
/etc/ni-provisioning/mmc_storage_device_codes.allow. If the device's
`DeviceCode` smbios entry matches one of the entries in that whitelist,
MMC provisioning will be permitted.

---

# Summary
* Adds a DeviceCode whitelist at `/etc/ni-provisioning/mmc_storage_device_codes.allow` - which works like you would expect.
  * `/etc/ni-provisioning/` is a new path. I think it makes sense to put this in `/etc/`, since I think it makes FHS-sense, and because I can see us or a user wanting to modify the whitelist to either workaround a bug, or force their device to consider MMC.
  * I think we should start moving the provisioning scripts and utilities out of `/`.  I'm considering some minor rearchitecture commits as a part of the interactive provisioning changes (to come), like moving the `ni_provisioning.common` source-script (and others) to `/usr/lib/ni-provisioning/`. Let me know if there are any objections.
* Adds some skip-logic to the `ni_provisioning.common` script to avoid any block device whose basename starts with `mmc`, unless the DeviceCode (as reported by `fw_printenv`) is present in the whitelist.

**When `DeviceCode` is undefined (as in many VMs) or a zero-length string (invalid)**, the provisioning script will avoid MMC.

**When the `/etc/ni-provisioning/mmc_storage_device_codes.allow` file does not exist**, the script will print the following warning to stderr, and avoid MMC.
```
Warning: eMMC device whitelist not found. eMMC block devices will be disqualified from provisioning.
```

**When verbose provisioning is enabled, the device is not allowed, and an MMC block device is detected,** the script prints:
```
SKIPPED: mmcblk0: mmc block device && MMC provisioning is not permitted for this device.
```

I'm still investigating which DeviceCodes need to be in the whitelist; I will add them in future PRs.

Some devices seem to be captured by the `sdhci-acpi` driver instead (including the ISC-1781. I don't think it makes a substantive difference to this implementation.

# Testing
By adding `-device sdhci-pci -device sd-card,drive=sdcard -drive id=sdcard,if=none,format=raw,file=/home/usr0/projects//qemu-sd/static/blank-sd.raw` to the qemu args, you can create an SDHCI-PCI controller, populated with a block device. It gets captured by the `sdhci-pci` driver (just as on FireEagle targets), and udev populates its sysfs entry similarly as well. Without this commit, I can reliably reproduce [bug 58653](https://dev.azure.com/ni/DevCentral/_boards/board/t/DSW%20DOE%20RTOS/Work%20Items/?workitem=58653). 

@ni/rtos


recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -154,9 +155,12 @@ elif [ "$ARCH" = "x86_64" ]; then
...
 
     # For a given block device (represented by sysfs path $1 and block device path
-    # $2), return zero if suitable for provisioning.
+    # $2), return zero if suitable for provisioning. If $3 is not True, skip MMC
+    # block devices.
     check_block_device () {
-        local path=$1 block=$2
+        local path=$1
+        local block=$2
+        local mmc_ok=${3:false}
```
amstewart: :/ Forgot a `-` sign here. Will fix.

amstewart:
# Patch V2
* Actually use default variable substitution in the `check_block_device()` function. :)

billpittman:
For this commit, I think we should not do a "partial" rearchitecture of the provisioning scripts.  I think the mmc*.allow file should be named similar to the ni_provisioning.answers.default file and installed next to that file as well.

I fully understand your interest in rearchitecting the provisioning script(s), but doing part of it in this commit seems messy to me.  Also if we never get around to rearchitecting these scripts (which would never happen), then all the provisioning scripts/files would be co-located (even if that's "wrong" ... it remains consitently "wrong").

billpittman:
NITPICK (feel free to ignore/reject) ... the format of the allow file should allow for a comment on the same line as the device code, that way its easier for the people to list device codes, a simple comment for what it is, and keep the list in numerical order.

ie
0x1234   # This device
0xF426   # That device

Right now adding a comment on the line, breaks the functionality, and commenting the line isn't unreasonable.

amstewart:
> For this commit, I think we should not do a "partial" rearchitecture of the provisioning scripts.
> I fully understand your interest in rearchitecting the provisioning script(s), but doing part of it in this commit seems messy to me. 

I'm going to refactor more of the scripts in a PR-to-come, when I handle the interactive case. That implementation is necessarily going to touch a lot more of the scripts, and I was planning on moving them into *better* places in the filesystem. Do you disagree with doing that at all?

>Also if we never get around to rearchitecting these scripts (which would never happen), then all the provisioning scripts/files would be co-located (even if that's "wrong" ... it remains consitently "wrong").

Consistent with what? Not the FHS or other distro installation tools (which use standard FHS locations, I'm pretty sure). And I don't think anyone is building tooling based on the location of our provisioning scripts.

>I think the mmc*.allow file should be named similar to the ni_provisioning.answers.default file

In what way?

amstewart:
>NITPICK (feel free to ignore/reject) ... the format of the allow file should allow for a comment on the same line as the device code, that way its easier for the people to list device codes, a simple comment for what it is, and keep the list in numerical order.

I wanted to do that, but couldn't get the regex to be robust and concise in the ~8 minutes I was willing to spend on it. If you have a reasonably laconic implementation, I'm happy to integrate it.

amstewart:
# Patch V3
* Added more eMMC storage devices, based on a best-effort BOM traversal.
  * I went looking around in Echo for devices with eMMC memory modules. Most of them were covered by [this component](https://echo.natinst.com/part/whereused/215383457/269695468). To confirm, I spot-checked manuals for the controllers I found for discussion about the eMMC storage device. And I added DeviceCodes based on what is declared in the [SystemImage CDF](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Frtos%2Frtosinstallersd%2Fnilrt_onert_system_image%2Fsource%2Frss%2Fcdf%2Flinux-rt-rss.cdf) - which is the most authoritative source I could find for what is supported.
  * Notably, the ISC hardware is manufactured externally. So they are an example of a whitelisted device which is not in that Echo query.
  * I also checked the Ettus hardware - which use sd cards frequently. But I think all the devices in those cases either didn't release with NILRT support (like the N310), or only use the sdcard for secondary storage.

billpittman:
> > For this commit, I think we should not do a "partial" rearchitecture of the provisioning scripts.
> > I fully understand your interest in rearchitecting the provisioning script(s), but doing part of it in this commit seems messy to me.
> 
> I'm going to refactor more of the scripts in a PR-to-come, when I handle the interactive case. That implementation is necessarily going to touch a lot more of the scripts, and I was planning on moving them into _better_ places in the filesystem. Do you disagree with doing that at all?
> 
No disagreement at all, I just think it may be easier (or cleaner) to do the refactoring all at once instead of bits at a time.  Honestly my only concern is that we will deprioritize and end up with half the refactoring done, in which case, I think we'd be better off staying consistent with our current implementation where all the provisioning files live next to one another and have a similar name.
> > Also if we never get around to rearchitecting these scripts (which would never happen), then all the provisioning scripts/files would be co-located (even if that's "wrong" ... it remains consitently "wrong").
> 
> Consistent with what? Not the FHS or other distro installation tools (which use standard FHS locations, I'm pretty sure). And I don't think anyone is building tooling based on the location of our provisioning scripts.
> 
> > I think the mmc*.allow file should be named similar to the ni_provisioning.answers.default file
> 
> In what way?
Consistent in name and location instead of putting it in /etc right now.  If for some reason we don't get to the refactoring, it seems consistent to me to keep with the existing naming practices (even if they're going to change).

recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -217,11 +231,23 @@ elif [ "$ARCH" = "x86_64" ]; then
     find_target_block_device()
     {
         local selected_block=""
+        local mmc_ok=false
+
+        # Check if the machine's DeviceCode is in the MMC device whitelist.
+        if [ -e "${MMC_DEVICE_ALLOW_PATH}" ]; then
+                device_code=$(fw_printenv -n DeviceCode 2>/dev/null)
+                [ -n "$device_code" ] && \
+                        grep -iq -E "^\s*(0x)?${device_code}\s*$" "${MMC_DEVICE_ALLOW_PATH}" && \
```
billpittman: I like regular expressions a lot, love the matching, hate how easy they are to get wrong *laugh*.  That being said, could you change it to "^\s*(0x)?${device_code}\s*(#.*)?$"

This is a nitpick honestly, so I'll leave it to you on how you want to implement it (or leave it as is).

amstewart:
>No disagreement at all, I just think it may be easier (or cleaner) to do the refactoring all at once instead of bits at a time. Honestly my only concern is that we will deprioritize and end up with half the refactoring done, in which case, I think we'd be better off staying consistent with our current implementation where all the provisioning files live next to one another and have a similar name.

@billpittman and I discussed offline, and we agree that its fine to put the `mcc_*.allow` file in `/etc/`, since I'm currently working on the more-complete recipe re architecture.

recipes-core/initrdscripts/files/ni_provisioning.common
amstewart: I'll give that a whirl.

amstewart:
# Patch V4
* Implemented @billpittman 's inline-comment-aware regex string in `ni_provisioning.common`.
* Added inline comments to the entries in `mmc_storage_device_codes.allow`, to describe the products which correspond to each device code.

## Testing
* Installed the new scriptlet and mmc.allow file. Tested various device codes and verified that the whitelist matching function still works as expected.

billpittman (approved):

shruthi-ravi (approved):

PR #189, opened 2021-06-09
nilrt.conf: bump distro version to 8.9

amstewart:
Service for the planned 21Q3 release.

Signed-off-by: Alex Stewart <alex.stewart@ni.com>

## Maintainership Note
Only merge this PR after the heisenberg 21.0 pipeline has been switched to use the `nilrt/21.0/sumo` ref, and there exists a new - 21.1 pipeline - set to build the mainline branch.

PR #197, opened 2021-06-21
nilrt.conf: bump NILRT version

jpautler:

### DIFF
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -4,7 +4,7 @@ DISTRO_NAME = "NI Linux Real-Time"
 
 DISTRO_VERSION = "8.9"
 
-NILRT_FEED_NAME = "2021.0"
+NILRT_FEED_NAME = "2021.3"
 
 DISTRO_FEATURES_append_x64 = "\
         x11 \
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -2,7 +2,7 @@ require nilrt.inc
 
 DISTRO_NAME = "NI Linux Real-Time"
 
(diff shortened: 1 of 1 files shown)


conf/distro/nilrt.conf
```diff
@@ -2,7 +2,7 @@ require nilrt.inc
 
 DISTRO_NAME = "NI Linux Real-Time"
 
-DISTRO_VERSION = "8.8"
+DISTRO_VERSION = "8.9"
 
 NILRT_FEED_NAME = "2021.0"
```
shruthi-ravi: What about this variable? Is the change coming later?

shruthi-ravi (approved):

amstewart:
Do you just want me to pull #189 instead?

Edit: Oh; wait. We need to set the new feed name to `2020.3`. :/ I guess I'll drop #189, and we can make that change in this PR?

conf/distro/nilrt.conf
jpautler: I had been waiting to update the value of NILRT_FEED_NAME because it is used to build the feed URLs and we don't yet have ni-* feeds ready for 2021.3. However, we decided today it's better to update now and just let the feeds be broken until those feeds are created. I'll push an update to this PR for that.

shruthi-ravi:
That second merge commit is odd. There aren't any changes in it? Can you remove that commit?

jpautler:
> That second merge commit is odd. There aren't any changes in it? Can you remove that commit?

Done. Sorry, I think I had messed something up when trying to fetch other changes from master to my dev branch. I think it's good now.

amstewart (approved):

PR #199, opened 2021-06-23
kernel-test-nohz: Fix test duration math

gratian:
The test function for the kernel-test-nohz ptest was incorrectly computing
the test end time based on the global 'test_duration' variable as opposed
to the 'duration' parameter passed in. This resulted in tests running for
double the amount specified due to the warm-up iterations taking the
full 'test_duration'. Fix this by using the local variable as intended.


gratian:
FYI: @ni/rtos

billpittman (approved):

amstewart (approved):

PR #200, opened 2021-06-23
kernel-test-nohz: Add postinst/postrm script fragments for configuring CPU isolation

gratian:
The NO_HZ_FULL kernel ptest relies on CPU isolation for running an
undisturbed RT polling test.

Add a postinst script fragment that changes the kernel boot parameters via
the 'othbootargs' variable to include the necessary 'isolcpus' and
'nohz_full' parameters configured to use the last CPU available on the
system under test. These parameters isolate that CPU and flag it for
stopping the scheduler tick whenever possible.

The corresponding postrm script fragment undoes changes to 'othbootargs'
(which is normally empty).

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-test-nohz.bb
+++ b/recipes-kernel/kernel-tests/kernel-test-nohz.bb
@@ -31,4 +31,25 @@ do_install_ptest_append() {
     install -m 0755 ${S}/nohz_test ${D}${PTEST_PATH}
 }
 
+pkg_postinst_ontarget_${PN}-ptest_append() {
+    CPUS=`nproc`
+    ISOLATED_CPU=$((CPUS - 1))
+    OTHBOOTARGS=`fw_printenv othbootargs 2>/dev/null | sed -e "s/^othbootargs=//g" | sed -e "s/isolcpus=[^ ]*[ ]*\|nohz_full=[^ ]*[ ]*//g"`
+
+    if [ $ISOLATED_CPU -gt 0 ]; then
+        fw_setenv othbootargs $OTHBOOTARGS 'isolcpus='$ISOLATED_CPU' nohz_full='$ISOLATED_CPU
+    else
+        echo "[kernel-test-nohz:error] This test requires a system with 2 or more CPUs"
+        exit 1
+    fi
+
+    echo "[kernel-test-nohz:info] This test requires a reboot after install for kernel changes to take effect"
+}
+
+pkg_postrm_${PN}-ptest_append() {
+    OTHBOOTARGS=`fw_printenv othbootargs 2>/dev/null | sed -e "s/^othbootargs=//g" | sed -e "s/isolcpus=[^ ]*[ ]*\|nohz_full=[^ ]*[ ]*//g"`
+    fw_setenv othbootargs $OTHBOOTARGS
+    echo "[kernel-test-nohz:info] This test requires a reboot after uninstall for kernel changes to take effect"
+}
+
 PACKAGE_ARCH = "${MACHINE_ARCH}"


gratian:
FYI: @ni/rtos

recipes-kernel/kernel-tests/kernel-test-nohz.bb
```diff
@@ -31,4 +31,20 @@ do_install_ptest_append() {
...
+    if [ $ISOLATED_CPU -gt 0 ]; then
+        fw_setenv othbootargs 'isolcpus='$ISOLATED_CPU' nohz_full='$ISOLATED_CPU
+    else
+        echo "[error] This test requires a system with 2 or more CPUs"
+        exit 1
+    fi
+}
+
+pkg_postrm_${PN}-ptest_append() {
+    fw_setenv othbootargs
```
billpittman: I'm concerned that this would unset other othbootargs from other ipks as well as this one.

gratian: That's fair. In practice I don't expect this test to ever get uninstalled (rather the test system gets re-formatted) but let me see if I can do better.

billpittman: Ah right, I forgot this was a ptest.  Then in that case, do better if you want, or simply leave it.  In fact you could even remove the postrm all together.  There's no real harm in leaving othbootargs "dirty" in a test where the system gets formatted.

recipes-kernel/kernel-tests/kernel-test-nohz.bb
```diff
@@ -31,4 +31,20 @@ do_install_ptest_append() {
     install -m 0755 ${S}/nohz_test ${D}${PTEST_PATH}
 }
 
+pkg_postinst_ontarget_${PN}-ptest_append() {
+    CPUS=`nproc`
+    ISOLATED_CPU=$((CPUS - 1))
+
+    if [ $ISOLATED_CPU -gt 0 ]; then
+        fw_setenv othbootargs 'isolcpus='$ISOLATED_CPU' nohz_full='$ISOLATED_CPU
```
amstewart: Does this imply that the ptest requires a reboot between package installation and executing the test?

In that case, I think it would be more appropriate to include a tool with the ptest subpackage which the user can intentionally run and insert these options into their cmdline (or run-in-reverse to remove them). Setting them at package-install time seems inappropriate to me.

amstewart (changes requested):

recipes-kernel/kernel-tests/kernel-test-nohz.bb
gratian: Yes, this test requires a reboot; however:

- changing the kernel command line is not the only reason a reboot is required for this test. The main reason is it depends on and installs a new kernel (configured for NO_HZ_FULL). Basically even w/o these postinst/postrm scripts a reboot will be required.
- the ATS this is going to be run in already has a reboot in place after the install.

I would prefer to keep this as a postinst script because it greatly simplifies the main use case: running this ptest as part of the ATS. With the current implementation all the ATS needs to do is install this ptest, reboot and run ptest-runner. If I start to introduce custom scripts that means the ptest is no longer self contained and the ATS framework now needs to know that for this particular ptest additional scripts need to be called (to change the command line). This adds unneeded complication.

For somebody running this test by hand I can print a message advising about a reboot being required. I expect manually running this test to be a corner case that's not encountered very often outside the ATS.

recipes-kernel/kernel-tests/kernel-test-nohz.bb
gratian: I've added some sed magic in v2 to only add/remove the 'isolcpus' and 'nohz_full' parameters.

gratian:
Changes since V1:

-  I've added some sed magic to preserve the contents of 'othbootargs' on postinst/postrm and only append or remove the 'isolcpus' and 'nohz_full' parameters.

- I've tested that this functions correctly and doesn't clobber any parameters that might have been added by the user before or after installing this test.

billpittman (approved):

shruthi-ravi (approved):

PR #202, opened 2021-06-24
fw_printenv: Fix variable values getting truncated on '=' symbols

gratian:
Using the current fw_printenv implementation to print the value for a
single grub variable results in truncated output if the value field
contains '=' symbols.

e.g.

```
  # fw_printenv othbootargs
  othbootargs=isolcpus

```
instead of the expected:

```
  # fw_printenv othbootargs
  othbootargs=isolcpus=7 nohz_full=7
```

Fix the 'cut' call used to extract the variable value in
show_variable_grub() to include the rest of the line after the 'key' field.

Signed-off-by: Gratian Crisan <gratian.crisan@ni.com>

PR #203, opened 2021-06-24
fw_printenv: Fix variable values getting truncated on '=' symbols

gratian:
Using the current fw_printenv implementation to print the value for a
single grub variable results in truncated output if the value field
contains '=' symbols.

e.g.

```
  # fw_printenv othbootargs
  othbootargs=isolcpus
```

instead of the expected:

```
  # fw_printenv othbootargs
  othbootargs=isolcpus=7 nohz_full=7
```

Fix the 'cut' call used to extract the variable value in
show_variable_grub() to include the rest of the line after the 'key' field.

### DIFF
--- a/recipes-ni/ni-utils/files/x64/fw_printenv
+++ b/recipes-ni/ni-utils/files/x64/fw_printenv
@@ -43,7 +43,7 @@ show_variable_grub(){
 			key=$(echo "$line" | cut -d "=" -f 1)
 			#if the $1 matches exactly the key, extract the value
 			if [ "$key" = "$1" ] ; then
-				value_tmp=$(echo "$line" | cut -d "=" -f 2)
+				value_tmp=$(echo "$line" | cut -d "=" -f 2-)
 				echo "$value_tmp"
 				break
 			fi


gratian:
@ni/rtos This is the Sumo version of this Dunfell PR: https://github.com/ni/meta-nilrt/pull/202
Same testing comment applies here.

recipes-ni/ni-utils/files/x64/fw_printenv
```diff
@@ -43,7 +43,7 @@ show_variable_grub(){
 			key=$(echo "$line" | cut -d "=" -f 1)
 			#if the $1 matches exactly the key, extract the value
 			if [ "$key" = "$1" ] ; then
-				value_tmp=$(echo "$line" | cut -d "=" -f 2)
+				value_tmp=$(echo "$line" | cut -d "=" -f 2-)
```
shruthi-ravi, line 46: Does the ARM version work as is for these new `othbootargs`? Or is it not applicable in ARM?
Also, I just realized that there is a version of this file in [os-common](https://p4.natinst.com/browser/penguin/nilinux/os-common/trunk/8.8/source/arch/x64/fw_printenv). And, I'm not sure which one of these get installed to the target.

gratian: ARM uses the u-boot version of fw_printenv/setenv which doesn't have this problem.

I've noticed the os-common version too but I don't see it referenced in any of the makefiles. I have a p4 change ready for os-common with this change but I was waiting to see if it's necessary or if that version is dead code.

shruthi-ravi (approved):

amstewart (approved):

PR #206, opened 2021-07-06
linux-nilrt-debug: disable lockdep

bstreiff:
Enabling DEBUG_LOCK_ALLOC (which is enabled by PROVE_LOCKING and LOCK_STAT)
results in inlining a call into a GPL-only symbol on every RCU lock access,
and RCU lock accesses are frequently done via inline functions across various
subsystems. This option is therefore highly incompatible with non-GPL modules.
Turn these off in the debug kernel configuration so that the debug kernel can
still be used with such modules.

Natinst-AzDO-ID: 1490704


amstewart (approved):

amstewart:
@gratian ACK?

gratian (approved):

PR #210, opened 2021-07-09
kernel-test-nohz: Fix detection of CPUs in postinst script

gratian:
The `nproc` utility by default returns the number of CPUs available to the
current (nproc) process.  When cgroup CPU sets are used to set the CPU mask
for system processes (like the nohz test does) this number can be less than
the total number of CPUs available.  This messes up the logic for setting
kernel boot parameters for CPU isolation and nohz functionality during the
post-install step.

Fix it by using `nproc --all` which returns the total number of CPUs
installed.  This option is not supported by the BusyBox version of `nproc`
so we also need to add a run-time dependency on coreutils.

Fixes: 1ac78688b87e ("kernel-test-nohz: Add postinst/postrm script fragments for configuring CPU isolation")

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-test-nohz.bb
+++ b/recipes-kernel/kernel-tests/kernel-test-nohz.bb
@@ -10,7 +10,7 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}-files:"
 
 S = "${WORKDIR}"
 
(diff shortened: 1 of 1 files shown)


gratian:
FYI: @ni/rtos

amstewart (approved):

billpittman (approved):

PR #211, opened 2021-07-12
linux: upgrade kernel and debug kernel to 5.10

amstewart:

---

Service for the 21Q3 release.

Also see [OE-core PR #27](https://github.com/ni/openembedded-core/pull/27) for a sibling commit which fixes a warning with the 5.10 kernel build on sumo.

## Testing
Builds on my dev machine. @gratian; should I hand-validate these new IPKs. Or are we reasonably sure that they won't break downstream clients?

@ni/rtos


amstewart:
Closing this PR in favor of a revision which affects only x64.

PR #194, opened 2021-06-17
Upgrade uboot branch to 2017

kimkhiam:
Tested with roboRIO 2.0 with the steps below:
1. build safemode with the OE that consist of the fw_printenv
2. copy the safemode to roboRIO 2.0
3. boot up roboRIO 2.0 and run fw_printenv, able to see all env variable
4. run fw_setenv to update bootdelay variable and perform reset
5. boot into safemode again and run fw_printenv, able to see the updated bootdelay variable
6. reboot the system and boot into UBOOT
7. update the bootdelay variable in UBOOT and reset the system
8. boot into safemode, run fw_printenv the bootdelay variable display the value as in configure in UBOOT.

PR #214, opened 2021-07-14
Update u-boot LIC_FILES_CHKSUM

amstewart:
meta-nilrt #194 upgraded the u-boot source from from 2012 to 2017. Over that timeframe, u-boot upstream changed the location and architecture of their license files. `u-boot-fw-utils:do_populate_lic` fails because the license file is no longer at `:COPYING`.

This PR updates the license files' locations and md5sums to match what ships with 2017. As a nice side-effect, we're now packaging *all* of the licenses which ship with u-boot, rather than just u-boot's COPYING. :)

This PR also compresses the source meta-data into the includes file - which should be a little easier to maintain.

## Testing
Rebuilt all the `u-boot*.bbappend` recipes in the recipe directory. `u-boot-fw-utils` now builds correctly. `u-boot_2018` fails do_compile - but I think it has always failed. And `-mkimage` still passes (I don't think we use that one anyway).

FYI @kimkhiam


shruthi-ravi (approved):
Thanks for fixing the build!

PR #218, opened 2021-08-05
Move GRUB_BRANCH var to the grub recipe directory

amstewart:
The `GRUB_BRANCH` variable is used by (a) our `grub*` recipes and (b) the `safemode-image` recipe in `recipes-org/`.

The safemode-image recipe only uses it to write its value into the `/payload/imageinfo` file. But a search of the codebase suggests that no one actually uses that key-value pair, and the [original commit](https://github.com/ni/meta-orgconf/commit/a6a70ead04e5c2b37b2fb62493fe3162fe86be06) doesn't provide any more context.

So remove the KVP from the imageinfo file, and move the GRUB_BRANCH variable from the distro.conf, back into the grub recipes.

## Testing
* Rebuilt the grub recipes and verified that they still work.
* I can't rebuild the safemode-image recipe on my dev machine, but it's very unlikely that the lines I removed could break the recipe build anyway. And that recipe is also unlikely to be seriously used for a while yet.

@ni/rtos

### DIFF
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -2,8 +2,6 @@ MAINTAINER = "NI Linux Real-Time Maintainers <nilrt@ni.com>"
 
 TARGET_VENDOR = "-nilrt"
 
-GRUB_BRANCH = "nilrt/19.0"
-
 # Inherit the default LIBC features superset from OE-core
 DISTRO_FEATURES += "${DISTRO_FEATURES_LIBC}"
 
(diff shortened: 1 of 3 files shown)


billpittman (approved):

shruthi-ravi:
What was the motivation for this change? General cleanup?

amstewart:
> What was the motivation for this change? General cleanup?

It was a part of servicing [this NI work item](https://dev.azure.com/ni/DevCentral/_workitems/edit/1549901). Originally, this PR would have also included switching the `GRUB_BRANCH` over to use the new `main` ref. But PXI PS have expressed that they do not want to put their commits into grub until the 21Q4 release.

PR #219, opened 2021-08-09
Revert "Move GRUB_BRANCH var to the grub recipe directory"

amstewart:
Reverts ni/meta-nilrt#218

Somehow, my codesearch missed [this ni_provisioning.safemode](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-core/initrdscripts/files/ni_provisioning.safemode#L377) script call, which tests the `GRUB_VERSION` string in `/payload/imageinfo`. Without the string, the provisioning scripts error during provisioning. So it isn't safe and trivial to drop this variable from the safemode recipes. :(

@ni/rtos

### DIFF
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -2,6 +2,8 @@ MAINTAINER = "NI Linux Real-Time Maintainers <nilrt@ni.com>"
 
 TARGET_VENDOR = "-nilrt"
 
+GRUB_BRANCH = "nilrt/19.0"
+
 # Inherit the default LIBC features superset from OE-core
 DISTRO_FEATURES += "${DISTRO_FEATURES_LIBC}"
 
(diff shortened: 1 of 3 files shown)


gratian (approved):

amstewart:
By the way: while root-causing the build failure that led to this revert, I discovered that the `BUILD_IDENTIFIER` string in `imageinfo` is blank.

```bash
++ set_versions
+++ get_image_info BUILD_IDENTIFIER
+++ TOKEN=BUILD_IDENTIFIER
+++ grep -Po '(?<=BUILD_IDENTIFIER=).*' /payload/imageinfo
++ BUILD_IDENTIFIER=
+++ get_image_info GRUB_VERSION
+++ TOKEN=GRUB_VERSION
+++ grep -Po '(?<=GRUB_VERSION=).*' /payload/imageinfo
++ GRUB_IDENTIFIER=
```

So I'm still skeptical that we're *really* using these variarbles. But they are at least more tightly integrated into the provisioning scripts than I expected.

shruthi-ravi (approved):

PR #222, opened 2021-08-17
nisystemformat: flesh out partial recovery of failed reformats

rtollert:
Previously, the only recovery that occurred here was that rootfs format
failures would mount_rootfs on exit. Here, we extend that to the entire
configuration/service management stack surrounding the format operation, so
that nondestructive failures are generally fully recovered from To simplify the
cleanup logic, we break format_rootfs_or_userfs() into a chain of smaller
functions, each of which is responsible for setup/cleanup of a specific layer
of the stack.

Copyright bump.

Natinst-AzDO-ID: 1505704
Signed-off-by: Richard Tollerton <rich.tollerton@ni.com>

### DIFF
--- a/recipes-ni/ni-systemformat/files/nisystemformat
+++ b/recipes-ni/ni-systemformat/files/nisystemformat
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (c) 2021 National Instruments
+# Copyright (c) 2013-2021 National Instruments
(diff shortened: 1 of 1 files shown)


amstewart (approved):
Thanks!

amstewart:
It will be some weeks still until we have a full hardknott pipeline that we can use to test this PR, so I'm just going to pull it now.

PR #226, opened 2021-08-18
Drop the meta-nilrt libnss-mdns bbappend

amstewart:
The contents of the `libnss-mdns` bbappend are outdated. The make configuration options it asserts were deprecated by upstream, and we no longer need to drop its avahi dependencies - since we ship avahi in the base images now.

I'm not 100% sure that we still don't need whatever `legacy` mode was enabled by the configuration, so I created [AZDO-1585850](https://dev.azure.com/ni/DevCentral/_workitems/edit/1585850) to track checking that the bug which motivated that config doesn't reproduce.

For now, I'm dropping the whole of the bbappend.

## Testing
`libnss-mdns` now builds without error.

@ni/rtos

### DIFF
--- a/recipes-connectivity/libnss-mdns/libnss-mdns_0.%.bbappend
+++ b/recipes-connectivity/libnss-mdns/libnss-mdns_0.%.bbappend
@@ -1,4 +1,2 @@
-EXTRA_OECONF = "--libdir=${base_libdir} --disable-lynx --enable-legacy"
-
 DEPENDS = ""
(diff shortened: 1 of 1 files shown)


shruthi-ravi (approved):
If this was needed, seems like it would show up when connecting to the device from MAX, maybe? Your changes are reasonable for now, but I hope we remember the tracking item if we run into issues.

gratian (approved):

PR #227, opened 2021-08-18
Reconcile rauc recipe with meta-rauc/hardknott.

amstewart:
This PR
1. bypasses a QA check using INSANE_SKIP. 
  * Depends on [oe-core PR#30](https://github.com/ni/openembedded-core/pull/30).
2. drops the "ignore bundle verification" CLI arg .patch.
  * I added a step to the [RAUC rebase work item](https://dev.azure.com/ni/DevCentral/_workitems/edit/1582462) to rebase this .patch upon rauc upstream.

## Testing
The `rauc` recipe now builds.

@ni/rtos


recipes-core/rauc/nilrt_rauc.inc
```diff
@@ -1,6 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
 PACKAGECONFIG_remove += "network json service"
 
+INSANE_SKIP_${PN} += "native-last"
```
shruthi-ravi, line 4: Any idea what can go wrong when we don't inherit last?

shruthi-ravi (approved):

billpittman (approved):

recipes-core/rauc/nilrt_rauc.inc
amstewart: btibake says: `native/nativesdk class is not inherited last, this can result in unexpected behaviour.`

There are like 10 ML threads about this, but none of them provide any more info. The best I can find is [yocto bug #5729](https://bugzilla.yoctoproject.org/show_bug.cgi?id=5729).

>Since these are BBCLASSEXTEND operations which are usually done last, if a user manually inherits them, together with other classes (like packagegroup for example), the behavior might not be what the user wanted (an allarch package instead of two SDK_ARCH dependent packages, for instance). This might lead to unnecessary rebuilds and warnings for duplicate files, etc.

So it doesn't sound like something that would cause an issue between `-native` and `update-rc.d`.

PR #231, opened 2021-08-19
pkggrps: remove u-boot-fw-utils

amstewart:
Primary commit:

    pkggrps: remove u-boot-fw-utils

    u-boot-fw-utils has been deprecated by OE upstream in favor of
    libubootenv.
    
    Since ARM32 targets aren't supported in this branch, we have no need for
    either package. So remove u-boot-fw-utils from the packagegroups
    entirely.

----

Bonus commit:

    u-boot: remove all .patch files
    
    The hardknott mainline has already cherry-picked changes from the
    dunfell mainline which deprecated all of the u-boot/ recipes. However,
    between the merge, a u-boot .patch file was left unattended.
    
    Remove that .patch file for cleanliness.

@ni/rtos


recipes-core/packagegroups/packagegroup-ni-base.bb
```diff
@@ -17,7 +17,6 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS ?= ""
 
 ALL_DISTRO_ARM_PACKAGES = "\
```
shruthi-ravi, line 18: Why aren't you removing all ARM-specific packages entirely? I have a few guesses, but want to hear from you explicitly.

shruthi-ravi (approved):

recipes-core/packagegroups/packagegroup-ni-base.bb
amstewart: I'm going to do that in an independent patchset, once I have everything in the base packagegroup building and the concerning warnings are resolved.

PR #233, opened 2021-08-19
pstore-save: add bash to -ptest RDEPENDS

amstewart:
Newer OE package QA checks will error when building pstore-save. OE
detects that the run-ptest script in the pstore-save-ptest subpackage
uses a /bin/bash shebang line, but that the -ptest subpackage doesn't
declare an dependency on any package which provides /bin/bash.

Give the -ptest subpackage an RDEPENDS on `bash` to provide the bash
binary.

----

OE error looks like this:

```make
ERROR: pstore-save-1.0-r0 do_package_qa: QA Issue: /usr/lib/pstore-save/ptest/run-ptest contained in package pstore-save-ptest requires /bin/bash, but no providers found in RDEPENDS_pstore-save-ptest? [file-rdeps]
```

And it's actually kinda' neat that OE detected this.

## Testing
`pstore-save` now builds without error.

@ni/rtos

### DIFF
--- a/recipes-ni/pstore-save/pstore-save_1.0.bb
+++ b/recipes-ni/pstore-save/pstore-save_1.0.bb
@@ -30,7 +30,7 @@ do_install () {
 
 inherit ptest
 
-RDEPENDS_${PN}-ptest += "${PN}"
+RDEPENDS_${PN}-ptest += "${PN} bash"
 
 do_install_ptest_append () {
     install -d ${D}${PTEST_PATH}/src


recipes-ni/pstore-save/pstore-save_1.0.bb
```diff
@@ -30,7 +30,7 @@ do_install () {
 
 inherit ptest
 
-RDEPENDS_${PN}-ptest += "${PN}"
+RDEPENDS_${PN}-ptest += "${PN} bash"
```
shruthi-ravi, line 33: The 'packagegroup-core-device-devel.bb` package group you just removed, included bash. I assume there are other package groups that ensure bash is available in the feeds?

shruthi-ravi (approved):

recipes-ni/pstore-save/pstore-save_1.0.bb
amstewart: Bash is such a common dependency that I'm sure it does end up getting built into the core feed, but it is only explcitly declared as a part of the -restoremode packagegroup. I'll add it to -base, since that will eventually mutate into the core feed packagegroup.

PR #237, opened 2021-08-25
linux-nilrt-debug: Move debug vmlinux out of /boot

jpautler:
The debug vmlinux file is significantly larger than the entire boot
partition. As a result, installing it was causing errors.

This file is not actually used by the target, but is intended to be
copied to a host machine running kgdb. Therefore it does not need to be
installed on the boot partition. This change moves it out of the boot
partition and into /lib/modules/<kernel-version>/build.

### DIFF
--- a/recipes-kernel/linux/linux-nilrt-debug_git.bb
+++ b/recipes-kernel/linux/linux-nilrt-debug_git.bb
@@ -13,3 +13,14 @@ SRC_URI += "\
 # This is the place to overwrite the source AUTOREV from linux-nilrt.inc, if
 # the kernel recipe requires a particular ref.
 #SRCREV = ""
+
+# Move vmlinux-${KERNEL_VERSION_NAME} from /boot to /lib/modules/${KERNEL_VERSION}/build/
+# to avoid filling the /boot partition.
+FILES_${KERNEL_PACKAGE_NAME}-vmlinux_remove = "/boot/vmlinux-${KERNEL_VERSION_NAME}"
+FILES_${KERNEL_PACKAGE_NAME}-vmlinux += "${nonarch_base_libdir}/modules/${KERNEL_VERSION}/build/vmlinux-${KERNEL_VERSION_NAME}"
+
+do_install_append(){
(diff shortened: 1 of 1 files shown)


amstewart (approved):
Please also cherry pick this over to the `nilrt/master/hardknott` mainline, once it is accepted.

gratian (approved):
LGTM.
A very minor nitpick: in the commit message you say "host machine running kgdb" - the host is actually running plain old gdb. 'kgdb' is the kernel piece on the target side that acts like an in-kernel gdb server.

recipes-kernel/linux/linux-nilrt-debug_git.bb
```diff
@@ -13,3 +13,14 @@ SRC_URI += "\
 # This is the place to overwrite the source AUTOREV from linux-nilrt.inc, if
 # the kernel recipe requires a particular ref.
 #SRCREV = ""
+
+# Move vmlinux-${KERNEL_VERSION_NAME} from /boot to /lib/modules/${KERNEL_VERSION}/build/
+# to avoid filling the /boot partition.
+FILES_${KERNEL_PACKAGE_NAME}-vmlinux_remove = "/boot/vmlinux-${KERNEL_VERSION_NAME}"
+FILES_${KERNEL_PACKAGE_NAME}-vmlinux += "${nonarch_base_libdir}/modules/${KERNEL_VERSION}/build/vmlinux-${KERNEL_VERSION_NAME}"
```
shruthi-ravi: Can you put `${nonarch_base_libdir}/modules/${KERNEL_VERSION}/build` in a variable?

shruthi-ravi (approved):

recipes-kernel/linux/linux-nilrt-debug_git.bb
jpautler: Done.

jpautler:
I incorporated Shruthi's feedback and also updated the commit message to gdb vs kgdb per Gratian's recommendation.

PR #240, opened 2021-09-09
FIx looping kernel-devsrc postinst on systems living in the past.

amstewart:
The NI implementation of the `kernel-devsrc` packages have a postinst which `cd`s into the kernel source directory and invokes `make` to prepare the kernel modules. GNU `make` uses file modification times to determine which target dependencies require rebuilding. When `opkg` unpacks the `kernel-devsrc` data segment, it intentionally does not refresh the file modification times; which means that the extracted files have their mtimes set to when they were *packed* by the IPK build server. On systems where the software clock is set backwards in time (either due to not having an RTC, or being enslaved to a network time server), it is possible for the system time to be behind the date on the source files' mtime.

Normally, you would expect `make` to skip creating some targets in this case; and `make` indeed throws a warning during compilation to that affect. However, the `kernel-devsrc` package also includes the `objtool` utility, which generates - as a consequence of its Makefile target - intermediate files which are reingested by `make`. Some exotic interaction in this special circumstance causes the `kernel-devsrc` postinst to loop indefinitely. The system appears to hang in this case.

The thrust of this PR is to fix this loop by reconciling the mtime on the `kernel-devsrc` data files. Before invoking `make`, the postinst will now find and touch the mtime of all the source files - rolling them back to whenever the software clock is set. The remainder of this PR cleans up the includes files which are shared between the `kernel-devsrc`(prime) bbappend and the *alternative* kernel flavors' recipes.

[NI-AZDO #1605300](https://dev.azure.com/ni/DevCentral/_workitems/edit/1605300)

Note: the `kernel-devsrc` postinst will still throw a warning during compilation, alerting the user that the `/usr/include/stdc-predef.h` file is living in the future (for the same reason as above.) However, this doesn't seem to be a problem for the Makefile, and I don't think its appropriate for `kernel-devsrc` to refresh its mtime, because the file is owned by the `libc6-dev` package.

As a result, the user might see a few iterations of this warning during compilation:

```make
make[4]: warning:  Clock skew detected.  Your build may be incomplete.
make[3]: Warning: File '/usr/include/stdc-predef.h' has modification time 670029495 s in the future
```

## Testing
* Rebuilt all of the `kernel-devsrc{-*}` recipes. Verified that the resulting packages look correct.
* OE no longer warns about the `/usr/src` directory being unpackaged when building the alternative -devsrc recipes.
* Manually installed the new `kernel-devsrc` package to a VM which has its soft-clock setback to 2000-01-01. Verified that 1) the `gcc` and `as` tools are now installed as dependencies, and 2) the compilation now succeeds without looping.

## Maintainership
* This PR should be cherry-picked to the hardknott mainline, once it is approved.

@ni/rtos


gratian (approved):
This looks good and I like the clean-up.

I would sleep a lot better if you could also test that installing one of the alternate kernels (for example `opkg install packagegroup-ni-debug-kernel`) still works and NI out of tree modules get versioned correctly: for example you can see `nikal` when running `lsmod` after booting the alternate kernel. This requires installing the NI drivers on the VM or using real hardware.

billpittman:
I'm fine with this change, but I wonder if we should have a warning output from "opkg update" if Packages.gz is newer than the system time.  In general, Packages.gz should always be "older" than the current system time, and if not output a warning message similar to what gcc/make does.

I'm just a bit concerned that this may happen to other IPK/postinst/prerm/etc scripts.  Maybe even promote the error all the way to MAX - but that's a UI element that I really don't want to get into.

billpittman (approved):

shruthi-ravi (approved):

chaitu236 (approved):

amstewart:
@gratian 
I booted a system with a 21.3 runmode and setup a custom feed with the new `kernel-devsrc-next` package. Installing the `packagegroup-ni-next-kernel` meta-package installed the kernel-devsrc-next package and the new kernel successfully. Installing the ni-kal driver after rebooting worked and rebuilt with the -next kernel headers.

I also tried installing the -next packagegroup on a 21.3 runmode with ni-kal already installed. It successfully rebuilt the driver and I could reload it.

gratian:
Thanks. LGTM. Ship it!

PR #249, opened 2021-10-14
Remove ti-wifi-utils as its ARM specific

chaitu236:
ti-wifi-utils is not compiling on hardknott.
But it's ARM specific so removing the recipe

@ni/rtos


shruthi-ravi (approved):
Can you add a note about the compilation issue to the commit message as well?

chaitu236:
> Can you add a note about the compilation issue to the commit message as well?

Done

amstewart:
If our intention is to not build it for any architecture, then lets also remove [the recipe](https://github.com/ni/meta-nilrt/tree/nilrt/master/sumo/recipes-connectivity/ti-wifi-utils) from the layer,  for cleanliness.

chaitu236:
> If our intention is to not build it for any architecture, then lets also remove [the recipe](https://github.com/ni/meta-nilrt/tree/nilrt/master/sumo/recipes-connectivity/ti-wifi-utils) from the layer, for cleanliness.

Done

amstewart (approved):

PR #255, opened 2021-10-21
procps: Use custom sysctl.conf file in recipe

billpittman:
Moved the nilrt sysctl.conf file from base-files-nilrt to procps
since the procps recipe wants to install the sysctl.conf file and
having both packages doing it was causing a build failure.


recipes-extended/procps/procps_3.%.bbappend
```diff
@@ -0,0 +1 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
```
chaitu236, line 1: Procps was present in sumo but was probably not getting pulled in. Do we know what pulls it in now and if we really need the package?

billpittman: No I don't know what pulls it in and didn't spend much time trying to figure it out since its one of the packages that I tend to like to have installed anyway.  Does Toaster reveal that info ?

chaitu236: image

Hmm I don't see reverse dependencies for this in toaster - wonder what that means. Let me see if I can dig into it.

chaitu236: Seems like kernel-dev is pulling in some dev packages that eventually pull in procps. Probably not worth figuring out what changed from sumo.

Do you think we should add an explicit dependency on procps from packagegroup-ni-base.bb so that pulling sysctl.conf is not left to chance in case something changes in future?

billpittman: That's probably not a bad idea at all, I'll add another commit to this PR.

chaitu236 (approved):

jeminor:
It seems a bit odd to me that this file is only consumed by procps, when that used to be an optional package and I see things in here that look to be not optional.  What else is using the content from here?

shruthi-ravi (approved):

recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb
```diff
@@ -81,10 +80,6 @@ do_install () {
 	# script for setting locale when a user logs in
 	install -m 0644 ${WORKDIR}/nisetlocale.sh ${D}${sysconfdir}/profile.d/
 
-	# add sysctl.conf file to adjust system configuration parameters
-	# FIXME: Add sysctl.conf after conflict with procps-sysctl is resolved
```
amstewart, line 85: Do you know what conflict this comment was referring to?

amstewart:
@billpittman What is the build failure that is occurring now? And what does NI use the sysctl.conf file for?

I'd prefer to setup an actual package dependency, over just adding it to the base packagegroup.

billpittman:
The build failure is a duplicate file install (sysctl.conf), both base-files-nilrt and procps want to install it.  But only 1 of them can, so this change just moves the sysctl.conf that we want to use into the procps package to avoid the duplicate file install.

Then, since procps is installing the sysctl.conf, I added it to the packagegroup to make sure it is specifically called out as a package to be installed and not just installed due to some downstream dependency eventually listing it.

recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb
billpittman: That was temporary - @chaitu236 removed sysctl.conf from base-files-nilrt just to get past that error.  The fixme was to go back and resolve it.  He made this temporary change in parallel to this change.

amstewart:
So that sounds like `base-files-nilrt` has a dependency on `procps` now; right? Can we just add that RDEPENDS and let bitbake pull `procps` into the core feed via `base-files-nilrt`, instead of adding it as a direct dependency of the packagegroup?

billpittman:
Alex I can make that change no problem.  Do me a favor and don't pull this branch, I accidentally pushed it too early.

billpittman:
@amstewart Here's the change you requested.

amstewart (approved):

PR #256, opened 2021-10-21
nilrt-image-common: Build images into tar.gz files

billpittman:
Added the tar.gz type so images will be built into tar.gz files
from bitbake.


billpittman:
Not needed - so deleted.

PR #257, opened 2021-10-21
Some changes for building compatible runmode. Part 2 of 2

chaitu236:
This contains remaining changes from Brenda's branch with some modifications.
Also moved CUSTOM_KERNEL_PATH into nilrt-dkms-image and replaced it in nilrt-safemode-image.

### Testing
Built nilrt_dkms_image and /boot directory structure of the image looks good (all runmode files end up in /boot/tmp/runmode). Image itself wasn't tested.

image

@ni/rtos

### DIFF
--- a/recipes-core/images/files/nilrt-runmode-system-image.cdf
+++ b/recipes-core/images/files/nilrt-runmode-system-image.cdf
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<?CDF VERSION="7.0"?>
+<DEFINITION>
+  <SOFTPKG NAME="%guid%" VERSION="%version%" OLDESTCOMPATIBLEVERSION="%version%" PROVIDESOS="YES" TYPE="HIDDEN">
+    <TITLE>SystemLink Base Image</TITLE>
+    <IMPLEMENTATION>
+      <OS VALUE="NI-Linux x64"><OSVERSION VALUE="7.0"/></OS>
+      <CODEBASE FILENAME="%filename%" TYPE="TAR" />
+    </IMPLEMENTATION>
+  </SOFTPKG>
+</DEFINITION>
--- a/recipes-core/images/files/nilrt-runmode-system-image.postinst
+++ b/recipes-core/images/files/nilrt-runmode-system-image.postinst
@@ -1,0 +1,107 @@
+#!/bin/bash
+#
+# Copyright (c) 2013,2017 National Instruments
+#
+
+# Something in base is installing /etc/natinst/share/uninstall that may not be
+# getting installed for SystemLink. nisid will fail if this directory doesn't
+# install, so this component ought to install /etc/natinst/share/uninstall
+# itself, regardless of if some other component ultimately will too.
+mkdir -p -m 755 /mnt/userfs/etc/natinst/share/uninstall
+
+# Enable ssh by default when installing SystemLink.
+/usr/local/natinst/bin/nirtcfg -s section=systemsettings,token=sshd.enabled,value=True
+
+# Add the memory reservation for Base. We assert that if the section exists,
+# it's already been added.
+if ! grep -q 'RtLinuxMemReserve' /mnt/userfs/etc/natinst/share/ni-rt.ini; then
+	cat >> /mnt/userfs/etc/natinst/share/ni-rt.ini <<EOF
+[RtLinuxMemReserve]
+Base=24
+EOF
+fi
+
+# Artemis introduced a new type of design which combined the use of zynq with an
+# ext4 filesystem drive as its main disk. Hence, some configurations are needed
+# for the systemlink software installation to be executed properly.
+
+# check for artemis compatibility from devicetree
+if grep -qs artemis /sys/firmware/devicetree/base/compatible ; then
+
+    # configure the partitions to mount by label in fstab
+    sed -i '/ubifs/d' /mnt/userfs/etc/fstab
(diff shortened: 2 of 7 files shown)


recipes-core/images/nilrt-safemode-image.bb
```diff
@@ -42,9 +40,9 @@ bootimg_fixup() {
 	# The kernel was installed with a symbolic link from 'bzImage'
 	# to the actual versioned file. Remove the redirection so that
 	# we just have a 'bzImage'
-	mv "$(realpath ${IMAGE_ROOTFS}/${CUSTOM_KERNEL_PATH}/bzImage)" "${IMAGE_ROOTFS}/${CUSTOM_KERNEL_PATH}/bzImage.real"
+	mv "$(realpath ${IMAGE_ROOTFS}/${KERNEL_IMAGEDEST}/bzImage)" "${IMAGE_ROOTFS}/${KERNEL_IMAGEDEST}/bzImage.real"
```
shruthi-ravi, line 43: What is the value of KERNEL_IMAGEDEST?

recipes-core/images/files/nilrt-runmode-system-image.postinst
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
```
shruthi-ravi, line 1: Is this source the same as in os-common or are there modifications?

recipes-core/images/files/nilrt-runmode-system-image.cdf
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<?CDF VERSION="7.0"?>
```
shruthi-ravi, line 2: I don't think this CDF is needed. Base image installers already have a CDF. @bstreiff can you comment?

recipes-core/images/nilrt-safemode-image.bb
chaitu236: Its "boot/runmode" (defined in [nilrt.conf](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/conf/distro/nilrt.conf#L20)).

recipes-core/images/files/nilrt-runmode-system-image.postinst
chaitu236: Aside from the [commit in this PR for modifying fstab entries](https://github.com/ni/meta-nilrt/pull/257/commits/46a4bf88a140ea7f6a20e486a520118d7c8f8d90), the only change is following

image

amstewart (approved):
Approved; assuming that the CDF is necessary.

recipes-core/images/files/nilrt-runmode-system-image.cdf
shruthi-ravi: I was wrong, base image installers pull the CDF from os-common. The CDF itself is no longer an issue. But the title and GUID are different from the os-common version. Downstream clients will have to update. Is this going to work ok with upgrades/downgrades? I'm ok if you want to follow this up in a follow-on PR if necessary.

shruthi-ravi (approved):

recipes-core/images/files/nilrt-runmode-system-image.cdf
bstreiff: os-common exports CDFs and they have to link up with the CDFs in the nilrt_onert_system_image component. The GUID that I used here is the one that matches up with `linux-rt-rss-featureToggle.cdf` which is different than the base-image GUID used for the non-featureToggle version.

Would it might be organizationally easier if the CDFs live in `rtos_nilinuxrt` so that they're all within the same git repo?

chaitu236: I don't have a strong opinion on where the CDF should live, we could make case for either - its makes sense to have all CDFs in ni-central but then the tar and CDF go together so it also makes sense to keep it here.
Does anyone want to argue for one or the other?

bstreiff (approved):
I'm okay with the CDF stuff going in here now even if we decide it should live elsewhere later.

recipes-core/images/files/nilrt-runmode-system-image.cdf
bstreiff: I believe @jeminor expressed an opinion of keeping CDF-related stuff out of the OE sphere of repos.

I'm not sure that it really matters; even if the OE artifacts include both the tar+CDF, you still need other CDFs (e.g. `linux-rt-rss-featureToggle.cdf`) before you have a complete set of things that is actually "installable" from MAX.

shruthi-ravi: How would downstream components in Azdo get the CDF? Do you imagine one of the rtos components fetching the file from this repo and exporting it? My preference is to host the file in Azdo then. Again, can be resolved later. I can merge this patch set for now.

bstreiff: Yeah, `rtos_nilinuxrt` would pull it from the OE build, same as it does the tar. [Sorta like this.](https://dev.azure.com/ni/DevCentral/_git/ni-central/commit/3c3f02f6b4f9d9ff97e5b83e83332e0827be1478?refName=refs%2Fheads%2Fusers%2Fbstreiff%2Fdunfell-safemode) That said, anything upstream would be pulling it out of an `rtos_nilinuxrt` export _anyway_, so it really doesn't matter if the CDF comes out of running bitbake or if it exists in the source tree local to that component, as long as the filename referenced in the CDF lines up.

PR #258, opened 2021-10-23
Move bootimage.ini into its own recipe

chaitu236:
Add bootimage.ini to nilrt-dkms-image

### Testing
Verified `nilrt-dkms-image` contains `bootimage.ini` in `/boot/tmp/runmode`
Verified `nilrt-runmode-system-image.tar` can be installed from MAX and we can boot into runmode

@ni/rtos

### DIFF
--- a/recipes-core/images/nilrt-dkms-image.bb
+++ b/recipes-core/images/nilrt-dkms-image.bb
@@ -1,5 +1,9 @@
 DESCRIPTION = "Runmode image for ethernet-based NI Linux Real-Time targets running XFCE (DKMS)."
 
+SRC_URI += "\
+	file://bootimage.ini \
+"
+
 IMAGE_INSTALL = "\
 	packagegroup-ni-runmode \
 	packagegroup-ni-wifi \
@@ -26,6 +30,9 @@ PACKAGE_EXCLUDE += "rauc rauc-mark-good"
 CUSTOM_KERNEL_PATH ?= "/boot/tmp/runmode"
 
 bootimg_fixup() {
+	install -m 0644 "${THISDIR}/files/bootimage.ini" "${IMAGE_ROOTFS}/boot/runmode/bootimage.ini"
+	sed -i "s/%component_version%/${BUILDNAME}/" "${IMAGE_ROOTFS}/boot/runmode/bootimage.ini"
+
 	# Postinst script is going to want this all in /boot/tmp/runmode
 	install -d `dirname "${IMAGE_ROOTFS}/${CUSTOM_KERNEL_PATH}"`
 	mv "${IMAGE_ROOTFS}/${KERNEL_IMAGEDEST}" "${IMAGE_ROOTFS}/${CUSTOM_KERNEL_PATH}"


recipes-ni/ni-bootimageini/ni-bootimageini_1.0.bb
```diff
@@ -0,0 +1,16 @@
...
+
+SRC_URI = "file://bootimage.ini"
+
+FILES_${PN} = "/boot/bootimage.ini"
+
+do_install () {
+    install -d ${D}/boot
+
+    install -m 0644 ${WORKDIR}/bootimage.ini ${D}/boot/bootimage.ini
+    sed -i "s/%component_version%/${BUILDNAME}/" ${D}/boot/bootimage.ini
```
amstewart: I'm not sure it makes sense for `bootimage.ini` to live in a package in the feed. That INI file's contents are trying to express pipeline-information about when the image was built; right? So what would it even mean to install it at runtime? Or upgrade it?

What's the motivation in moving it to its own package?

chaitu236: The motivation for moving it into its own package is that this is a file we want to install in both runmode and safemode and I want to avoid putting more things in images' do_install(). I think ideally nothing that can go in packages should go in do_install().

Yes, it isn't very useful to put it in a feed, but is there a way to make it a package but not make it part of a feed?

amstewart: >Yes, it isn't very useful to put it in a feed, but is there a way to make it a package but not make it part of a feed?

Not using our current tooling; but it would be possible to implement.

Could you paste the contents of what this INI file looks like typically? That might help to clear up my confusion.

bstreiff: > That INI file's contents are trying to express pipeline-information about when the image was built; right? So what would it even mean to install it at runtime? Or upgrade it?

FWIW, this was my rationale for keeping it with the image recipes.

(Then again we do end up with "pipeline information" by way of `BUILDNAME` in the `os-release` package, although maybe that's an argument for that not being in os-release.)

chaitu236: Looks like this in released system image.

> [BootImageDetails]
component_version=8.7.0f0
device_code=0x7755 0x774B 0x7735 0x774D 0x773B 0x7773 0x76CE 0x77AA 0x77AD 0x77B9 0x77BC 0x77BD 0x77DB 0x77DC 0x77DD 0x77E1 0x787E 0x787F 0x7828 0x7833 0x7835 0x7840 0x7841 0x7875 0x7878 0x7880 0x78C8 0x78C9 0x78CA 0x78CB 0x78CC 0x78CD 0x78CE 0x78CF 0x78E3 0x78E4 0x78E5 0x78E6 0x78E7 0x78E8 0x78E9 0x78EA 0x78EB 0x78EC 0x78B7 0x79D0 0x79CE 0x7A38 0x7A39 0x7A3A 0x7A3B 0x7A41 0x7A43 0x79DD 0x79DE 0x79DF 0x79E0 0x7A2A 0x7A2B

amstewart: So, I don't think the `component_version` information should be in a package in the feeds, for the reasons I stated above. And for similar reasons, I'm not really happy with that info being in the `os-release` either.

The `device_code` information seems like its more generic, but I guess that is added in another file?

chaitu236: `device_code` info is also in the [same file](https://github.com/ni/meta-nilrt/blob/0f9caed8f8c98539839bf5187659f670ff9de382/recipes-ni/ni-bootimageini/files/bootimage.ini). We replace `%component_version%` with the correct version during build.

> So, I don't think the component_version information should be in a package in the feeds, for the reasons I stated above. And for similar reasons, I'm not really happy with that info being in the os-release either.

I can revert this commit for now and add bootimage.in directly in nilrt-dkms-image.bb. If in future we manage to separate the two, we can revisit putting just `device_code` stuff in a package.

chaitu236: Ok made the change to add `bootimage.ini` directly to `nilrt-dkms-image`.

shruthi-ravi (approved):

amstewart (approved):

PR #260, opened 2021-10-28
Dev/hardknott/grub

billpittman:
This change allows us to use upstream grub instead of ni/grub.  Also removes some dated TPM code.  Fixes some naming issues and recipe issues that allow grub to build without warnings.

Also this has been tested on my 9032 hardware and the target still behaves wrt the CPLD operations.


recipes-bsp/grub/grub-nilrt.inc
```diff
@@ -1,18 +1,14 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/grub:"
 
-PV = "2.02+git${SRCPV}"
-SRCREV = "832e7862089a7cbc985ae09a17f6827de6c85bf8"
-
-SRC_URI = "\
-	${NILRT_GIT}/grub.git;branch=${GRUB_BRANCH} \
-	file://x86-64-Treat-R_X86_64_PLT32-as-R_X86_64_PC32.patch \
+SRC_URI += "\
+	file://cmd-test-Add-bitwise-AND-document-the-feature.patch \
```
amstewart, line 4: If any of these .patches are related to each other, we should express an ordering by prepending the file name like `0001_*`. Is that not the case?

recipes-bsp/grub/grub-efi_git.bbappend
```diff
@@ -2,13 +2,13 @@ require grub-nilrt.inc
...
                 minicmd iorw echo reboot terminfo loopback memdisk tar help serial \
-                ls search_fs_uuid udf btrfs ntfs reiserfs xfs lvm ata tpm measure \
+                ls search_fs_uuid udf btrfs ntfs reiserfs xfs lvm ata tpm \
                 regexp probe"
 
 # Downstream NI-branch code quality is not yet ready to build with -Werror
 CFLAGS_append += "-Wno-error"
 
-PACKAGES_prepend = "${PN}-nilrt"
+PACKAGES_prepend = "${PN}-nilrt "
```
amstewart, line 11: NIT: Should this line have been a part of the previous commit?

recipes-bsp/grub/grub/grub-safemode.cfg
```diff
@@ -1,9 +1,5 @@
 # NI Linux RT grub boot configuration script
```
amstewart, line 1: What's with all the changes in this file?

recipes-bsp/grub/grub/add-inbit-command-to-io-module.patch
```diff
@@ -0,0 +1,86 @@
+From 48de0b2faec205e874099e0afdc14ab70919e702 Mon Sep 17 00:00:00 2001
```
bstreiff, line 1: Do we need this one? The bitwise-and patch describes itself as replacing the functionality of "inbit".

recipes-bsp/grub/grub-nilrt.inc
billpittman: They are only related in the sense that they're all NI commits/patches.  They don't share files and they don't depend on one another.  That said I really don't care how they're named, want me to name them 0001-* 0002-* and 0003-* arbitrarily (which lets all admit will turn into alphabetical) ??

recipes-bsp/grub/grub-efi_git.bbappend
billpittman: True enough, but I also removed the measure component, and that moved the "file" into this commit.  Want me to split the two changes so they can go in appropriate commits ?

recipes-bsp/grub/grub/grub-safemode.cfg
billpittman: Removing the measure.mod from grub (the commit Haris added that we've now dropped), caused the measure command to be unknown, so I removed all the references to measure (since they're related to TPM).

recipes-bsp/grub/grub/add-inbit-command-to-io-module.patch
billpittman: Totally agree with you, however looking at the grub-safemode.cfg file, there are still calls to inbit if the cpld_test value is $e5 (not sure what $e5 means, probably the "old version" of the cpld/smbios handling code?!?!?).

recipes-bsp/grub/grub/grub-safemode.cfg
chaitu236: Why did we add `measure` before and why are we removing it now?

billpittman: measure was for TPM implementations - my understanding on that is two fold, one the API has changed, so most/some/all of our implementation will need to change, and two we never sold it, and nobody's asking for it, so its not is the mvp.

recipes-bsp/grub/grub-nilrt.inc
amstewart: Nope; if they're not related, then I don't see a need to enforce an order.

recipes-bsp/grub/grub-efi_git.bbappend
amstewart: I mean... I would like you to; yes. :)

amstewart (changes requested):
Looks good; pending testing on a physical system and the minor NIT above.

chaitu236 (approved):

billpittman:
Both fixed now.

amstewart (approved):

amstewart:
Do you approve, @bstreiff ?

bstreiff (approved):

PR #261, opened 2021-10-29
Dev/hardknott/hwclock

billpittman:
Enable hwclock implementation in busybox and move the hwclock-init tests to a new package (util-linux-nilrt-ptest).  The busybox implementation of hwclock costs less than 6k in the final executable, but allows the use of the hwclock.sh script and since both busybox and util-linux use update-alternatives, the util-linux implementation is symlink'd because util-linux uses a higher priority.  This also allows us to stop carrying a custom patch for hwclock-init in openembedded-core.


recipes-core/util-linux/util-linux-nilrt.bb
```diff
@@ -1,4 +1,7 @@
-FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+LICENSE = "MIT"
```
amstewart, line 1: Why not just install the ptests in a `util-linux-ptest` subpackage?

billpittman: I considered it, but there's already one for util-linux.  Thought this would be equally good and not cause conflicts with the existing ptests.  If there's a better way I'm all ears.

amstewart: Well, we could try to make them play nicely together. But it's not a big deal, so dropped.

amstewart (approved):
Lets give it a shot. :+1:

PR #262, opened 2021-11-02
Update git:// fetch GitHub URLs to https:// URLs

shruthi-ravi:
GitHub is deprecating the git:// fetch protocol and is currently
enforcing a brownout causing all git:// fetches to fail.
https://github.blog/2021-09-01-improving-git-protocol-security-github/

@ni/rtos


gratian (approved):
At some point I suspect we'll want to update the README.md with the new URIs too but it's not critical for this PR.

gratian:
Also we'll need a similar change for Sumo until we switch to Hardknott.

amstewart (approved):

PR #267, opened 2021-11-04
curl: Updated recipe to 7.% and updated patches to apply to 7.75 source

billpittman:
Most patches applied cleanly, those with high enough fuzz to cause a warning got updated and those that didn't apply at all got updated.

Patches and builds curl now with our NI specific patches.


amstewart:
I was hoping that we could drop some/all of these patches at this point. Did we evaluate if that is possible?

billpittman:
Looking at the patches, most of them are deemed Inappropriate for upstreaming, so as long as we need them, we'll have to carry them out of tree.  Most of those patches are Pharlap and/or nibuild specific.  The few patches that have been submitted upstream aren't yet accepted, so I think we're stuck with these for a while (at least as long as we're stuck with Pharlap and in some cases nibuild).

amstewart:
Well, we're not supporting Pharlap in new OE builds. So that's why I think we might be able to drop it.

Likewise, the `nicurl` component is going to pull its curl source from previous OE releases. So I'm not sure the nibuild patches are relevant either.

billpittman:
Ok then that changes things a bit.  I'll drop the Pharlap changes and the Windows specific changes then and see what's left.  I think there may still be a few (those 5 submitted, patches likely, but I'm not sure if Windows is the only one using those or not).

billpittman:
Redoing this PR without pharlap & nibuild patches.

PR #272, opened 2021-11-11
Use local safemode for safemode-restore-image

chaitu236:
Previously we used to get safemode from feeds to build
safemode-restore-image. Change this so we use locally built safemode.

Also added grub-efi and safemode-image to packagegroup-ni-restoremode.

### Testing
`bitbake packagegroup-ni-coreimagerepo`
`bitbake pacakge-index`
`bitbake safemode-restore-image`

Validated that created .wic file can be used to provision NI Linux RT on a VM and that the VM boots into safemode after.

We will probably want to move these recipes into recipes-core/images and do more cleanup later as described in [this work item](https://ni.visualstudio.com/DevCentral/_workitems/edit/1604743). For now I just wanted to build a restoremode image for the VM work.

@ni/rtos

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-restoremode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-restoremode.bb
@@ -36,10 +36,12 @@ RDEPENDS_${PN}_append_x64 = "\
     efivar              \
     fw-printenv         \
     grub                \
+    grub-efi            \
     grub-editenv        \
     eudev               \
     ni-smbios-helper    \
     nilrtdiskcrypt      \
+    safemode-image      \
     "
 
 RDEPENDS_${PN}_append_xilinx-zynqhf = "\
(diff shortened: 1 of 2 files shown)


recipes-core/packagegroups/packagegroup-ni-restoremode.bb
```diff
@@ -36,10 +36,12 @@ RDEPENDS_${PN}_append_x64 = "\
     efivar              \
     fw-printenv         \
     grub                \
+    grub-efi            \
```
amstewart, line 39: Why do we need both `grub` and `grub-efi` here? I was under the impression that `grub-efi` was a RAUC thing.

chaitu236: I believe we are using EFI for current gen also. Unpacked dist-nilrt-grub-gateway ipk from nickdanger and the iso in there has efi stuff.

image

chaitu236: As for whether we need `grub`, I'm not sure. I can add this as a task to figure out in the recovery image work item.

amstewart: I find it very confusing that `grub-efi` isn't just a package split of `grub`, but uses the same source. But OK.

amstewart (approved):

shruthi-ravi (approved):

PR #273, opened 2021-11-15
packagegroup-ni-restoremode: Remove safemode-image

chaitu236:
safemode-image depends on nilrt-safemode-image which we build
after package feeds are built. But if safemode-image is in
a packagegroup there is a circular dependency.

So removing it to fix this.

@ni/rtos


jeminor:
I might recommend doing this as a fixup for the previous commit so we can squash later.

chaitu236:
> I might recommend doing this as a fixup for the previous commit so we can squash later.

Done

jeminor (approved):
I'm good with this, but will leave any follow-up to @shruthi-ravi .

shruthi-ravi (approved):

PR #274, opened 2021-11-16
 sysvinit-inittab: remove custom start_getty script

billpittman:
The original consolegetty script is no longer necessary as it had
some ARM specific code in it that is now outdated.  The default
start_getty script seems to work without having to be modified.

NOTE: If you find this commit because you are having problems with
the serial port on your target, undoing this commit will put you
back to a time when we had combined ARM and x64 support for serial
ports.


chaitu236 (approved):

recipes-core/images/files/nilrt-runmode-system-image.postinst
```diff
@@ -63,9 +63,9 @@ fi
 # in normal operation, /boot better exist
 awk '{print $2;}' < /proc/mounts | grep -q /boot || exit 1
 
-# Remove OE getty lines for serial terminal and replace with consolegetty
-sed -i'' -e"s/\/s\?bin\/\(start_\)\?getty \(-L \)\?[0-9]\+ tty[OPS]\+[0-9]\+/\/etc\/init.d\/consolegetty tty2/g" /mnt/userfs/etc/inittab
-grep -q "/etc/init.d/consolegetty" /mnt/userfs/etc/inittab || exit 1
+# Edit OE getty lines for serial terminal to remove the speed and set the default console to tty2
+sed -i'' -e"s/\/s\?bin\/\(start_\)\?getty \(-L \)\?[0-9]\+ tty[OPS]\+[0-9]\+/\/bin\/start_getty tty2/g" /mnt/userfs/etc/inittab
+grep -q "/bin/start_getty tty2" /mnt/userfs/etc/inittab || exit 1
```
amstewart: Why `tty2`?

billpittman: Tried to track down the reasoning behind the tty2 change, but couldn't find anything describing the reasoning.  It was an old changelist from 2013 that I actually buddied, but it had nothing describing why the tty2 change was needed and its been that way ever since.

billpittman: Looking at the script, the tty2 value isn't even selected.  Ultimately ttyS0 is selected based on the output of /proc/consoles and the flags in that file.  So maybe tty2 is specific to a device with no serial output or just a "this shouldn't be unassigned" value.

billpittman:
Tested this change with QEMU and on a cRIO-9032.

amstewart (approved):
Thanks for deleting the part. :)

PR #276, opened 2021-11-17
safemode-restore-image: Move into recipes-core

chaitu236:
Moved safemode-restore-image into recipes-core/images.

Removed safemode-image recipe and merged its contents into
safemode-restore-image.

Removed `build-services.bbclass`

### Testing
`bitbake packagegroup-ni-coreimagerepo && bitbake package-index && bitbake safemode-restore-image` works.
Generated iso works on a VM.

@ni/rtos

PR #277, opened 2021-11-17
Add a hardknott recovery media image recipe

amstewart:
This PR adds an image recipe `nilrt-recovery-image`, which installs the modern, hardknott safemode image to unprovisioned targets. It also removes the historic and confusingly named `restore-` images and the deprecated, historic `lvcomms` and `minimal` recipes.

This PR replaces ni/meta-nilrt PR #276.

There is more work to do on the `images/` directory, but it is not as relevant to getting the recovery media pipeline online, so can be deferred to later.

# Testing

This recipe builds on my dev machine's nicentral repo. I did not bother checking the deprecated images, for obvious reasons.

@ni/rtos


chaitu236 (approved):
Looks good. Thanks!

shruthi-ravi (approved):

PR #281, opened 2021-11-19
Fix/Remove safemode ramdisk TODOs

shruthi-ravi:
#### Remove radeon firmware from safemode.
Previously, we removed the firmware during the image creation stage.
Instead, move the firmware package dependency from packagegroup-ni-base
to packagegroup-ni-runmode, so that the package is included only in
runmode images.
Also, remove TODO captured in AzDO WI 1742019.

-------

#### nilrt-safemode-ramdisk: Remove alsa uninstallation
alsa packages are not installed into safemode. So, the steps to
remove alsa packages from the image are unnecessary.

@ni/rtos

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-base.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-base.bb
@@ -21,7 +21,6 @@ ALL_DISTRO_ARM_PACKAGES = "\
 
 ALL_DISTRO_x64_PACKAGES = "\
 	linux-firmware-i915 \
-	linux-firmware-radeon \
 	dmidecode \
 	efivar \
 	fw-printenv \
--- a/recipes-core/images/nilrt-safemode-ramdisk.bb
+++ b/recipes-core/images/nilrt-safemode-ramdisk.bb
@@ -38,14 +38,9 @@ remove_alsa () {
 		alsa-utils-alsactl alsa-utils-alsamixer libasound2
 }
 
-# Radeon firmware is huge. Remove it from safemode and blacklist the module.
-# TODO: Maybe we should have a list of modules to install instead of getting
-#       the kernel-modules metapackage and then having to remove things
-#       individually?
-#       we probably don't need things like infiniband either...
+# Radeon firmware is huge and is not included in the safemode.
+# Blacklist the kernel module that gets automatically included.
 remove_radeon () {
-	opkg -o ${IMAGE_ROOTFS} -f ${IPKGCONF_TARGET} --force-depends remove \
-		linux-firmware-radeon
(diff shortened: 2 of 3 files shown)


recipes-core/packagegroups/packagegroup-ni-base.bb
```diff
@@ -21,7 +21,6 @@ ALL_DISTRO_ARM_PACKAGES = "\
 
 ALL_DISTRO_x64_PACKAGES = "\
 	linux-firmware-i915 \
```
gratian, line 23: I wonder if we should do this change for the i915 firmware too. It shouldn't be needed for the text console in safemode.

gratian (approved):

gratian:
We should figure out what pulls in alsa. We don't need sound in runmode either and it would be nice to remove it altogether.

recipes-core/packagegroups/packagegroup-ni-base.bb
shruthi-ravi: Yeah, I wondered about that especially because the i915 firmware adds 10Mb to the safemode (10Mb more than the sumo safemode). But I wasn't sure what feature the firmware enabled. What does it do?

gratian: It's the GPU firmware for i915 based controllers (cRIO-903x, 904x). I guess we should make sure that the safemode text console works w/o it if you connect a monitor - I expect it to continue working with the generic VGA driver.

shruthi-ravi: I've created a work [item](https://ni.visualstudio.com/DevCentral/_workitems/edit/1745937) to look into this.

bstreiff:
FWIW, when I was making the stab at it, I was loathe to yank alsa from `DISTRO_FEATURES` because I think that prevents all alsa-related packages from being used, and I wasn't sure that we should shut out sound entirely (a user could have audio over a DisplayPort channel, or attach a USB device).

I do think there's no use for it in safemode and also probably doesn't need to be in the runmode base images (e.g. if anything it should be an optional install)

gratian (approved):
I don't have a strong opinion on alsa. Based on Brenda's comment sounds like it would prevent building alsa packages for the feed so maybe we should leave it alone for runmode in case somebody somewhere wants to use it. Either way Ship It!

shruthi-ravi:
Some of the alsa packages are pulled into runmode by xfce packages. It's not easy to untangle them without messing with DISTRO_FEATURE. Based on Brenda's comment, I've reverted the change to remove alsa from DISTRO_FEATURE.

PR #284, opened 2021-12-07
Enable runmode to work on Hyper-V

chaitu236:
Hyper-V storage driver is required for runmode to be able to mount rootfs.
So build the drivers into initramfs and enable it if running on Hyper-V.

### Testing
Hyper-V VM provisioned with safemode with these changes has `enable_initramfs=1` in `/boot/grub/grubenv`.
Runmode with these changes can subsequently be installed on this VM and boots.

@ni/rtos

### DIFF
--- a/recipes-core/initrdscripts/files/ni_provisioning.safemode
+++ b/recipes-core/initrdscripts/files/ni_provisioning.safemode
@@ -181,12 +181,14 @@ has_tpm_chip ()
 	[ -e "/dev/tpm0" ]
 }
 
-check_for_tpm_chip()
+check_if_ramfs_required()
 {
 	# If use_ramfs was unspecified by user, set it based on
-	#  presence of TPM chip
+	#  presence of TPM chip or Hyper-V
 	if [ -z "$use_ramfs" ]; then
 		if has_tpm_chip; then
+			use_ramfs=true
+		elif lsmod | grep hv_storvsc &> /dev/null; then
 			use_ramfs=true
 		else
 			use_ramfs=false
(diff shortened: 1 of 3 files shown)


recipes-core/initrdscripts/files/ni_provisioning.safemode
```diff
@@ -181,13 +181,15 @@ has_tpm_chip ()
...
-check_for_tpm_chip()
+check_if_ramfs_required()
 {
 	# If use_ramfs was unspecified by user, set it based on
-	#  presence of TPM chip
+	#  presence of TPM chip or Hyper-V
 	if [ -z "$use_ramfs" ]; then
 		if has_tpm_chip; then
 			use_ramfs=true
+		elif lsmod | grep hv_storvsc &> /dev/null; then
```
billpittman, line 191: What loads the hv_storvsc driver in the provisioning disk?  Does the provisioning disk always use an initramfs?

chaitu236: I added the drivers in a [previous commit](https://github.com/ni/meta-nilrt/commit/bc80fdde2156ba33641eaf740e762d1c00c03eda). Yes, both the provisioning mode and safemode always use initramfs.

billpittman (approved):

recipes-core/initrdscripts/files/ni_provisioning.safemode
```diff
@@ -317,8 +319,8 @@ install_grubenv()
 	grub-editenv $BOOTFS_MOUNTPOINT/grub/grubenv set "BIOSBootMode=efi"
 	grub-editenv $BOOTFS_MOUNTPOINT/grub/grubenv set "NITarget=$NI_TARGET"
 
-	# If target has a TPM chip, we need to enable an alternate boot flow
-	#  involving initramfs and bootflags dir
+	# If target has a TPM chip or runs on Hyper-V, we need to enable an alternate boot flow
+	# involving initramfs and bootflags dir
```
amstewart, line 323: What is the benefit of *not* enabling `initramfs`?

If we enable it on every target which has a TPM, it seems like we would hit this branch with relative frequency anyway. Should we just enable it on all targets?

chaitu236: One benefit I can think of is slightly quicker boot time.
I'm not sure if the initramfs is unloaded after booting rootfs but if its not then some RAM savings also.

chaitu236: Oh and also RAM savings from not loading unnecessary drivers.

amstewart: OK. I suppose I don't have a hard argument for removing it; just looking for places to remove complexity.

amstewart (approved):

PR #285, opened 2021-12-07
Revert "salt: remove salt bbappends from meta-nilrt"

andzn:
Reverts ni/meta-nilrt#224

Based on discussion with @amstewart :
@andzn: "I wanted to change the salt branch in the meta-nilrt repo, but I didn't find the corresponding bbappend in the master/hardknott branch. I was looking for a file similar to this one: meta-nilrt/salt_2018.3.%.bbappend at f09f29a847134688036cdf03d92b7e6a995b8049 · andzn/meta-nilrt (github.com). The only one I found was this one in meta-cloudservices: meta-cloud-services/salt_3001.1.bb at nilrt/master/hardknott · ni/meta-cloud-services (github.com) but this one doesn't let us specify a git branch. Instead it seems to download packages from pythonhosted.org. Am I missing something?"

@amstewart : "I dropped the meta-nilrt salt bbappends during the hardknott rebase, because they conflicted with salt 3001, and we weren't sure at the time how or when we were going to address that. See the PR here. If you check the git history of that repo, it should be straightforward to revert the commits which removed the bbappends and apply your changes."

### DIFF
--- a/recipes-connectivity/salt/files/run-ptest
+++ b/recipes-connectivity/salt/files/run-ptest
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+# Kill any salt test processes which may be left around from orphaned test
+# runs.
+killall -q cli_master.py
+killall -q cli_syndic.py
+killall -q cli_minion.py
+killall -q cli_run.py
+
+python /usr/lib/python3.5/site-packages/salt-tests/tests/runtests.py \
+    --ptest-out \
+    --no-report \
+    --transport=tcp \
+    --run-destructive \
+    -n unit.modules.test_opkg \
+    -n integration.modules.test_system \
+    -n integration.modules.test_timezone \
+    -n integration.modules.test_shadow \
+    -n integration.states.test_user \
+    -n integration.modules.test_groupadd \
+    -n integration.modules.test_nilrt_ip \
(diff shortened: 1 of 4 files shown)


amstewart:
As it stands, this PR wouldn't do anything. The `salt_3000.bbappend` is namespaced to only apply to the [upstream salt recipe](https://github.com/ni/meta-cloud-services/blob/nilrt/master/hardknott/meta-openstack/recipes-support/salt/salt_3001.1.bb) if it is version `3000`, and it is evidently version `3001.1` in our branch of meta-cloud-services.

On second-review, I may have led you astray by linking you to PR 224. The previous *real* version of the salt.bbappend we had in the meta-nilrt layer was removed by @shruthi-ravi in meta-nilrt commit [2025c8cbbef427a94760c3686f19f1fb7193d44f](https://github.com/ni/meta-nilrt/commit/2025c8cbbef427a94760c3686f19f1fb7193d44f), and replaced (in the subsequent commit) by a stub bbappend to apply on salt_3000. Then I removed that stub later in PR 224.

So you will need to...
1. restore the `salt_2018.*.bbappend` that Shruthi removed in the above commit.
2. rename the bbappend to something like `salt_3001.%.bbappend`.
3. try to build the `salt` target in bitbake and reconcile incompatibilities between our bbappend and the upstream recipe.

amstewart:
I'll close this PR, since we have agreed to pursue a different salt recipe implementation for hardknott.

PR #290, opened 2021-12-11
[hardknott] Fix parsing failures in packagegroup-ni-extra

amstewart:
During the sumo-hardknott rebase, many recipes that were a part of the extra packagegroup broke when they were dropped upstream. Since we don't guarantee the extra feed contents (and build it with `--continue`) it wasn't a blocking issue. This PR tries to fix those errors and make the -extra packagegroup at least buildable.

1. Many recipes in packagegroup-ni-extra were removed, because they were dropped by yocto upstream and I couldn't find any justification for their inclusion in the feed. (NOTE: anywhere in the commits I say "no discrete requirement", I mean that they seemed to be included arbitrarily.)
2. The `florence` recipe was broken, and had previously been migrated to meta-nilrt after it was dropped by yocto upstream. Since then, we have identified other virtual keyboard solutions that we would prefer to use in hardknott. So I removed `florence` and `libglade`.
3. Added a static uid/gid for FreeRadius, which allows bitbake to actually build that package family. It is a low-level dependency for many packages in the `packagegroup-core-security` group, so that fixed many recipes from that layer.
4. `nirtcfg-tests` was broken due to a grammatical error in the recipe. I fixed it.

I have more -extra packagegroup recipe fixups I would like to put in (removing the `nilrt`/`nilrt-nxg` split, et c.), but they are unrelated to the goals of this patchset.

# Testing
I can now perform parsing operations like `bitbake -e` and `bitbake -g` on the -extra packagegroup. I haven't tried to build it on my dev machine, since we don't guarantee it like that. But I will monitor the extra feed build once this is accepted.

@ni/rtos


amstewart:
# Patch V2
* Clean rebase on latest hardknott mainline ref.

recipes-core/packagegroups/packagegroup-ni-extra.bb
```diff
@@ -279,7 +278,6 @@ RDEPENDS_${PN} += "\
 	umip \
 	usbmuxd \
 	zeromq \
-	wireless-tools \
```
shruthi-ravi, line 282: And is the idea that we would add upstream's alternate recommendations if someone requests it explicitly?

recipes-core/packagegroups/packagegroup-ni-extra.bb
```diff
@@ -590,7 +590,6 @@ RDEPENDS_${PN} += "\
 RDEPENDS_${PN} += "\
 	python3-future \
 	python3-gsocketpool \
-	python3-mprpc \
```
shruthi-ravi, line 604: Is the reference to `myrpc` in the commit message supposed to be `mprpc`?

shruthi-ravi (approved):

recipes-core/packagegroups/packagegroup-ni-extra.bb
amstewart: Yes; correct.

recipes-core/packagegroups/packagegroup-ni-extra.bb
amstewart: It is; good catch. Will fix.

amstewart:
# Patch V3
* `s/myrpc/mprpc/` in the `pkggrp-ni-extra: remove python3-mprpc ` commit. @shruthi-ravi

PR #292, opened 2021-12-16
packagegroups: Remove python3 from packagegroups

chaitu236:
### Description
packagegroups: Remove python3 from packagegroups

python and python3 packages no longer exist and anything depending on
them should instead depend on pythonX-* or pythonX-modules.

python3-modules "provides python3" so things work even without
this change but it is cleaner to switch to python3-modules.

So renamed python3 to python3-modules in runmode packagegroup.

Removed python3 from extra packagegroup since it is already in runmode.

### Testing
`bitbake packagegroup-ni-coreimagerepo && bitbake package-index && bitbake nilrt-runmode-system-image` works.
`packagegroup-ni-runmode.ipk` now depends on `python3-modules`. Did not test `packagegroup-ni-extra.ipk`

@ni/rtos

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-extra.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-extra.bb
@@ -148,7 +148,6 @@ RDEPENDS_${PN} += "\
 	python3-nose \
 	python3-numpy \
 	python3-pyiface \
-	python3 \
 	python3-setuptools \
 	rpm \
 	rsync \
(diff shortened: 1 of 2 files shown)


recipes-core/packagegroups/packagegroup-ni-extra.bb
```diff
@@ -148,7 +148,7 @@ RDEPENDS_${PN} += "\
 	python3-nose \
 	python3-numpy \
 	python3-pyiface \
-	python3 \
+	python3-modules \
```
amstewart: My understanding is that `python3-modules` is now providing the core python3 runtime and content. Since it's in the `-runmode` packagegroup, its contents will always end up in the core feed. So I think we can remove it from `-extra`.

amstewart (changes requested):

recipes-core/packagegroups/packagegroup-ni-extra.bb
chaitu236: Done

amstewart (approved):

PR #293, opened 2021-12-17
gdb: remove bbappend

amstewart:
The meta-nilrt gdb_%.bbappend did nothing but add `python` to the gdb
PACKAGECONFIG. However, since the meta-nilrt bbappend was authored,
upstream OE has improved the gdb bb recipe such that python is already
detected and enabled. The bbappend is currently superfluous.

Remove the bbappend entirely.

* [OE-core's equivalent line](https://github.com/ni/openembedded-core/blob/nilrt/master/hardknott/meta/recipes-devtools/gdb/gdb-common.inc#L35)

# Testing
Rebuilt `gdb` on my dev machine.

Old `PACKAGECONFIG`:
```
# $PACKAGECONFIG [12 operations]
<snip />
# pre-expansion value:
#   "readline python python"
PACKAGECONFIG="readline python python"
```

New `PACKAGECONFIG`:
```
# $PACKAGECONFIG [11 operations]
<snip />
# pre-expansion value:
#   "readline python"
PACKAGECONFIG="readline python"
```

@ni/rtos

### DIFF
--- a/recipes-devtools/gdb/gdb_%.bbappend
+++ b/recipes-devtools/gdb/gdb_%.bbappend
@@ -1,1 +1,0 @@
-PACKAGECONFIG_append = " python"


chaitu236:
It seems like `python` is enabled in the [meta-qt5's bbappend](https://github.com/ni/meta-qt5/blob/nilrt/master/hardknott/recipes-devtools/gdb/gdb_%25.bbappend#L1). I agree with this change - the bbappend in `meta-nilrt` can be removed.

> However, since the meta-nilrt bbappend was authored,
upstream OE has improved the gdb bb recipe such that python is already
detected and enabled.

I'm not sure about this comment though - is it really detected and enabled? Seems like its just blindly enabled in meta-qt5.

amstewart:
You're right; good catch. While reviewing the env output, I didn't notice that meta-qt5 was adding those values.

Since NILRT doesn't have a discrete requirement to support the GDB python extensions, I'll still merge this PR. If the other meta layers then decide to drop those PACKAGECONFIG values, and we decide that we have a particular reason to support the python extensions, we can trivially re-enable them in meta-nilrt.

PR #295, opened 2021-12-17
linux: staple kernel recipes to 21.5 release branch

bstreiff:
Service for the NILRT 21.5 release.

### DIFF
--- a/recipes-kernel/linux/linux-nilrt-next_git.bb
+++ b/recipes-kernel/linux/linux-nilrt-next_git.bb
@@ -1,5 +1,5 @@
 DESCRIPTION = "NILRT linux kernel next development build"
-NI_RELEASE_VERSION = "master"
+NI_RELEASE_VERSION = "21.5"
 LINUX_VERSION = "5.15"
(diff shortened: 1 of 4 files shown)


bstreiff:
and of course i miss an obvious typo in the commit message

recipes-kernel/linux/linux-nilrt-next_git.bb
```diff
@@ -1,10 +1,10 @@
 DESCRIPTION = "NILRT linux kernel next development build"
-NI_RELEASE_VERSION = "master"
+NI_RELEASE_VERSION = "21.5"
 LINUX_VERSION = "5.15"
```
bstreiff, line 3: hm, actually, this breaks because we don't have a nilrt/21.5/5.15; should I make that as a branch too? what is our policy for linux-next in release branches?

bstreiff: branches are cheap, just made one for nilrt/21.5/5.15, will update the srcrev here too.

amstewart (approved):

PR #298, opened 2022-01-05
libfmi: resolve broken build, update to 2.3

bstreiff:
Update the libfmi recipe to their 2.3 release version instead of the [last commit](https://github.com/svn2github/FMILibrary/commit/d49ed3ff2dabc6e17cc4a0c6f3fa6d2ae64a1683) from the [now-defunct svn2github](http://piotr.gabryjeluk.pl/blog:closing-svn2github) gateway, and also incorporate the [meta-ros2 layer's fix](https://github.com/ros/meta-ros/blob/master/meta-ros2/recipes-devtools/fmilibrary/fmilibrary_2.2.3.bb) for the expat-related build issues by using system expat.


recipes-devtools/libfmi/libfmi/0001-fmixml-use-system-expat-instead-of-building-own-with.patch
```diff
@@ -0,0 +1,126 @@
...
+From: Martin Jansa <martin.jansa@lge.com>
+Date: Tue, 9 Jun 2020 12:48:53 +0000
+Subject: [PATCH] fmixml: use system expat instead of building own with
+ ExternalProject
+
+it generates duplicate rules for ExpatEx/libexpat.a and newer ninja-1.9.0 fails because of that:
+NOTE: VERBOSE=1 cmake --build fmilibrary/2.0.3-r0/build --target all -- -j 20
+ninja: error: build.ninja:1652: multiple rules generate ExpatEx/libexpat.a [-w dupbuild=err]
+
+Upstream-Status: Pending
```
amstewart, line 11: This `.patch` was authored in June of 2020. Is it still actually `pending` upstream? Or can we bump our libfmi version and drop the .patch?

recipes-core/packagegroups/packagegroup-ni-extra.bb
```diff
@@ -78,6 +78,7 @@ RDEPENDS_${PN} += "\
 	bind \
 	cifs-utils \
 	libpcap \
+	libfmi \
```
amstewart: Since this package family is required by Veristand, could you move `libfmi` to `packagegroup-ni-coreimagrepo` and put it in an `RDEPENDS_${PN}_append` section with a comment declaring that it is a Veristand requirement, and tag Marcelo Izaguirre as the engineering contact.

recipes-core/packagegroups/packagegroup-ni-runmode.bb
```diff
@@ -20,6 +20,7 @@ RDEPENDS_${PN} = "\
 	glibc-gconv-cp936 \
 	glibc-gconv-iso8859-1 \
 	iproute2-tc \
+	libfmi-dev \
```
amstewart: Similarly, let's break this into its own block and add the same Veristand + Marcelo comment.

amstewart (changes requested):

recipes-devtools/libfmi/libfmi/0001-fmixml-use-system-expat-instead-of-building-own-with.patch
bstreiff: I don't see it in their `master` or `dev_3.0` trees and it still appears to be necessary as of 2.3 (the latest release; May 2021).

amstewart (approved):

PR #300, opened 2022-01-07
Hardknott packagegroup refactors and cleanup

amstewart:
This patchset tries to cleanup some of the `recipes-core/packagegroups` content by removing deprecated content (ARM, `nilrt-nxg`, et c.) and consolidating information. Recommend that you review-by-commit.

Major changes in this PR:
1. Renamed `packagegroup-ni-{coreimagerepo, extra}` to `packagefeed-ni-{core, extra}` to more clearly advertise their role as the top-level target for each feed.
2. Create a `packagegroup-ni-internal-deps` to house the open-source dependencies of NIFeeds (proprietary) content. This packagegroup is added to the `packagefeed-ni-core` to ensure that these dependencies always get build into the core feed.
    * All content that gets added to this recipe must come with a description of the product line which asserts the dependency and an engineering contact who can speak to the technical requirements for the dep.
    * I don't expect that this recipe is nearly complete as-is.
3. Removed ARM-only and `nilrt-nxg`-only content where I saw it. I probably didn't get all of it, but I think this is a good first pass.

In general, I'm trying to move away from packagegroup contents being declared by *layer*, and toward it being declared by *requirement*. These are the first steps in that direction.

# RFC
* Any objections to the name `packagegroup-ni-internal-deps`?

# Testing
We're having some trouble getting the new -extra feed pipeline online in AZDO, so I haven't been able to get full builds of the feeds with these changes yet. Once it is unblocked, I will diff the contents of the feeds with and without these changes. I expect there should be no substantial differences. But I wanted to get this patchset out for review in the mean time.

@ni/rtos


recipes-core/packagegroups/packagefeed-ni-extra.bb
```diff
@@ -65,6 +65,7 @@ RDEPENDS_${PN} = "\
 # meta-nilrt
 RDEPENDS_${PN} += "\
 	packagegroup-ni-debug-kernel \
+	packagegroup-ni-desirable \
```
shruthi-ravi, line 61: The build steps you've described above (--continue...) don't seem straightforward, and would need insider knowledge. For that reason, I don't like how this feed is structured. I don't have any immediate alternate suggestions either.

shruthi-ravi (approved):

recipes-core/packagegroups/packagegroup-ni-base.bb
```diff
@@ -15,94 +15,90 @@ inherit packagegroup
...
+	kmod \
 	libavahi-client \
 	libavahi-common \
 	libavahi-core \
 	libcap-bin\
 	libnss-mdns \
 	libpam \
 	librtpi \
+	libstdc++ \
+	linux-firmware-i915 \
```
gratian, line 66: Did we accidentally drop the radeon firmware in Hardknott? It was here in Sumo: https://github.com/ni/meta-nilrt/blob/e7eb865054449d1b104f19fc4a376da84217cd03/recipes-core/packagegroups/packagegroup-ni-base.bb#L25

amstewart: We intentionally moved it to the runmode packagegroup [here](https://github.com/ni/meta-nilrt/commit/dbc1d0ce84e33a22fbf67a2e9c0d164d1ad46435).

recipes-core/packagegroups/packagefeed-ni-extra.bb
amstewart: My intention was that an uninformed builder wouldn't *have* to know that we build one with `--continue` and not the other. That's an implementation detail that we care about, and if the builder wants to do it exactly as we do, they can just run the [`build.extra-feeds.sh`](https://github.com/ni/nilrt/blob/nilrt/master/hardknott/scripts/pipelines/build.extra-feeds.sh) script, and it will do it for you.

recipes-core/packagegroups/packagegroup-ni-base.bb
gratian: Ah, yes. Sorry, I forgot about that.

gratian: Having only the i915 firmware in there threw me off. Now I seem to recall commenting on the other commit you've linked that we should look into if we can make the i915 firmware runmode only too. I think the answer is yes but we need to do some testing to make sure.

gratian (approved):

recipes-core/packagegroups/packagegroup-ni-base.bb
shruthi-ravi: A follow-on work item for moving the i915 package was created [here](https://ni.visualstudio.com/DevCentral/_workitems/edit/1745937).

amstewart:
# Patch V2
* Rebase atop latest `nilrt/master/hardknott` ref.
* Remove the `move dkms into pkggrp-ni-base` commit and its revert, since they were mutually-defeating.
* In an attempt to make the extra feed build take slightly less time and be slightly more reliable, remove packages from layers on the verge of deprecation.
    * Remove the `openjdk-7` package (the only meta-java dependency, I think).
    * Remove the meta-mono packages.
    * Remove the meta-ivi packages.

shruthi-ravi (approved):

amstewart:
# Patch V2 Testing
Diffed the feed contents between the latest build of the hardknott extra feed (AZDO 1999812) and this patchset (AZDO 1999947).

### oe-extra/

```diff
# oe-extra/all/**
-- mozroot-certdata*  # expected
-- packagegroup-ni-mono-extra*  # expected

# core2-64/
-- fsharp*  # expected
-- gtk-sharp*  # expected
-- libgdiplus0  # OK
-- libgif7  # OK
-- liboil*  # expected
-- librhino-java  # OK
-- mono*  # expected
-- openjdk-8  # thought we were dropping openjdk-7... but OK?
-- rhino  # OK

# x64/
# no changes
```

### oe-core/
```diff
# all/
++ packagefeed-ni-core  # expected
-- packagegroup-ni-coreimagrepo  # expected
++ packagegroup-ni-internal-deps  # expected

# core2-64/
# no changes

# x64/
# no changes
```

All changes were expected with a couple exceptions.
1. Seems `openjdk-8` was being provided instead of `-7`; we want to drop it either way.
2. I was expecting to see `packagefeed-ni-extra` in the `extra/all` subfeed. But it now occurs to me that,  since we run the build with `--continue`, it is possible for the packagefeed's deps to fail, in which case, BB will not create an IPK for the feed itself. Good to know.

----

## Bonus
Here's a dirty grep of the extra packages which are still failing.
```
bundler-native-1.6.2-r0 do_compile
chef-zero-4.2.3-r0 do_package
clamav-0.104.0-r0 do_fetch
diff-lcs-1.2.5-r0 do_package
erlang-R16B03-1-r0 do_configure
erubis-2.7.0-r0 do_package
facter-2.5.0-r0 do_package
hashie-3.4.2-r0 do_package
hiera-1.3.4-r0 do_package
highline-1.7.3-r0 do_package
ipaddress-0.8.0-r0 do_package
json-1.8.5-r0 do_package
libest-3.2.0-r0 do_package
method-source-0.8.2-r0 do_package
mime-types-2.5-r0 do_package
mixlib-authentication-1.3.0-r0 do_package
mixlib-cli-1.5.0-r0 do_package
mixlib-config-2.2.1-r0 do_package
mixlib-log-1.4.1-r0 do_package
mixlib-shellout-2.1.0-r0 do_package
mkinitcpio-v23+gitAUTOINC+37df4fe219-r0 do_fetch
net-ssh-2.9.1-r0 do_package
net-ssh-gateway-1.2.0-r0 do_package
net-ssh-multi-1.2.1-r0 do_package
nirtcfg-tests-1.0-r0 do_install_ptest_base
ohai-8.5.0-r0 do_package
pry-0.10.1-r0 do_package
pry-native-0.10.1-r0 do_compile
python3-pyconnman-0.2.0-r0 do_patch
qpid-0.20-r0 do_configure
rack-1.6.3-r0 do_package
rest-client-2.0.0.rc1-r0 do_package
ruby-shadow-2.5.0-r0 do_install
slop-4.2.0-r0 do_package
slop-native-4.2.0-r0 do_compile
systemu-2.5.2-r0 do_package
trousers-0.3.14-r0 do_compile
uwsgi-2.0.14+gitAUTOINC+af44211739-r0 do_install
yajl-ruby-1.4.0-r0 do_package
yard-0.8.7.6-r0 do_package
yard-native-0.8.7.6-r0 do_compile
```

The only failure which sticks out to me is the `mkinitcpio do_fetch` error, which based on the log is probably a consequence of github temporarily disabling the `git://` protocol; not concerning just yet.

----

## Conclusion

I think this patchset is ready to merge.

PR #301, opened 2022-01-10
Add onboard to provide a virtual keyboard for touch panels

bstreiff:
As per AzDO#1639936, `florence` is deprecated and we'd like to use `onboard` as a replacement.

This patch set does several things:
- Configures `dconf` to provide a system-wide settings store (in addition to the user-only one that exists without configuration). This lets the `onboard` package push a setting to indicate to GNOME libraries that accessibility interfaces should be enabled.
- Configures onboard to start automatically with XFCE sessions.
- Configures onboard with a default setting to auto-show for text entry fields, but only when no USB keyboards are connected, e.g., the "I am only using a touch panel" case. This may require tweaking, since I don't have the recommended touch panel and I do not know if it enumerates as a keyboard.
- Adds a dependency on the "DejaVu Sans" font, so that we are guaranteed to have one font that supports the Unicode glyphs that Onboard wants to use for Tab, Shift, Return, Caps Lock, et al. This could be implemented as a set of `key-label-overrides` but 1) I did not wish to enumerate them all and 2) I didn't want to test to see how, for example, `Backspace` would look in the default "Compact" layout where it is a 1.0U key.
- Perhaps contentious: I didn't like the default "Classic Onboard" theming, so I am taking advantage of right-of-first-commit-set and adding an "NI" theme to be used as the default, which uses the corporate greens.
- Adds `onboard` to the `coreimagerepo` package group; this makes it optional and one must `opkg install` it, which is the same experience we had with `florence`

With this set of changes, I'm able to use the on-screen keyboard from a QEMU-based VM to type into a terminal and things appear to work as expected.

Abc page
123 page
more options page

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-xfce.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-xfce.bb
@@ -24,6 +24,7 @@ RDEPENDS_${PN} = "\
 	mousepad \
 	ttf-pt-sans \
 	xserver-xorg-udev-rules \
+	onboard \
 "
 RDEPENDS_${PN}_append_x64 += "\
 	xf86-video-ati \
--- a/recipes-gnome/dconf/dconf_%.bbappend
+++ b/recipes-gnome/dconf/dconf_%.bbappend
@@ -1,0 +1,17 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI += "file://user \
+"
+
+CONFFILES_${PN}_append := " ${sysconfdir}/dconf/profile/user"
+
+do_install_append () {
+	install -d ${D}${sysconfdir}/dconf/db/local.d
+	install -d ${D}${sysconfdir}/dconf/profile
+
+	install -m 644 ${WORKDIR}/user ${D}${sysconfdir}/dconf/profile/
+}
+
+pkg_postinst_${PN} () {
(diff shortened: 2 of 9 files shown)


recipes-core/packagegroups/packagegroup-ni-coreimagerepo.bb
```diff
@@ -34,6 +34,7 @@ RDEPENDS_${PN}_append_x64 = "\
 	nilrt-grub-runmode \
 	nilrt-grub-safemode \
 	${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'packagegroup-ni-xfce', '', d)} \
+	onboard \
```
amstewart: I know the naming isn't ideal, but could you put this dependency into [`packagegroup-ni-xfce`](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-core/packagegroups/packagegroup-ni-xfce.bb#L12) with the other graphical applications, rather than loose in the coreimagrepo.

As a part of my more ambitious packagegroup refactors in meta-nilrt #300, I'm trying to get all the loose packages out of what will be `packagefeed-ni-core`.

amstewart (changes requested):
Looks great; pending the packagegroup location request below.

amstewart (approved):

PR #302, opened 2022-01-11
Update the git fetch URL to use the https protocol

shruthi-ravi:
GitHub has deprecated the git fetch protocol causing all git://
fetches that use the git protocol to fail.
https://github.blog/2021-09-01-improving-git-protocol-security-github/

Verified that `bitbake -c fetch linux-nilrt` ran successfully.


jeminor (approved):

jeminor:
Thanks for the wireless one also.

PR #306, opened 2022-01-13
Add salt 3000.2 recipe

shruthi-ravi:
RTOS would like to upgrade the salt packages supported on NILRT to match the SystemLink host. This patchset upgrades the salt recipe to v3000.2 and is based on the salt v2018.3 recipe in sumo.

@andzn has tested that the resulting salt 3000 packages work as expected.
This patchset does not include the salt-ptest. There are issues getting those tests to run in hardknott that I am working through at this time.

AzDO [1571913](https://ni.visualstudio.com/DevCentral/_workitems/edit/1571913).

@ni/rtos

### DIFF
--- a/recipes-support/salt/files/set_python_location_hashbang.patch
+++ b/recipes-support/salt/files/set_python_location_hashbang.patch
@@ -1,0 +1,107 @@
+Upstream-Status: Pending
+
+# The Salt SysV scripts require that the process name of the salt
+# components have the form "salt-<component>".
+# The current python shebangs on the salt components scripts spwans
+# processes that are generically named python. Changed shebang so
+# process names will be identifiable by the init scripts.
+
+diff -Naur a/scripts/salt-api b/scripts/salt-api
+--- a/scripts/salt-api	2015-04-08 16:48:01.912294278 -0500
++++ b/scripts/salt-api	2015-04-08 16:49:03.336483297 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/python
+ 
+ # Import salt libs
+ from salt.scripts import salt_api
+diff -Naur a/scripts/salt-call b/scripts/salt-call
+--- a/scripts/salt-call	2015-04-08 16:48:01.912294278 -0500
++++ b/scripts/salt-call	2015-04-08 16:49:11.360507977 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/python
+ '''
+ Directly call a salt command in the modules, does not require a running salt
+ minion to run.
+diff -Naur a/scripts/salt-cloud b/scripts/salt-cloud
+--- a/scripts/salt-cloud	2015-04-08 16:48:01.912294278 -0500
++++ b/scripts/salt-cloud	2015-04-08 16:49:20.612536436 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/python
+ '''
+ Publish commands to the salt system from the command line on the master.
+ '''
+diff -Naur a/scripts/salt-cp b/scripts/salt-cp
+--- a/scripts/salt-cp	2015-04-08 16:48:01.912294278 -0500
++++ b/scripts/salt-cp	2015-04-08 16:49:30.132565723 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/python
+ '''
+ Publish commands to the salt system from the command line on the master.
+ '''
+diff -Naur a/scripts/salt-key b/scripts/salt-key
+--- a/scripts/salt-key	2015-04-08 16:48:01.912294278 -0500
++++ b/scripts/salt-key	2015-04-08 16:49:39.912595801 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/python
+ '''
+ Manage the authentication keys with salt-key
+ '''
+diff -Naur a/scripts/salt-master b/scripts/salt-master
+--- a/scripts/salt-master	2015-04-08 16:48:01.912294278 -0500
++++ b/scripts/salt-master	2015-04-08 16:49:50.224627508 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/python
+ '''
+ Start the salt-master
+ '''
+diff -Naur a/scripts/salt-minion b/scripts/salt-minion
+--- a/scripts/salt-minion	2015-04-08 16:48:01.912294278 -0500
++++ b/scripts/salt-minion	2015-04-08 16:49:57.808650832 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/python
+ '''
+ This script is used to kick off a salt minion daemon
+ '''
+diff -Naur a/scripts/salt-run b/scripts/salt-run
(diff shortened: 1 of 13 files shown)


recipes-support/salt/salt_3000.2.bb
```diff
@@ -0,0 +1,203 @@
...
+           file://salt-master \
+           file://master \
+           file://salt-syndic \
+           file://cloud \
+           file://roster \
+"
+
+SRCREV = "${AUTOREV}"
+
+S="${WORKDIR}/git"
```
amstewart: NIT: spaces before and after the `=`.

recipes-support/salt/salt_3000.2.bb
```diff
@@ -0,0 +1,203 @@
...
+DEPENDS = "\
+           python3-msgpack \
+           python3-pyyaml \
+           python3-jinja2 \
+           python3-markupsafe \
+"
+
+DEPENDS = "\
+           python3-distro-native \
+"
```
amstewart: Is there a reason that these are split? And that you're overwriting the depends instead of doing `DEPENDS_${PN} += ...`?

recipes-support/salt/salt_3000.2.bb
```diff
@@ -0,0 +1,203 @@
+HOMEPAGE = "http://saltstack.com/"
+SECTION = "admin"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=c996f5a78d858a52c894fa3f4bec68c1"
+DEPENDS = "\
```
amstewart, line 5: NIT: While we're authoring this file, could you update the tab indent-styling. We're converging on the hanging-tab style:
````
VARIABLE = "\
<tab>foobar \
"
```

recipes-support/salt/salt_3000.2.bb
```diff
@@ -55,6 +55,37 @@ PACKAGES += "\
...
+    python3-mmap \
+    python3-pyinotify \
+    python3-pyroute2 \
+    python3-pika \
+    python3-psutil \
+"
+
+# these dependencies should NOT be added to the salt bb recipe as they're added
+# here in the bbappend for the NIFeeds ni-skyline packages. In the future these
+# will be removed and skyline packages made to depend on them directly.
```
amstewart, line 71: Do the SL folks have any more context on this effort? Do we still need to add these?

recipes-support/salt/files/set_python_location_hashbang.patch
```diff
@@ -0,0 +1,107 @@
+Upstream-Status: Pending
+
+# The Salt SysV scripts require that the process name of the salt
+# components have the form "salt-<component>".
+# The current python shebangs on the salt components scripts spwans
+# processes that are generically named python. Changed shebang so
+# process names will be identifiable by the init scripts.
```
amstewart, line 7: Since we own the sysVinit scripts, can we just fixup the initscript's logic to correctly handle the salt python scripts? The shebang lines they're using is best-practice, and I would prefer a local modification to our initscript over applying a `.patch` to a dozen files.

amstewart (changes requested):
```
salt: Use version-agnostic variable to reference python dir

The files in the current Python 3.9 lib directory were not being
packaged correctly. This change fixes the build breakage.
```

Could you add some detail to this commit message about how it was being incorrectly packaged.

recipes-support/salt/salt_3000.2.bb
shruthi-ravi: I wanted to keep the native dependency separate, but did not realize I was overwriting the variable. Good catch, thanks.

recipes-support/salt/salt_3000.2.bb
shruthi-ravi: Yes, still needed. I asked them and they hadn't made the required changes to the skyline packages.

recipes-support/salt/files/set_python_location_hashbang.patch
shruthi-ravi: I'm not sure how we can match a python process in an init script, if not for the process name.
Also, I think a change like this would generate more churn than necessary at this time.

amstewart: I'm pretty sure `start-stop-daemon` could still match the process based on the `PIDFILE` alone, and the `--name` argument is just there to catch the case that the original process was killed and another process happened to startup with the same PID. That being said, I understand the complication with running a python script which uses `env` as an indirection layer and it seeting `/proc/$pid/comm` to `python` instead of the script name.

I still think there is a better way to handle this, but after searching it isn't obvious and I'm not willing to block these changes on it.

recipes-support/salt/salt_3000.2.bb
```diff
@@ -0,0 +1,203 @@
...
+
+inherit setuptools3 update-rc.d
+
+# Avoid a QA Warning triggered by the test package including a file
+# with a .a extension
+INSANE_SKIP_${PN}-tests += "staticdev"
+
+# Note ${PN}-tests must be before ${PN}-common in the PACKAGES variable
+# in order for ${PN}-tests to own the correct FILES.
+PACKAGES += "\
```
andzn, line 46: FYI: out of all those packages we only use (and test) salt-minion and salt-common, so if it helps you could get rid of the others.

andzn (approved):

recipes-support/salt/salt_3000.2.bb
shruthi-ravi: Can you clarify? Do you want a single tab instead of multiple? Do you want to replace spaces with a tab?

recipes-support/salt/salt_3000.2.bb
shruthi-ravi: Ok, I'll see how well I can clean up the remaining packages.

recipes-support/salt/salt_3000.2.bb
amstewart: Both or either; depending on which variable assignment we're talking about.

recipes-support/salt/salt_3000.2.bb
shruthi-ravi: I've looked into this some more, and I won't be able to remove the other packages cleanly. The files - salt-syndic, salt-cloud, etc.- currently owned by the other packages are installed by distutils. And if there are no corresponding salt packages in the recipe that claim ownership of these files, bitbake errors out during the build. So I think I will let the remaining packages be as is, instead of trying to remove these files from the installed directories.
@amstewart  thoughts?

recipes-support/salt/salt_3000.2.bb
shruthi-ravi: Ok. That did not do much for my confusion :). I took a pass at it. Take a look and lmk if you would like additional changes.

amstewart: You got it. Looks good.

recipes-support/salt/salt_3000.2.bb
amstewart: It will probably be more maintenance burden to remove the packages than to maintain them - since bitbake will normally complain if all of the files installed to the sysroot don't get packaged *somewhere*. So I would recommend you leave them in the recipe.

amstewart:
@shruthi-ravi Could you post some patch-rev notes with a summary of what changed in the force-push?

shruthi-ravi:
Patch v2

- Added more context to the `salt: Use version-agnostic variable to reference python dir` commit
- Fixed variable assignment spacing and variable indentation in the salt recipe
- Fixed bug where I was overwriting the DEPENDS with the native python-distro dependency

amstewart (approved):

PR #307, opened 2022-01-13
[Hardknott] Refactor and cleanup images/ directory

amstewart:
This patchset majorly refactors the images/ directory in an attempt to cleanup old and deprecated content. Recommend you review by commit. The `README` file has also been updated and is a good reference for the images/ content.

Major changes int this patchset:
* `*.inc` files have been moved into an `includes/` subdirectory.
* RAUC content has been moved into a `rauc/` subdirectory and marked as "deprecated".
* ARM content in the images recipes has been dropped.
    * some ARM content remains in the includes files, more cleanup there to come.
* `nikms` content has been deprecated, since all images are now DKMS.
* Renamed basically all of the modern image recipes to more descriptively advertise their purpose. The cpio.gz archives with `init` capabilities are now called `-initramfs.bb`, their encapsulating rootfses are now `-rootfs.bb`, and the additional CDF and packaging logic for the runmode is still `-runmode-system-image.bb`.

I have more changes in-work to consolidate and modernize the includes files. And more simple changes like making the recipe `DESCRIPTIONS` more consistent. But this patchset is a good first pass which shouldn't change the contents of any of the images. So I'm using it as an RFC on the new names and to test that I haven't messed anything up.

# Testing
Nothing in this patchset should substantively change the contents of any of the modern images.

All modern image targets build on my dev machine. I have an AZDO run churning as well to ensure that I don't break anything in there.

@ni/rtos


recipes-core/images/README
```diff
@@ -1,17 +1,29 @@
...
-i    | nilrt-initramfs.inc
- k   | nilrt-nikms.inc
-  x  | nilrt-xfce.inc
+c     | nilrt-image-common.inc  # common rules for all NILRT images
+cn    | niconsole-image.inc     # bootable rootfs w/ package management
+  i   | nilrt-initramfs.inc     # install niconsole-initramfs as INITRAMFS
+   x  | nilrt-xfce.inc          # install pkggrp-ni-xfce and enable X11
+    p | nilrt-proprietary.inc   # define NI-proprietary packages
+
 Images:
```
shruthi-ravi: Do you need a note here saying that the images are supported only on x64?

recipes-core/images/README
```diff
@@ -13,7 +13,7 @@ c     | * nilrt-runmode-initramfs.bb
 cn xpl| * nilrt-runmode-rootfs.bb
       | * nilrt-runmode-system-image.bb
 c   pl| * nilrt-safemode-initramfs.bb
-      | * nilrt-safemode-image.bb        # NILRT safemode (firmware) rootfs
```
shruthi-ravi: I'm confused by this renaming. How is nilrt-safemode-image different from nilrt-runmode-system-image? They are both images, right? I would like to keep the image part of a name to indicate that's it's the top-level recipe I need to be building.

shruthi-ravi (approved):
My eyes started glossing over after about 10 commits. This is a lot of refactoring for one PR :).

recipes-core/images/README
amstewart: >My eyes started glossing over after about 10 commits. This is a lot of refactoring for one PR :).

If it makes you feel any better, my brain almost melted trying to figure out what was what.

>Do you need a note here saying that the images are supported only on x64?

I feel like that is almost an implementation detail; especially for the hardknott mainline - which only supports the one architecture.

recipes-core/images/README
amstewart: The `nilrt-safemode-image` (now `nilrt-safemode-rootfs`) is the safemode initramfs (`ramdisk.xz` on disk?), plus the non-volatile rootfs files which are shared between safemode and runmode (eg. the /etc/ files which are shared between the mounts.)

`nilrt-runmode-system-image` is an archive of the `nilrt-runmode-rootfs`, plus the runmode post-installation script and CDF metadata necessary to make a single payload that MAX can install.

>I would like to keep the image part of a name to indicate that's it's the top-level recipe I need to be building.

I considered that, but preferred to make it clear what kind of image it was instead. And I'm not sure its meaningful to say that some of the images are "top-level" versus other ones. The `nilrt-safemode-rootfs` is what is used to reformat the "firmware" image through MAX, but it is only a payload in the context of provisioning - where the top level recipe is `nilrt-recovery-image`. Likewise, both `nilrt-runmode-rootfs` and `nilrt-runmode-system-image` are viable standalone packages, just for different purposes.

shruthi-ravi: What I meant by "top-level" is that both of those packages - `nilrt-safemode-image` and `nilrt-runmode-system-image` can be used as is to install a functional safemode and runmode on a target. I understand that the `nilrt-recovery-image` is an additional means to install the safemode, but because the `nilrt-safemode-image` can also be used to install a complete image, I would prefer to keep the `-image` part of the name.

amstewart:
# Testing

Built the new image recipes from this patchset using [AZDO PR 253926](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/253926) and verified that the image artifacts haven't changed in an unexpected way.

```
## x64 Safemode Rootfs (`standard_x64_safemode.tar.gz`)
* Only differs in metadata (build export number)  # expected

## x64 Safmode Initramfs (`stnadard_x64_safemode.tar.gz/ramdisk.xz`)
* Used `libnitargetcfg_21.8` instead of `libnitargetcfg_21.5` due to mundane progression in the nifeeds, unrelated to this patchset  # OK

## x64 Runmode Rootfs (`nilrt-runmode-system-image-x64.tar`)
* /postinst does not differ  # expected
* /data.tar.gz
  * `libnitargetcfg` upgraded (see above)  # OK
  * fontconfig generated with new hashes (timestamp-based, I think)  # OK

## x64 Runmode Initramfs (`nilrt-runmode-system-image-x64/boot/tmp/runmode/runmode_initramfs.gz`)
* identical  # expected

## x64 recovery media (`nilrt-recovery-image-x64.iso`)
* identical  # expected
```

## Conclusion
* All differences are expected or OK. This patchset should be good-to-go.

recipes-core/images/README
amstewart: Let's talk offline at the knowledge sharing today.

amstewart: We reviewed this decision as a team and decided not to add the `-image` suffix.

amstewart:
# Patch V2
* Rebased on `nilrt/master/hardknott`. No substantive changes to the patchset.

PR #308, opened 2022-01-14
openssh: Modify sysvinit script to close connections

billpittman:
Modify the init script to close all open ssh connections when the
target reboots or shuts down.

### Testing

Tested locally and when the target reboots or halts, the ssh connection is closed.

### DIFF
--- a/recipes-connectivity/openssh/openssh_%.bbappend
+++ b/recipes-connectivity/openssh/openssh_%.bbappend
@@ -1,5 +1,20 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
 
+SRC_URI += " \
+	file://close-all-ssh-connections-patch \
+"
+
+addtask patch_oe_source after do_patch before do_configure
+
+do_patch_oe_source () {
+    # Patch sysvinit file to close all open ssh connections on shutdown or reboot
+    #
+    # The typical do_patch logic will not work for this file since the do_patch logic is designed to
+    # patch the source code for openssh, and the file being patched is part of the recipe that builds
+    # the openssh IPK.
+    patch -u ${WORKDIR}/init -i ${WORKDIR}/close-all-ssh-connections-patch
+}
+
 do_install_append () {
     # this logic is only for nilrt and nilrt-xfce, not for nilrt-nxg
     if ${@oe.utils.conditional('DISTRO', 'nilrt-nxg', 'false', 'true', d)}; then
--- a/recipes-connectivity/openssh/files/close-all-ssh-connections-patch
+++ b/recipes-connectivity/openssh/files/close-all-ssh-connections-patch
@@ -1,0 +1,10 @@
+--- a/init
++++ b/init
+@@ -53,6 +53,7 @@
+   stop)
+ 	echo -n "Stopping OpenBSD Secure Shell server: sshd"
+ 	start-stop-daemon -K -p $PIDFILE -x /usr/sbin/sshd
++	$(readlink -q $0 > /dev/null 2>&1) && killall sshd
+ 	echo "."
+ 	;;
+ 


recipes-connectivity/openssh/files/init
```diff
@@ -0,0 +1,91 @@
...
+	echo "Starting OpenBSD Secure Shell server: sshd"
+	@LIBEXECDIR@/sshd_check_keys
+	check_privsep_dir
+	start-stop-daemon -S -p $PIDFILE -x /usr/sbin/sshd -- $SSHD_OPTS
+	echo "done."
+	;;
+  stop)
+	echo -n "Stopping OpenBSD Secure Shell server: sshd"
+	start-stop-daemon -K -p $PIDFILE -x /usr/sbin/sshd
+	$(readlink -q $0 > /dev/null 2>&1) && killall sshd
```
billpittman: Here's the only line I added (Line 56 - readlink etc).  The rest of the script is from the original recipe from the openembedded-core layer.

amstewart: Since this is a one-liner, I think I would prefer for it to live as a `.patch` in meta-nilrt which applies overtop the initscript from oe-core.

amstewart (changes requested):

amstewart:
I grok the `killall`, but what's the purpose of the `readlink` part of the logic?

billpittman:
The readlink part of the logic verifies/checks that /etc/rc6.d/K09sshd (or /etc/rc0.d/K09sshd) is being called instead of /etc/init.d/sshd - catches a "shutdown" or "reboot", but not the developer restarting ssh.

billpittman:
WRT a patch file, ME TOO, but the specific file is part of the recipe, not part of the source, and I couldn't get/create a patch file that would patch the "recipe".  Does that make sense ?  The only other thing I could do would be to edit the bbappend file and do an do_install_append() function and "sed" the change in - but I thought that wasn't worth the effort/explanation.  Thoughts ?

amstewart:
> The readlink part of the logic verifies/checks that /etc/rc6.d/K09sshd (or /etc/rc0.d/K09sshd) is being called instead of /etc/init.d/sshd - catches a "shutdown" or "reboot", but not the developer restarting ssh.

I would think that we would want the openssh connections to be closed, even in the case of a developer-induced stop; wouldn't we?

amstewart:
> WRT a patch file, ME TOO, but the specific file is part of the recipe, not part of the source, and I couldn't get/create a patch file that would patch the "recipe". Does that make sense ? The only other thing I could do would be to edit the bbappend file and do an do_install_append() function and "sed" the change in - but I thought that wasn't worth the effort/explanation. Thoughts ?

Eugh. I don't like the idea that we will be overwriting all upstream changes to that init script when we're just trying to add one line.

I think you must be able to apply a `.patch` to OE-source files, but you may not be able to use the default `do_patch`. What about creating a `.patch` file for this change (in meta-nilrt) and then writing a little function in our bbappend to apply that .patch to OE-upstream's init file in the root of the recipe's work directory. You should be able to attach that function as a task to the do_patch step for the recipe, and that should run between do_unpack and do_compile. Can you give that a shot and see if it is feasible?

billpittman:
> > WRT a patch file, ME TOO, but the specific file is part of the recipe, not part of the source, and I couldn't get/create a patch file that would patch the "recipe". Does that make sense ? The only other thing I could do would be to edit the bbappend file and do an do_install_append() function and "sed" the change in - but I thought that wasn't worth the effort/explanation. Thoughts ?
> 
> Eugh. I don't like the idea that we will be overwriting all upstream changes to that init script when we're just trying to add one line.
> 
> I think you must be able to apply a `.patch` to OE-source files, but you may not be able to use the default `do_patch`. What about creating a `.patch` file for this change (in meta-nilrt) and then writing a little function in our bbappend to apply that .patch to OE-upstream's init file in the root of the recipe's work directory. You should be able to attach that function as a task to the do_patch step for the recipe, and that should run between do_unpack and do_compile. Can you give that a shot and see if it is feasible?

I'll give it a shot.

billpittman:
> > The readlink part of the logic verifies/checks that /etc/rc6.d/K09sshd (or /etc/rc0.d/K09sshd) is being called instead of /etc/init.d/sshd - catches a "shutdown" or "reboot", but not the developer restarting ssh.
> 
> I would think that we would want the openssh connections to be closed, even in the case of a developer-induced stop; wouldn't we?

Sort of, as a developer I can allways "killall sshd", but its more typical to not do that when ssh is simply restarted.  So generally speaking it is desirable that all existing ssh connections are maintained and undisturbed even if the main sshd daemon is restarted.  However its also desirable to have them disconnect on shutdown and restart.

billpittman:
Ok have a look at the new changes.  It wasn't exactly as straightforward as I hoped, but it looks a whole lot better and still uses a patch file during do_patch_append.

It sort of indirectly does it in the bb.build.exec_func call -- without which, the ${<VAR>} wouldn't expand properly.  Saw this pattern several times in other recipes that used do_patch_append.

recipes-connectivity/openssh/openssh_%.bbappend
```diff
@@ -1,4 +1,16 @@
...
+
+do_patch_append() {
+    bb.build.exec_func('do_close_all_ssh_sessions', d)
+}
+
+do_close_all_ssh_sessions () {
+    # Patch sysvinit file to close all open ssh connections on shutdown or reboot
+    #
+    # perform the patch operation on the version of init in the WORKDIR, not the original recipe
+    patch -u ${WORKDIR}/init -i ${SAVED_DIR}/files/close-all-ssh-connections.patch
```
amstewart: Wait a moment... :thinking:  Do we even need a custom step here? If we're just trying to patch the `/init` file in the source root, after it has already been written by the oe-core layer, then I think a regular `.patch` file should work using the normal bitbake `do_patch` mechanism. I may have led you into reimplementing logic that already exists.

What happens if you replace all these `.bbappend` changes with `SRC_URI += "close-all-ssh-connections.patch"` ? Does bitbake just do the correct thing?

amstewart (changes requested):
I like this new `.patch` a lot better. Let's try to make this work.

recipes-connectivity/openssh/openssh_%.bbappend
billpittman: No you didn't.  The init file is part of the RECIPE, not part of the source, the default bitbake logic for do_patch operations only applies to the SOURCE.  Its a super fine line, but one that I haven't been able to convince bitbake to cross.

Essentially it would require a patch that executes during do_patch and operates on files in the SOURCE/../init file -- and that won't work.  

I really think this is the right way to go, not as straightforward as a simple .patch file, but it gets the job done and allows for future distro upgrades.

amstewart: Oh; I see. The oe-core `init` script is installed directly into the rootfs. I was thinking that it was a `.patch` and that we could just order our `.patch` after theirs.

In that case, I agree that this is the best approach.

recipes-connectivity/openssh/openssh_%.bbappend
```diff
@@ -1,4 +1,16 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+SAVED_DIR := "${THISDIR}"
```
amstewart: Doesn't the `close-all-ssh-connections.path` file get copied into the `WORKDIR` during the `do_unpack`? Or do you need to add it explicitly to our bbappends `SRC_URI`?

Either way bitbake should copy it into the workspace for you, and you shouldn't have to apply it directly from the meta-nilrt source.

recipes-connectivity/openssh/openssh_%.bbappend
```diff
@@ -1,4 +1,16 @@
...
+do_patch_append() {
+    bb.build.exec_func('do_close_all_ssh_sessions', d)
+}
+
+do_close_all_ssh_sessions () {
+    # Patch sysvinit file to close all open ssh connections on shutdown or reboot
+    #
+    # perform the patch operation on the version of init in the WORKDIR, not the original recipe
+    patch -u ${WORKDIR}/init -i ${SAVED_DIR}/files/close-all-ssh-connections.patch
+}
```
amstewart: What about...
```
do_patch_oe_source () {
    ...
    patch -u ...
}

addtask do_patch_oe_source before do_patch
```

So that it lives as its own task.

amstewart (changes requested):

recipes-connectivity/openssh/openssh_%.bbappend
billpittman: > Doesn't the `close-all-ssh-connections.path` file get copied into the `WORKDIR` during the `do_unpack`? Or do you need to add it explicitly to our bbappends `SRC_URI`?

Right, unless you specifically add it to the SRC_URI variable, the file is essentially "not in the recipe", so it doesn't get copied to the WORKDIR.

recipes-connectivity/openssh/openssh_%.bbappend
billpittman: Ok a bit closer to what you'd ideally like, the patch file still gets sourced from the original location in the recipe since it can't be added to SRC_URI, but now there's a new task after patch do_patch_oe_source, and its cleaner than the do_patch_append implementation.

recipes-connectivity/openssh/openssh_%.bbappend
amstewart: Why can't you add the `.patch` to the `SRC_URI` in the bbappend?

billpittman: Because --- when I try, this happens ....


ERROR: openssh-8.5p1-r0 do_patch: Command Error: 'quilt --quiltrc /ssd/hardknott/nilrt/build/tmp-glibc/work/core2-64-nilrt-linux/openssh/8.5p1-r0/recipe-sysroot-native/etc/quiltrc push' exited with 0  Output:
stdout: Applying patch close-all-ssh-connections.patch
can't find file to patch at input line 3
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|--- a/init
|+++ b/init
--------------------------
No file to patch.  Skipping patch.
1 out of 1 hunk ignored
Patch close-all-ssh-connections.patch does not apply (enforce with -f)

stderr: 
ERROR: Logfile of failure stored in: /ssd/hardknott/nilrt/build/tmp-glibc/work/core2-64-nilrt-linux/openssh/8.5p1-r0/temp/log.do_patch.584
ERROR: Task (/ssd/hardknott/nilrt/sources/openembedded-core/meta/recipes-connectivity/openssh/openssh_8.5p1.bb:do_patch) failed with exit code '1'


I suppose I could try to rename the file to something that doesn't end in .patch that do_patch will ignore, but then I might start getting errors about having a file in SRC_URI that wasn't "installed" anywhere.  If you want, I can try that after the Dr. Appointment today.

billpittman: Ok that last idea worked.  I think you've got everything you asked for in this version!  -- And no that wasn't a challenge to find more things wrong.

amstewart (approved):

PR #311, opened 2022-01-19
kernel-devsrc-nilrt: make olddefconfig before make prepare

bstreiff:
Sometimes when we bump minor kernel versions, new options can show up in the kernel config; a plain `make prepare` will then prompt for values for those options. This is a problem when postinsts are being run on first boot, because there's no way to see nor respond to such prompts-- a `make olddefconfig` beforehand will set these to defaults and avoid user interaction.

### DIFF
--- a/recipes-kernel/linux/kernel-devsrc-nilrt.inc
+++ b/recipes-kernel/linux/kernel-devsrc-nilrt.inc
@@ -4,6 +4,7 @@
 pkg_postinst_ontarget_${PN} () {
 	cd /lib/modules/${KERNEL_VERSION}/build
 	find ./ -exec touch -m {} \;
+	make olddefconfig
 	make prepare
 	make modules_prepare
 


bstreiff:
In this particular instance it appears that the configuration options in question are compiler-related:
image

amstewart (approved):
LGTM.

For reference, can you comment on approximately how much time this adds to the postinst? Is `make olddefconfig` an intensive operation?

bstreiff:
Timing method: I installed the runmode, but instead of letting the reboot into runmode happen, I forced it into safemode, and then ran:
```
# chroot /mnt/userfs
# time /etc/init.d/run-postinsts
```
With d116, I get: real 0m32.222s, user 0m23.390s, sys 0m3.505s
With d132 (plus this postinst change), I get: real 0m33.035s, user 0m24.202s, 0m3.444s

Also note that there are a few more other packages (dconf, onboard, etc) between those two versions, so a few more postinsts exist.

gratian (approved):

chaitu236 (approved):

PR #312, opened 2022-01-20
syslog-ng: rename config file to match upstream and bump config file version

ni-jxie:
This patchset update our syslog-ng config file to work correctly on Hardknott.

**Testing:**
- Built packagegroup-ni-coreimagerepo locally.
- Installed built syslog-ng package on a cRIO-9049 and verified that `/var/log` folder has the right content. Checked that no error message related to syslog initialization in any of the logs.

@ni/rtos


gratian (approved):

amstewart (changes requested):
Changes seem fine to me.

Could you take the justifications from your PR message and insert them into your commit messages? We use the commit as the atomic unit of change in the meta-nilrt branch - not PRs. For example, see my commits in [meta-nilrt #300](https://github.com/ni/meta-nilrt/pull/300/commits).

shruthi-ravi:
Can you elaborate on why you had to do the version bump from within the config file? Do you foresee that needing to be bumped every time the recipe version is bumped? If so, maybe you can make the recipe write the version to the file.

ni-jxie:
@shruthi-ravi

The warning message on starting up is this
`Starting syslog-ng[2022-01-20T17:44:47.120773] WARNING: Configuration file format is too old, syslog-ng is running in compatibility mode. Please update it to use the syslog-ng 3.31 format at your time of convenience. To upgrade the configuration, please review the warnings about incompatible changes printed by syslog-ng, and once completed change the @version header at the top of the configuration file; config-version='3.8'` It is actually harmless.

I looked at the syslog-ng source, and apparently this is by design. If syslog-ng sees a config file with an older version declared, it will start in compatibility mode so it will not fail to start even if the older config file has deprecated config options. It will warn about those incompatible changes, and the maintainer of the config is supposed to review the log to see if there is any warning and fix that in their config file before bumping the version. If we don't want the warning message on syslog-ng, then we would have to bump our config version every single time when syslog-ng bumps its version.

Currently since we are manually bumping the version, it forces us to check and verify there is no incompatibilities. If we bump the version automatically, then syslog-ng would fail to start if our config file has incompatible changes, which would be a very visible breakage assuming we test for it. If we have an automated test to verify syslog-ng is starting correctly, then maybe automatically updating is a good idea.

ni-jxie:
@amstewart The commit messages were amended, but somehow the order of the commits changed though. But it should not matter for this change.

Since commit message is our atomic unit of change, for the PR description, I am assuming we just put some high-level description? Unless the patchset is really complicated, we don't need to spell out what was changed in the PR description again?

gratian:
@ni-jxie this is a bit of a nitpick but you might consider wrapping the commit message at 75 columns. 

This a kernel commit message rule but we've been using similar rules for OE. I don't think we've been very strict about it but it does make going to through the logs easier in a console.

ni-jxie:
@gratian Done, thanks for pointing that out, good to know.

amstewart (approved):

PR #314, opened 2022-01-21
salt: Add salt-ptest package

shruthi-ravi:
Add the salt-ptest package from the sumo salt bbappend recipe.
The goal of this change is to get the tests running. They do not all
pass at this time.
Changes from the sumo recipe:
- The --ptest-out option no longer exists in runtests.py and has been
removed from the test runner command in run-ptest.
- python3-mock and python3-distro have been added as dependencies to
the ptest package, without which the tests do not run.

@ni/rtos

### DIFF
--- a/recipes-support/salt/files/run-ptest
+++ b/recipes-support/salt/files/run-ptest
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+# Kill any salt test processes which may be left around from orphaned test
+# runs.
+killall -q cli_master.py
+killall -q cli_syndic.py
+killall -q cli_minion.py
+killall -q cli_run.py
+
+python -m salt-tests.tests.runtests \
+    --no-report \
+    --transport=tcp \
+    --run-destructive \
+    -n unit.modules.test_opkg \
+    -n integration.modules.test_system \
+    -n integration.modules.test_timezone \
+    -n integration.modules.test_shadow \
+    -n integration.states.test_user \
+    -n integration.modules.test_groupadd \
+    -n integration.modules.test_nilrt_ip \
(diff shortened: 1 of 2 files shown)


recipes-support/salt/files/run-ptest
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+# Kill any salt test processes which may be left around from orphaned test
+# runs.
+killall -q cli_master.py
+killall -q cli_syndic.py
+killall -q cli_minion.py
+killall -q cli_run.py
+
+python PYTHON_SITEPACKAGES_DIR/salt-tests/tests/runtests.py \
```
amstewart: Does python let you do something like `python -m salt-tests.tests.runtests` and do the site-package locating for you?

shruthi-ravi: Yes, that works. I'll replace this line with that.

shruthi-ravi:
Update v2: Modified run-ptest to make Python locate and run salt ptests instead of invoking the tests with the full path.
@amstewart

recipes-support/salt/salt_3000.2.bb
```diff
@@ -107,6 +108,10 @@ do_install_append() {
     cp -r ${S}/tests/ ${D}${PYTHON_SITEPACKAGES_DIR}/${PN}-tests/
 }
 
+do_install_ptest_append() {
+    install -m 0755 ${WORKDIR}/run-ptest ${D}${PTEST_PATH}
```
amstewart: This is already [a standard part](https://github.com/ni/openembedded-core/blob/nilrt/master/hardknott/meta/classes/ptest.bbclass#L44) of the `ptest.bbclass:do_install_base()`. So you can remove it; I think. :)

Sorry; I should have noticed that on my first review.

amstewart (changes requested):
Looks good, pending one improvement below.

shruthi-ravi:
Update v3: Remove ptest append, which is handled automatically by the ptest class. @amstewart

amstewart (approved):
Looks great!

PR #316, opened 2022-01-25
packagegroup-ni-desirable: Relocate onboard package

shruthi-ravi:
The onboard keyboard package is not required to be installed in the
runmode image. This is the current state because packagegroup-ni-xfce
is included in the runmode image. Instead, create a new graphical package group that includes the xfce package group and onboard. This packagegroup is included in packagefeed-ni-core now instead of packagegroup-ni-xfce.
This ensures that onboard is not installed by default into the image and is available for installation from the core feed.

Verified that packagegroup-ni-graphical builds successfully.

@ni/rtos

### DIFF
--- a/recipes-core/packagegroups/packagefeed-ni-core.bb
+++ b/recipes-core/packagegroups/packagefeed-ni-core.bb
@@ -8,6 +8,7 @@ RDEPENDS_${PN} = "\
 	packagegroup-core-boot \
 	packagegroup-ni-base \
 	packagegroup-ni-crio \
+	packagegroup-ni-graphical \
 	packagegroup-ni-internal-deps \
 	packagegroup-ni-nohz-kernel \
 	packagegroup-ni-ptest \
@@ -17,7 +18,7 @@ RDEPENDS_${PN} = "\
 	packagegroup-ni-transconf \
 	packagegroup-ni-tzdata \
 	packagegroup-ni-wifi \
-	${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'packagegroup-core-x11', '', d)} \
+	packagegroup-core-x11 \
 	packagegroup-core-standalone-sdk-target \
 	packagegroup-kernel-module-build \
 	dkms \
@@ -30,5 +31,4 @@ RDEPENDS_${PN}_append_x64 = "\
 	rauc-mark-good \
 	nilrt-grub-runmode \
 	nilrt-grub-safemode \
-	${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'packagegroup-ni-xfce', '', d)} \
 "
--- a/recipes-core/packagegroups/packagegroup-ni-graphical.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-graphical.bb
@@ -1,0 +1,11 @@
+SUMMARY = "Graphical packages for the NI Linux Realtime distribution"
+LICENSE = "MIT"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+inherit packagegroup
+
+RDEPENDS_${PN} = "\
+    onboard \
+    packagegroup-ni-xfce \
+"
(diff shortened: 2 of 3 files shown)


recipes-core/packagegroups/packagegroup-ni-desirable.bb
```diff
@@ -35,6 +35,7 @@ RDEPENDS_${PN} += "\
 	ldd \
 	ltrace \
 	ntpdate \
+	onboard \
```
gratian: Should we add this to recipes-core/packagegroups/packagegroup-ni-coreimagerepo.bb instead? I think we want to make sure this is available (but not installed by default).

shruthi-ravi: I believe that the coreimagerepo package group no longer exists. I think it was sort of replaced by packagefeed-ni-core, and as I understand, we do not want individual packages in there. I couldn't find a better location for this package in the current package group architecture.
Besides, packages in the desirable packagegroup should be available, and not installed by default. These packages are configured to break our CI build if they do not build.

gratian: OK, my information is probably outdated. I was commenting based on a similar commit @amstewart made for nohz kernel (but I guess that's a package group): https://github.com/ni/meta-nilrt/pull/299/commits/ba66f55a5340aa264e90db8e07738d6169776687

amstewart: I think @gratian is objecting to the `onboard` package being in the extra/ feed, and I agree that it doesn't belong there.

I think it makes sense that we would have a packgegroup for "the XFCE desktop environment" and another for "all graphical packages". So what about doing...

```
packagefeed-ni-core
\-> packagegroup-ni-graphical
    |-> packagegroup-ni-xfce
    |   \-> xfce & deps
    \-> onboard
```

And then the images recipes can just be:
```
nilrt-runmode-rootfs.bb
\-> includes/nilrt-xfce.bb
    \-> IMAGE_INSTALL += "packagegroup-ni-xfce"
```

So everything in `-graphical` (inlcuding XFCE) gets built into the core/ feed, but only the `-xfce` packages get installed to the rootfs image.

shruthi-ravi: Addressed in v2. Let me know if that's what you imagined.

shruthi-ravi:
Update v2: 
- Create a new graphical package group that includes the xfce package group and onboard
- Include this packagegroup in packagefeed-ni-core now instead of packagegroup-ni-xfce.

recipes-core/packagegroups/packagegroup-ni-graphical.bb
```diff
@@ -0,0 +1,15 @@
+# (C) Copyright 2013,
+#  National Instruments Corporation.
+#  All rights reserved.
+
```
amstewart: We don't need to include these comments.

recipes-core/packagegroups/packagefeed-ni-core.bb
```diff
@@ -30,5 +30,5 @@ RDEPENDS_${PN}_append_x64 = "\
 	rauc-mark-good \
 	nilrt-grub-runmode \
 	nilrt-grub-safemode \
-	${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'packagegroup-ni-xfce', '', d)} \
+	${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'packagegroup-ni-graphical', '', d)} \
```
amstewart: Since we only have one `DISTRO` these days and it always has the `x11` feature, we can drop the conditional logic here.

amstewart (changes requested):
The dependency relationship is correctly representing what I was suggesting. :+1: 

I just have a couple minors below.

shruthi-ravi:
Update v3:
- Remove copyright text in the graphical packagegroup.
- Remove x11 conditionals in the core package feed.
@amstewart

gratian (approved):

amstewart (approved):

PR #322, opened 2022-01-31
images: Fix oe licensing to work with BUILD_IMAGES_FROM_FEEDS

jpautler:
We use the Yocto COPY_LIC_DIRS option to make the build copy all the
licenses for the OE-built components into the
/usr/share/common-licenses directory. We later compress that directory
into an archive and delete the directory to save space.

The Yocto COPY_LIC_DIRS feature had code to disable the copy license
behavior when the Yocto BUILD_IMAGES_FROM_FEEDS option was also set.
I think the idea was that since BUILD_IMAGES_FROM_FEEDS uses other
already-built packages from a feed, that the COPY_LIC_DIRS feature
could not be guaranteed to locate all licenses and therefore should
do nothing.

I added a COPY_LIC_DIRS_SKIP_PKGS variable that you can set to a
list of package names and the COPY_LIC_DIRS code will assume those
packages are installed from a feed or otherwise already have their
licensing handled separately. For all other packages, the code will
do as before and ensure there is a license defined for that package
and copy it to the central location.

I also made a small change to licenses.inc to change the compression
option from "-I pbzip2" to "--bzip2" since the former was failing
in our nilrt-safemode-initramfs build.

### DIFF
--- a/classes/license_image.bbclass
+++ b/classes/license_image.bbclass
@@ -1,0 +1,281 @@
+ROOTFS_LICENSE_DIR = "${IMAGE_ROOTFS}/usr/share/common-licenses"
+
+python write_package_manifest() {
+    # Get list of installed packages
+    license_image_dir = d.expand('${LICENSE_DIRECTORY}/${IMAGE_NAME}')
+    bb.utils.mkdirhier(license_image_dir)
+    from oe.rootfs import image_list_installed_packages
+    from oe.utils import format_pkg_list
+
+    pkgs = image_list_installed_packages(d)
+    output = format_pkg_list(pkgs)
+    open(os.path.join(license_image_dir, 'package.manifest'),
+        'w+').write(output)
+}
+
+python license_create_manifest() {
+    import oe.packagedata
+    from oe.rootfs import image_list_installed_packages
+
+    skip_pkgs = set((d.getVar('COPY_LIC_DIRS_SKIP_PKGS') or "").split())
+    pkg_dic = {}
+    for pkg in sorted(image_list_installed_packages(d)):
+        if pkg in skip_pkgs:
+            continue
+
+        pkg_info = os.path.join(d.getVar('PKGDATA_DIR'),
+                                'runtime-reverse', pkg)
+        pkg_name = os.path.basename(os.readlink(pkg_info))
+
+        pkg_dic[pkg_name] = oe.packagedata.read_pkgdatafile(pkg_info)
+        if not "LICENSE" in pkg_dic[pkg_name].keys():
+            pkg_lic_name = "LICENSE_" + pkg_name
+            pkg_dic[pkg_name]["LICENSE"] = pkg_dic[pkg_name][pkg_lic_name]
+
+    rootfs_license_manifest = os.path.join(d.getVar('LICENSE_DIRECTORY'),
+                        d.getVar('IMAGE_NAME'), 'license.manifest')
+    write_license_files(d, rootfs_license_manifest, pkg_dic, rootfs=True)
+}
+
+def write_license_files(d, license_manifest, pkg_dic, rootfs=True):
+    import re
+    import stat
+
+    bad_licenses = (d.getVar("INCOMPATIBLE_LICENSE") or "").split()
+    bad_licenses = expand_wildcard_licenses(d, bad_licenses)
+
+    whitelist = []
+    for lic in bad_licenses:
+        whitelist.extend((d.getVar("WHITELIST_" + lic) or "").split())
+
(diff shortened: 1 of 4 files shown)


jpautler:
This is kind of an RFC to confirm if this is the right approach. If so, there's a question of whether we want to pull this alternate copy of license_image.bbclass directly into meta-nilrt now to fix the issue or if we'd rather try to upstream that file first. If I understand correctly, creating a copy of the bbclass in our layer is the solution when you need to change an upstream bbclass. If that's incorrect, let me know.

Separately, I changed the tar compression option from "-I pbzip2" to "--bzip2". For some reason, the original option was causing a build failure on my system that pbzip2 could not be found, but only when building nilrt-safemode-initramfs and not when building nilrt-runmode-rootfs. I don't understand why that was happening in one build and not the other, but I found the "--bzip2" option worked fine everywhere. If you understand what was happening or have an alternate solution, let me know.

classes/license_image.bbclass
```diff
@@ -0,0 +1,281 @@
+ROOTFS_LICENSE_DIR = "${IMAGE_ROOTFS}/usr/share/common-licenses"
```
amstewart, line 1: >If I understand correctly, creating a copy of the bbclass in our layer is the solution when you need to change an upstream bbclass.

Do you have a source for that? I think that logic makes sense if we are trying to *extend* the behavior of the core layers. In our images, we would inherit the core layer's bbclass, and then inherit the meta-nilrt bbclass (or includes). But, if we're trying to *change* core layer behavior, I think maintaining a wholesale-copy of the core layer bbclass would be inefficient.

Also, it makes diffing our changes hard. :)

So I think I would prefer for these changes to live as a commit against the OE-core bbclass - which will also make it easier to upstream.

recipes-core/images/nilrt-runmode-rootfs.bb
```diff
@@ -40,6 +40,12 @@ IMAGE_INSTALL_NODEPS += "\
 	${NI_PROPRIETARY_RUNMODE_PACKAGES} \
 "
 
+COPY_LIC_DIRS_SKIP_PKGS = "\
+        ${NI_PROPRIETARY_COMMON_PACKAGES} \
+	${NI_PROPRIETARY_RUNMODE_PACKAGES} \
+        ni-sysdetails-webservice \
```
amstewart, line 46: NIT: check your whitespace.

classes/license_image.bbclass
```diff
@@ -0,0 +1,281 @@
...
+    output = format_pkg_list(pkgs)
+    open(os.path.join(license_image_dir, 'package.manifest'),
+        'w+').write(output)
+}
+
+python license_create_manifest() {
+    import oe.packagedata
+    from oe.rootfs import image_list_installed_packages
+
+    skip_pkgs = set((d.getVar('COPY_LIC_DIRS_SKIP_PKGS') or "").split())
```
amstewart, line 20: I'm not sure I understand which issue this solution is fixing.

Is the problem that feed `-lic` packages might install their license information to anywhere in the rootfs - so BB doesn't know what content to copy into the common licenses directory, and therefore the tar archive? If so, how does skipping the NI-proprietary content resolve the issue? The NI IPKs shouldn't be more- or less-compliant than OE packages wrt. placing licenses in the correct locations.

amstewart (changes requested):

recipes-core/images/nilrt-runmode-rootfs.bb
billpittman: Extra NIT: Put the COPY_LIC_DIRS_SKIP_PKGS values in the same order for runmode/safemode

COPY_LIC_DIRS_SKIP_PKGS = "\
        ${NI_PROPRIETARY_(RUNMODE or SAFEMODE)_PACKAGES} \
        ${NI_PROPRIETARY_COMMON_PACKAGES} \
        ni-sysdetails-webservice \

classes/license_image.bbclass
billpittman: Adding onto Alex's question, it almost seems like we're saying "If its an NI package, its OK to ship without a license" ... and I'm not exactly sure that's right.  Maybe we should be filing bugs to the appropriate package owners to add a license to their packages?  Unless I'm completely misunderstanding the COPY_LIC_DIRS_SKIP_PKGS purpose.

amstewart:
@jpautler We're dropping this PR in favor of #328; correct?

jpautler:
> @jpautler We're dropping this PR in favor of #328; correct?

Correct. Closing this PR.

PR #324, opened 2022-02-07
shadow, passwd,group: reallocate users/groups between static/dynamic

bstreiff:
The range for dynamic system user/group selection was such that, between dynamically-chosen UID/GID selections for from `useradd` and our own static allocations, we could trivially create conflicts with certain package installation orders.

To resolve this, partition the space such that UIDs/GIDs below our `lvuser`/`ni` entries (which have a well-known UID/GID due to legacy compatibility constraints) is solely for static allocation, and those above that are solely for dynamic allocation. Additionally, renumber some of the users and groups that exist so that they're on the correct side of that partition and also no longer conflicting with each other (oops).

A test is also added to enforce this.

Fixes AzDO#1790954


bstreiff:
looking for feedback on:
- Is stuffing this in the `shadow` recipe a good place to be doing this test? I'm not sure how much of this could stand to be integrated into `useradd-staticids.bbclass` directly.
- renumbering users/groups might be controversial, but I don't think there are any compatibility constraints on the ones that have moved
- is descending order okay? would ascending be preferred?

billpittman:
1. Stuffing the test into shadow seems fine-ish, honestly I'm not sure where else it should go, and the shadow file should have all the necessary data, so it makes sense.
2. Renumbering should be mostly fine, I'm guessing there may be some groups that have fixed numeric uid/gids, but we'll find those fairly quickly as ATS's start to fail with permission errors.  Additionally renumbering _might_ be problematic, but duplicate numbers IS problematic.
3. I prefer descending under the assumption that we are leaving a gap for OE to occupy in the future.

billpittman (approved):

amstewart (approved):
>Is stuffing this in the shadow recipe a good place to be doing this test? I'm not sure how much of this could stand to be integrated into useradd-staticids.bbclass directly.

Assuming that `shadow` owns the `login.defs` file (and it looks to me like it does), I'm fine with it living in the recipe. Since our preferred long-term solution is to drop our dependence on `useradd-staticids` entirely, I don't think its worthwhile to invest in getting this integrated into the `.bbclass` directly, or upstreamed.

>renumbering users/groups might be controversial, but I don't think there are any compatibility constraints on the ones that have moved

:+1:

>is descending order okay? would ascending be preferred?

Seems fine to me.

-----

ACK from me. Are you ready to merge?

bstreiff:
If you're fine with all of this as it is and don't have any changes to recommend, then, yeah, ready to merge. I'll prep a PR to pull this back into sumo as well.

PR #325, opened 2022-02-08
Move kernel-devsrc.inc fixes to nilrt's kernel-dev src include files

shruthi-ravi:
It is difficult to maintain out-of-tree commits to kernel-devsrc in oe-core, especially when trying to do an upstream merge. All modifications to kernel-devsrc.bb have to instead be made to kernel-devsrc.inc. This will also mess up history when doing those merges. 
As an alternative, revert those commits in oe-core (oe-core [PR](https://github.com/ni/openembedded-core/pull/44) )and make the same updates to our nilrt-specific kerneldev src include files.

Builds in progress.

@ni/rtos

### DIFF
--- a/recipes-kernel/linux/kernel-devsrc.bbappend
+++ b/recipes-kernel/linux/kernel-devsrc.bbappend
@@ -1,1 +1,5 @@
 require kernel-devsrc-nilrt.inc
+
+KERNEL_VERSION = "${@get_kernelversion_headers('${B}')}"
+
+RDEPENDS_${PN}:remove = "python3"
--- a/recipes-kernel/linux/kernel-devsrc-debug.bb
+++ b/recipes-kernel/linux/kernel-devsrc-debug.bb
@@ -2,6 +2,6 @@ KERNEL_DEPENDENCY = "linux-nilrt-debug"
 KERNEL_DIRECTORY = "${DEPLOY_DIR_IMAGE}/kernel-debug/staging_kernel_dir"
 KERNEL_BUILD_DIRECTORY = "${DEPLOY_DIR_IMAGE}/kernel-debug/staging_kernel_builddir"
 
-require recipes-kernel/linux/kernel-devsrc.inc
+require recipes-kernel/linux/kernel-devsrc.bb
 require kernel-devsrc-nilrt.inc
 require kernel-devsrc-nilrt-alternate.inc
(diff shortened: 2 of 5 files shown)


recipes-kernel/linux/kernel-devsrc.bbappend
```diff
@@ -1 +1,7 @@
 require kernel-devsrc-nilrt.inc
+
+do_configure[depends] += "virtual/kernel:do_deploy"
```
chaitu236: I don't remember why it was needed for us but not upstream.. maybe its only required for alternate kernels? Can you check if this is required for this recipe?

shruthi-ravi: Can you share how I can check that? Just by building kernel-devsrc? I currently have everything built. Should I clean `linux-nilrt` before that?

chaitu236: Removing the line and building `kernel-devsrc` should validate it - yes I think `linux-nilrt` should be cleaned before that.

chaitu236 (approved):
Everything else looks good.

recipes-kernel/linux/kernel-devsrc.bbappend
shruthi-ravi: Seems to work without this line.

shruthi-ravi:
Update v2: 
Removed `do_configure[depends] += "virtual/kernel:do_deploy"` from the kernel-devsrc.bbappend. The recipe builds without this dependency. Only the alternate kernel devsrc recipes require it.
@chaitu236

recipes-kernel/linux/kernel-devsrc.bbappend
```diff
@@ -1 +1,5 @@
 require kernel-devsrc-nilrt.inc
+
```
amstewart, line 2: >kernel-devsrc: Move out-of-tree recipe fixes to bbappend

For git history, could you either: (a) copy the commit messages from the referenced hashes here into this message, or (b) re-express the justification for these changes in this commit message.

amstewart (changes requested):

recipes-kernel/linux/kernel-devsrc.bbappend
shruthi-ravi: I hoped to achieve that with the git commit hashes in the commit message. I guess I can also add one of the above.

shruthi-ravi:
Update v3: 
Copied reverted git commit messages to the current commit messages for reference. @amstewart

amstewart (approved):

PR #328, opened 2022-02-11
images: Fix licensing to work with BUILD_IMAGES_FROM_FEEDS

jpautler:
Prior to now, we have been using the OE COPY_LIC_MANIFEST and
COPY_LIC_DIRS flags to include license information for all OE-built
packages in our images. That mechanism requires access to the source
for those packages and to intermediate state information that is
generated while building those packages.

Starting with our Hardknott-based release, we are now building our
images from feed contents using the OE BUILD_IMAGES_FROM_FEEDS flag.
Images built from feeds do not usually have access to the package
source or intermediate state information generated when building the
feed packages. As such, OE does not support using the COPY_LIC_MANIFEST
and COPY_LIC_DIRS flags with BUILD_IMAGES_FROM_FEEDS.

This change turns off the COPY_LIC_MANIFEST and COPY_LIC_DIRS flags
and instead installs the license information using the -lic suffixed
packages generated by OE. We had previously been excluding the -lic
packages from installation using the BAD_RECOMMENDATIONS variable. This
was because we were getting the license information using the
COPY_LIC_* flags and did not want the license information to be in the
image twice.

This change gives up some canonicalization and de-duplication work done
by the COPY_LIC_* flags. This results in some size increase. However,
this change also does away with the compression of the licenses into an
archive, which has a bigger effect on disk footprint.

The runmode image size increases by about 15MB while the safemode image
size only increases by about 200k (the safemode rootfs is compressed
into the ramdisk file in the safemode image). If we ever needed to
reduce the disk footprint, we could look into identifying all the files
installed by the -lic packages, compress those files into an archive,
and uninstall those packages from the image. We're not pursuing this
now since these increases seem acceptable, especially in the context of
the x64 targets supported by the Hardknott-based releases.


jpautler:
Tested changes by building everything and examining the /usr/share/licenses/ and /usr/lib/licenses/ directories in the context of running safemode and runmode images.

recipes-core/images/includes/licenses.inc
```diff
@@ -1,16 +0,0 @@
-
-COPY_LIC_MANIFEST = "1"
```
amstewart, line 2: Can you remove the `l` mixin in the [images/README](https://github.com/ni/meta-nilrt/blob/9a3f91edebf83a4598944cc3bb5871d14b1e235b/recipes-core/images/README#L8)?

amstewart (changes requested):
Looks good to me; pending below.

recipes-core/images/includes/licenses.inc
jpautler: Done.

shruthi-ravi (approved):
Wow, the changes turned out to be really simple in the end huh?!
Also, can you squash the two commits, or add more details to the second fixup commit?

amstewart (approved):

jpautler:
> Wow, the changes turned out to be really simple in the end huh?! Also, can you squash the two commits, or add more details to the second fixup commit?

Commits squashed. Yes. For all the headache to figure it out, the solution is really straightforward!

PR #47, opened 2020-10-29
pkgrp-ni-restoremode: Add vmware keyboard support

chaitu236:
Added i8042 driver to restoremode and modprobed in init

Tested on vmware

@ni/rtos

PR #330, opened 2022-02-15
nilrt-recovery-media: Use the same kernel as the recovery image

billpittman:
Use the kernel that was installed into the recovery image to
create the wic boot device to avoid kernel/driver mismatches
in the recovery media ISO.

### Testing
Tested locally with the corresponding patch to openembedded-core/scripts/wic


amstewart (approved):
Is this PR ready to pull before we have OE-core [#47](https://github.com/ni/openembedded-core/pull/47)?

billpittman:
We can pull it before the OE-core #47 -- it just won't change anything.  The OE-core #47 makes that -k switch work.  So in this case it doesn't matter which order they get pulled in.

PR #332, opened 2022-02-19
libxkbcommon: Override recipe to source from git

chaitu236:
Existing recipe gets sources from a tar on xkbcommon.org and in these
sources, parser.c and parser.h are already generated using bison 3.0.4.
When bitbake builds the recipe, it regenerates parser.c and parser.h
(because thats what libxkbcommon's meson rules say) - but this time with
bison 3.7.6 as that's the version our recipe is on.
When compiling parser.o though, the original parser.h gets
included instead of the newly generated one causing build failure due to
parser.c and parser.h mismatch.

This wasn't a problem for 21.0 because bison recipe was on 3.0.4.
But it was updated to 3.7.6 in 21.3 timeframe from which time the recipe
started failing.

So get the sources from github directly where parser.h and parser.c are
not already generated. The SRCREV points to 0.8.0 release.

Also updated `packagegroup-ni-coreimagerepo` to include `qtbase`.

Azdo Bug [#1848107](https://dev.azure.com/ni/DevCentral/_workitems/edit/1848107)

### Testing
Verified libxkbcommon builds locally.

@ni/rtos

### DIFF
--- a/recipes-graphics/xorg-lib/libxkbcommon_0.8.0.bbappend
+++ b/recipes-graphics/xorg-lib/libxkbcommon_0.8.0.bbappend
@@ -1,0 +1,8 @@
+SRCREV = "b82e3b764e60df337ca695e8f8642e7bf42b0cca"
+SRC_URI = "git://github.com/xkbcommon/libxkbcommon.git;protocol=https"
+
+unset SRC_URI[md5sum]
+unset SRC_URI[sha256sum]
+unset UPSTREAM_CHECK_URI
+
+S = "${WORKDIR}/git"


recipes-graphics/xorg-lib/libxkbcommon_0.8.0.bb
```diff
@@ -0,0 +1,20 @@
...
+DESCRIPTION = "libxkbcommon is a keymap compiler and support library which \
+processes a reduced subset of keymaps as defined by the XKB specification."
+HOMEPAGE = "http://www.xkbcommon.org"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=e525ed9809e1f8a07cf4bce8b09e8b87"
+LICENSE = "MIT & MIT-style"
+
+DEPENDS = "util-macros flex-native bison-native"
+
+SRCREV = "b82e3b764e60df337ca695e8f8642e7bf42b0cca"
+SRC_URI = "git://github.com/xkbcommon/libxkbcommon.git;protocol=https"
```
amstewart: Why are you copying the whole recipe to meta-nilrt, instead of using a bbappend?

chaitu236: `SRC_URI[md5sum]`, `SRC_URI[sha256sum]`, `UPSTREAM_CHECK_URI ` are set in the original recipe which we don't want. It seemed cleaner to override the recipe instead of deleting these in bbappend.

recipes-graphics/xorg-lib/libxkbcommon_0.8.0.bb
```diff
@@ -0,0 +1,20 @@
+SUMMARY = "Generic XKB keymap library"
+DESCRIPTION = "libxkbcommon is a keymap compiler and support library which \
+processes a reduced subset of keymaps as defined by the XKB specification."
+HOMEPAGE = "http://www.xkbcommon.org"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=e525ed9809e1f8a07cf4bce8b09e8b87"
+LICENSE = "MIT & MIT-style"
+
```
shruthi-ravi: Do you know why the compiler finds the original parser.h first along with the newer parser.c?
Also, have newer releases of the project updated to use a newer bison?

recipes-graphics/xorg-lib/libxkbcommon_0.8.0.bb
amstewart: I don't think I agree. It is a fairly common pattern for bbappends to update those variables. And preferring a bbappend will let the meta-nilrt code more clearly communicate what parts from upstream we're *intending* to change (the SRC_URI) - which helps maintainability.

chaitu236: Ok, updated it to be a bbappend.

recipes-core/packagegroups/packagegroup-ni-coreimagerepo.bb
```diff
@@ -21,6 +21,9 @@ RDEPENDS_${PN} = "\
 "
 
 RDEPENDS_${PN}_append_x64 = "\
-	${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'packagegroup-ni-xfce', '', d)} \
+	${@bb.utils.contains('DISTRO_FEATURES', 'x11', '\
+	packagegroup-ni-xfce \
+	qtbase \
```
amstewart: `packagegroup-ni-coreimagerepo` isn't used in sumo. It was the core feed group for the deprecated `nilrt-nxg` distro flavor, but isn't invoked in our sumo CG components.

You'll need to add your deps here to the [`ADDITIONAL_CORE_FEED_RECIPES`](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/rtos/rtoslegacyd/NIOpenEmbedded/x64_dkms.mak&version=GBmain&line=11&lineEnd=12&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) variable in the `NIOpenEmbedded/x64_dkms.mak` target.

amstewart (changes requested):

recipes-graphics/xorg-lib/libxkbcommon_0.8.0.bb
chaitu236: > Do you know why the compiler finds the original parser.h first along with the newer parser.c?

I haven't looked closely at this but it is probably the order of include paths thats the issue - the includes contain paths to both original source and the directories where build is happening.

> Also, have newer releases of the project updated to use a newer bison?

Newer releases don't have `parser.c` and `parser.h` anymore and yes they have been building fine on hardknott where we have newer bison

recipes-core/packagegroups/packagegroup-ni-coreimagerepo.bb
chaitu236: Removed the commit from here.

shruthi-ravi (approved):
I wonder if the better change is to fix the compiler's include path hierarchy. But maybe that's not so easy to do in bitbake?
PS, can you add the azdo bug link to the PR description?

chaitu236:
> I wonder if the better change is to fix the compiler's include path hierarchy. But maybe that's not so easy to do in bitbake? PS, can you add the azdo bug link to the PR description?

That would require understanding libxkbcommon's build files and adding a patch to fix it there - I don't think its worth it seeing that we don't hit the issue on the same version from github (which this PR uses), and seeing that the newer versions don't have the issue at all.

> PS, can you add the azdo bug link to the PR description?

Done

amstewart (approved):

PR #335, opened 2022-02-23
Rework how sysconfig-settings are handled

bstreiff:
Multiple changes in this patch set relating to System Configuration INI/UIXML files.
- Updates the UIXML install path to the current canonical location.
- Previously, `sysconfig-settings(-ui)` would put down `.ini` files, and then scripts from `initscripts-nilrt` and `crio-support-scripts` would clobber right over them. Stop doing that; anything that does anything with these `.ini` files should be part of the `sysconfig-settings` package
- Do not clobber the `.ini` files anymore; instead we symlink to them if they're supposed to show up as available options on the target.
- Minor file renames and comments added to improve clarity.
- Move `niselectsystemsettings` into being properly an ontarget postinst task, as that was essentially its behavior anyway.
- A minor fix to `nisetembeddeduixml` to work with fbdev-type devices, which means that now I can get the "Embedded UI" checkbox in QEMU.

Addresses AzDO#1827936. Does not address related AzDO#1821122.


recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -119,5 +122,5 @@ pkg_postinst_ontarget_${PN} () {
 	# Add the memory reservation for Base
 	/usr/local/natinst/bin/nirtcfg -s section=RtLinuxMemReserve,token=Base,value=24
 
-	${sysconfdir}/natinst/niselectsystemsettings postinst
+	/etc/natinst/niselectsystemsettings postinst
```
amstewart: Looks like this change went the wrong direction. Shouldn't it have left the `${sysconfdir}` variable alone?

recipes-ni/sysconfig-settings/files/nisetembeddeduixml
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
amstewart, line 1: Merge commits:
* initscripts-nilrt: remove nisetembeddeduixml
* sysconfig-settings: add nisetembeddeduixml

I would prefer a single commit which shows the `nisetembeddeduixml` script being moved.

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -122,13 +120,20 @@ do_install () {
...
-	/etc/natinst/niselectsystemsettings postinst
+	TARGET_CLASS=$(/sbin/fw_printenv -n TargetClass 2>&1)
+
+	# Ethernet RIO targets should not have RT startup settings
+	if [ "$TARGET_CLASS" = "Ethernet RIO" ]; then
+		echo "" > ${systemsettingsdir}/rt_target.ini
+	fi
+
+	# cDAQ targets should not have FPGA startup settings
+	if [ "$TARGET_CLASS" = "cDAQ" -o "$TARGET_CLASS" = "CVS" ]; then
```
amstewart: Are all these `TARGET_CLASS`es actually supported in hardknott? If not, can we afford to drop some of this logic?

recipes-ni/sysconfig-settings/files/nisetembeddeduixml
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 system_has_display_driver() {
-	[ -d /sys/class/drm ]
+	[ -d /sys/class/drm -o -e /sys/class/graphics/fb0 ]
```
amstewart, line 7: I don't follow the change here; could you elaborate in the commit message?

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -5,6 +5,7 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 SECTION = "base"
 
 uixmldir = "${datadir}/nisysapi/uixml"
+settingsdatadir = "${datadir}/sysconfig-settings/systemsettings"
```
amstewart: MINOR: could we use `${PN}` here instead of the `sysconfig-settings` static string?

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -136,6 +138,19 @@ pkg_postinst_ontarget_${PN} () {
...
+		ARCH=$(uname -m 2>/dev/null)
+		if [ "$ARCH" = "x86_64" ]; then
+			efiConsoleOutEnable=$(/sbin/fw_printenv -n BootFirmwareConsoleOutEnable 2>/dev/null || true)
+			if [ -z "$efiConsoleOutEnable" ]; then
+				HAS_CONSOLE=0
+			fi
+		fi
+
+		if [ "$HAS_CONSOLE" = "1" ]; then
+			ln -sf ${settingsdatadir}/consoleout.ini ${systemsettingsdir}/consoleout.ini
```
amstewart: Since this is the only logic that we execute when `$HAS_CONSOLE` is true, can you just move it up into the `if [ -z "${efiConsoleOut...` block, and do-away with the `HAS_CONSOLE` indirection variable?

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -149,5 +164,6 @@ pkg_postinst_ontarget_${PN} () {
 pkg_prerm_ontarget_${PN} () {
 	rm -f ${systemsettingsdir}/target_common.ini \
 	      ${systemsettingsdir}/rt_target.ini \
-	      ${systemsettingsdir}/fpga_target.ini
+	      ${systemsettingsdir}/fpga_target.ini \
+              ${systemsettingsdir}/consoleout.ini
```
amstewart: NIT: whitespace?

amstewart (changes requested):
I like the refactors here. Good work. :+1:

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
bstreiff: Ethernet RIO is probably droppable here. Trying to confirm on CVS. cDAQ should stay.

I'd really rather these checks be against available hardware instead of target class, but I'm not sure how to detect some of these generically without, for example, enumerating the PCI bus looking for a LabVIEW-targetable FPGA on the bus with one of a growing list of PIDs. Yuck.

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -141,10 +143,16 @@ pkg_postinst_ontarget_${PN} () {
 		ln -sf ${settingsdatadir}/fpga_target.ini ${systemsettingsdir}/fpga_target.ini
 	fi
 
+	# add console out if we have a firmware variable for it (x86_64 targets only)
+	efiConsoleOutEnable=$(/sbin/fw_printenv -n BootFirmwareConsoleOutEnable 2>/dev/null || true)
```
amstewart: Can we reliably use `PATH` to find `fw_printenv` here (and on line 133 in this file), and add an RDEPENDS on the `fw-printenv` package?

amstewart (changes requested):

amstewart:
@bstreiff If you're ready for this to be reviewed, could you post a patchrev summary of your changes?

bstreiff:
last two pushes were:
- use PATH to find fw_printenv add an RDEPENDS on the fw-printenv package
- resolve patch conflicts against changes that had gone in since creating this PR

amstewart (approved):
I'll pull this after jxie's tpm removal patchset clears CI.

shruthi-ravi (approved):

PR #336, opened 2022-02-23
[hardknott] Create niacctbase-sudo and move to recipes-ni

amstewart:
While working the sudo conffiles issue in [Task 1856146](https://dev.azure.com/ni/DevCentral/_workitems/edit/1856146), I noticed that we have a sudo.bbappend which applies some NI-specific logic directly into `/etc/sudoers`. It's weird because (a) the sudo.bbappend contains information about accounts from `niacctbase`, which sudo otherwise shouldn't care about. And (b) injecting rules directly into `/etc/sudoers` should be avoided where we could more easily distribute an `/etc/sudoers.d/` run-parts file.

So this PR:
* moves the NI-specific (non `nilrtdiskcrypt`) logic into a new subpackage called `niacctbase-sudo`. (look at the first commit to see the diff)
    * `niacctbase` recommends `niacctbase-sudo`, and `-sudo` in turn depends on `niacctbase` and `sudo`.
    * `/etc/sudoers.d/niacctbase` (the run-parts file) is a CONFFILE of `niacctbase-sudo`.
* moves the `niacctbase` recipe from `recipes-core` into `recipes-ni`, since it is very much an NI recipe.

# Testing
* Rebuilt the affected files and confirmed on a new hardknott safemode VM:
    * `niacctbase-sudo` is installed by default (via the recommends :+1:
    * `niacctbase-sudo` installs and owns the `/etc/sudoers.d/niacctbase` file.
    * The rules in the sudoers file seem to be applied correctly

@ni/rtos ; is there any objection to creating this subpackage?


recipes-extended/sudo/sudo_%.bbappend
```diff
@@ -1,14 +1,10 @@
 do_install_append() {
 	# only for nilrt and nilrt-xfce
 	if ${@oe.utils.conditional('DISTRO', 'nilrt-nxg', 'false', 'true', d)}; then
-		echo >>"${D}${sysconfdir}/sudoers" ""
-		echo >>"${D}${sysconfdir}/sudoers" "admin ALL=(ALL) NOPASSWD: ALL"
 		echo >>"${D}${sysconfdir}/sudoers" ""
 		echo >>"${D}${sysconfdir}/sudoers" "User_Alias TPM_QUOTERS = lvuser, webserv"
 		echo >>"${D}${sysconfdir}/sudoers" "TPM_QUOTERS ALL=(#0:#0) NOPASSWD: /usr/sbin/nilrtdiskcrypt_quote -d"
 		echo >>"${D}${sysconfdir}/sudoers" "TPM_QUOTERS ALL=(#0:#0) NOPASSWD: /usr/sbin/nilrtdiskcrypt_quote -j"
 		echo >>"${D}${sysconfdir}/sudoers" "TPM_QUOTERS ALL=(#0:#0) NOPASSWD: /usr/sbin/nilrtdiskcrypt_quote -j -q 0x[0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F]"
```
amstewart, line 8: NOTE: I didn't bother doing anything with these rules, because we're planning to deprecate `nilrtdiskcrypt` in the near future.

recipes-ni/niacctbase/niacctbase/sudoers
```diff
@@ -0,0 +1,6 @@
+
+# Allow the admin user (UID 0) to invoke sudo without a password
+admin ALL=(ALL) NOPASSWD: ALL
+
+# Allow lvuser to modify the opkg state without a password
+lvuser ALL=(#0:#0) NOPASSWD: /usr/bin/opkg info *
```
amstewart, line 6: NOTE: I'm not actually sure this rule is needed, now that opkg will allow unpriviledged users to perform read-only operations. But it isn't worth the effort at the moment for me to track down this requirement.

amstewart:
```bash
(safemode) admin@NI-cRIO-903x-VM-c318a270:~# opkg files niacctbase-sudo
Package niacctbase-sudo (1.0-r0.1411) is installed on root and has the following files:
/etc/sudoers.d/niacctbase
(safemode) admin@NI-cRIO-903x-VM-c318a270:~# opkg info niacctbase-sudo
Package: niacctbase-sudo
Version: 1.0-r0.1411
Depends: niacctbase, sudo-lib
Recommends: niacctbase-lic
Status: install ok installed
Architecture: all
Conffiles:
 /etc/sudoers.d/niacctbase d41ecc990158c786d0383dd6bca6c379
Installed-Size: 206
Installed-Time: 1645575442
```

recipes-core/niacctbase/niacctbase.bb
```diff
@@ -27,17 +28,19 @@ GROUPADD_PARAM_${PN} = " --system ${LVRT_GROUP}; \
...
 
 # add parameter -m if you want home directories created with default files (.profile, .bashrc)
-USERADD_PARAM_${PN} = " -m -N -g ${LVRT_GROUP} -G network,niwscerts,plugdev,tty,video -c 'LabVIEW user' ${LVRT_USER}; \
-		-N -g openvpn -G network -c 'OpenVPN' -r openvpn; \
-		${@oe.utils.conditional('DISTRO', 'nilrt-nxg', '', d.getVar('USERADD_WEBSERVICES', True), d)}"
+USERADD_PARAM_${PN} = " \
+	-m -N -g ${LVRT_GROUP} -G network,niwscerts,plugdev,tty,video -c 'LabVIEW user' ${LVRT_USER}; \
+	-N -g openvpn -G network -c 'OpenVPN' -r openvpn; \
+	${@oe.utils.conditional('DISTRO', 'nilrt-nxg', '', d.getVar('USERADD_WEBSERVICES', True), d)} \
+"
```
amstewart: Fun fact: none of these settings actually get applied, because they are overwritten by a static assignment by the `useradd-staticids` class. I presume these were added for `nilrt-nxg`, but are now vestigial.

I didn't consider it within the scope of the work that I'm doing to track this down, but hopefully someone will remember this note later.

recipes-core/niacctbase/niacctbase/sudoers
```diff
@@ -0,0 +1,6 @@
+
```
shruthi-ravi: Confirming that this file is installed to a niacctbase namespaced directory.

shruthi-ravi (approved):

recipes-core/niacctbase/niacctbase/sudoers
amstewart: On-target? It gets [installed](https://github.com/ni/meta-nilrt/pull/336/files#diff-01422a91d9db4c848e9e979398fc25e4c9620da8553ab3dd96c836b1a3f64692R51) as `/etc/sudoers.d/niacctbase` (which I confirmed in the original round of testing.)

PR #341, opened 2022-02-26
nilrt-runmode-initramfs: Remove initramfs recipe and references

billpittman:
Remove the initramfs recipe since it is no longer needed by nilrt
boot sequence and unnecessarily complicates the boot logic.

This change will delete the enable_initramfs variable from the grub
environment during the installation of the runmode system.  If that
variable is not present, the system will not attempt to boot from
an initramfs, any implementations that require an initramfs in runmode
will not work with this change.

Natinst-AzDO-ID: 1868559

### Testing
Provisioned a cRIO, PXI-8861, Hyper-V VM and QEMU VM with the nilrt-recovery-media-x64.iso containing these changes and verified MAX could recognize and install to all of them.


amstewart (changes requested):
Do you need to do anything with the [`nisetbootmode(.functions)`](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-ni/ni-utils/files/nisetbootmode.functions#L6) files?

billpittman:
@amstewart Yup nice catch I forgot to search for bootflags.d references too.  Got it now and there appears to be no other references to bootflags.  Also tested the scripts on an installed runmode&safemode system.

recipes-ni/ni-systemformat/files/nisystemformat
```diff
@@ -648,15 +648,6 @@ if [ "$ENCRYPT" = yes ]; then
 		die_with_usage NO_TPM_HARDWARE \
 			"The TPM chip on this machine could not be detected (does it have one?)"
 	fi
-
-	# Sanity check, unlikely to hit
-	# Encrypted disk won't boot unless system is provisioned with initramfs enabled
```
chaitu236, line 653: So have we never enabled encryption of configfs, rootfs partitions? I do see some code related to that but wasn't sure if it was in use.

recipes-core/initrdscripts/files/ni_provisioning.safemode
```diff
@@ -319,14 +295,8 @@ install_grubenv()
 	grub-editenv $BOOTFS_MOUNTPOINT/grub/grubenv set "BIOSBootMode=efi"
 	grub-editenv $BOOTFS_MOUNTPOINT/grub/grubenv set "NITarget=$NI_TARGET"
 
-	# If target has a TPM chip or runs on Hyper-V, we need to enable an alternate boot flow
-	# involving initramfs and bootflags dir
-	if $use_ramfs; then
-		grub-editenv $BOOTFS_MOUNTPOINT/grub/grubenv set "enable_initramfs=1"
```
chaitu236, line 325: So going forward, a target provisioned with old safemode won't be able to boot new runmode on targets with TPM chip. I guess this is acceptable to us? Should we add checks somewhere to ensure new runmode isn't installed on old safemode?

recipes-ni/ni-systemformat/files/nisystemformat
billpittman: Right, pretty sure its been broken for a while now,  The upstream API changed so most of the plumbing we have won't work even if we enabled it.

recipes-core/initrdscripts/files/ni_provisioning.safemode
billpittman: Sort of, old safemode will still boot on all targets, however if enable_initramfs is set to 1 (because the target has a TPM chip, or whatever), the new runmode won't boot since the initramfs will be "missing on purpose due to this change".  If the old safemode has enable_initramfs==0, then it will boot the old safemode/new runmode combo just fine since neither one requires the initramfs.

Does that answer your question?

shruthi-ravi (commented):
Since there are changes to so many files, can you enumerate the repercussions of these changes in the commit message? i.e. what is possible and not once this change goes in? Chaitanya seems to have asked a few questions related to that.
And can you also add a link to the azdo WI or WI number to the PR?
The azdo WI has a lot of the details I am requesting, but I would prefer if some of it were in the commit message too.

amstewart (approved):

amstewart:
I'll wait until everyone on the review list has given their ACK before merging.

recipes-core/initrdscripts/files/ni_provisioning.safemode
chaitu236: I understand that thats the case. I'm wondering if/how we will enforce it - right now I believe MAX doesn't stop us from installing hardknott runmode on a sumo safemode. Should we create a story to research/implement this?

chaitu236: PS: Have we broken safemode compatibility for new releases like this before? @billpittman, @jeminor

billpittman: I don't believe we've broken compatibility necessarily.  In the case of a TPM enabled system, then yes, but we've dropped that capability, so its a compatibility break in the sense that we don't support it.  Safemode still installs runmode, even an older safemode.

billpittman: @chaitu236 Pointed out a specific case where TPM capable targets would require a safemode upgrade in order to properly install.  Without the update, MAX would "successfully" provision the target, but when it boots into the newly installed runmode, it would fail to find the initramfs file and force the target into  safemode with a software installation error.

Do we have a way to "force" a safemode upgrade from MAX or inform the user that the "problem" is the old safemode?  We used to have a mechanism like that, but it was removed a very long time ago and I don't remember if we put in something else in its place.

If we don't have a way to inform the user that there's a safemode upgrade required, we may have to delay/abandon this change for the moment until we can handle TPM capable systems.

amstewart: I think we do have a way for MAX to notify customers that a safemode upgrade is required. Could you ask Dan and co.

jeminor: There is a mechanism and can require it be invoked if needed.  Specifically, we can require that a system image install require >= some specific safemode.

I'm fine with this, and we can invoke that mechanism when we need it.

jeminor (approved):

recipes-core/initrdscripts/files/ni_provisioning.safemode
chaitu236: @billpittman can you create a story parented to hardknott work so we don't forget it.

chaitu236 (approved):

billpittman:
Created a new US to force a safemode update and expanded the commit message a bit to include the impact of the change.

billpittman:
Added code to remove the enable_initramfs variable from the grubenv to prevent older safemodes from attempting to use an initramfs to boot runmode.  As an intended side effect, the change to postinst also removes the previous requirement to force a safemode update.

amstewart (approved):
I assume the new content is in `nilrt-base-system-image.postinst`. If so, looks good to me.

shruthi-ravi (approved):

PR #342, opened 2022-02-28
ni: update nisetupkernelconfig to allow coredump to work in safemode

ni-jxie:
If it is safemode, create core dump in non-volatile /mnt/userfs.

Signed-off-by: Jerry Xie <jerry.xie@ni.com>

### Testing
Run the updated nisetupkernelconfig script locally and run cmds below to generate core dumps in safemode and verify that the core dump is generated at `/mnt/userfs/var/local/natinst/log/`.
```
ulimit -c unlimited
sleep 10 &
killall -SIGSEGV sleep
```

@ni/rtos

### DIFF
--- a/recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
+++ b/recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
@@ -7,6 +7,9 @@
 #executable name as the suffix.
 safemodeflag=/etc/natinst/safemode
 if [ -f $safemodeflag ]; then
+  if [ ! -d /mnt/userfs/var/local/natinst/log ]; then
+    mkdir -p /mnt/userfs/var/local/natinst/log
+  fi
   echo "/mnt/userfs/var/local/natinst/log/core_dump.%E" > /proc/sys/kernel/core_pattern
 else
   echo "/var/local/natinst/log/core_dump.%E" > /proc/sys/kernel/core_pattern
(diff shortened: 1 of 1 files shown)


ni-jxie:
`/mnt/userfs/var/local/natinst/log/` does not exist unless a runmode is installed, should we consider creating this directory so that safe mode can generate core dump even if no runmode is installed?

amstewart (approved):
>/mnt/userfs/var/local/natinst/log/ does not exist unless a runmode is installed, should we consider creating this directory so that safe mode can generate core dump even if no runmode is installed?

That sounds reasonable (and preferable) to me. Would you mind soliciting the team for a third opinion, though. I'm worried that I'm not considering an angle here.

recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
```diff
@@ -7,6 +7,9 @@
 #executable name as the suffix.
 safemodeflag=/etc/natinst/safemode
 if [ -f $safemodeflag ]; then
+  if [ ! -d /mnt/userfs/var/local/natinst/log ]; then
```
ni-jxie, line 10: This creates the directory for admin user. After a runmode is installed, this directory will be recreated with lvuser permission. Core dump works regardless of the permission.

amstewart (approved):

recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
shruthi-ravi: Are the directory and it's contents cleared and created again when runmode is installed?

shruthi-ravi (approved):
NIT: Update the first commit header to 
nisetupkernelconfig: allow....

recipes-ni/initscripts-nilrt/files/nisetupkernelconfig
ni-jxie: Yes

PR #343, opened 2022-02-28
packagegroup-ni-desirable: Add sysconfig-settings-ui

jpautler:
The sysconfig-settings-ui and sysconfig-settings packages are required by our niskyline__rt__client__runtimei builds. Since we require these packages, they should be in desirable.

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-desirable.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-desirable.bb
@@ -47,6 +47,7 @@ RDEPENDS_${PN} += "\
 	rsync \
 	sshpass \
 	strace \
+	sysconfig-settings-ui \
 	valgrind \
 	vim \
 "


chaitu236:
We add these in both [safemode](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-core/packagegroups/packagegroup-ni-base.bb#L91) and [runmode](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-core/packagegroups/packagegroup-ni-runmode.bb#L127) packagegroups so they end up in feeds also. Why add them separately to `desirable`?

jpautler:
Upon closer examination, it looks like this change actually isn't needed in hardknott since these packages are already in the main feed in hardknott. I only need this change in sumo. Closing this PR and will create a new one targetted to sumo.

PR #344, opened 2022-02-28
packagegroup-ni-desirable: Add sysconfig-settings-ui

jpautler:
The sysconfig-settings-ui package is required by our niskyline__rt__client__runtimei build. Since we require this package, it should be in the desirable packagegroup.

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-desirable.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-desirable.bb
@@ -52,6 +52,7 @@ RDEPENDS_${PN} += "\
 	rsync \
 	sshpass \
 	strace \
+	sysconfig-settings-ui \
 	valgrind \
 	vim \
 "


recipes-core/packagegroups/packagegroup-ni-desirable.bb
```diff
@@ -52,6 +52,7 @@ RDEPENDS_${PN} += "\
 	rsync \
 	sshpass \
 	strace \
+	sysconfig-settings-ui \
```
amstewart, line 55: If this is a dependency of NI-owned components, then it needs to go into the [`ADDITIONAL_CORE_FEED_RECIPES`](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/rtos/rtoslegacyd/NIOpenEmbedded/x64_dkms.mak&version=GBmain&line=11&lineEnd=12&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) variable, rather than the `-desirable` packagegroup.

amstewart (changes requested):

recipes-core/packagegroups/packagegroup-ni-desirable.bb
jpautler: Got it. I'll make the change there. Closing this PR.

PR #345, opened 2022-03-02
Create ni-rtfeatures package to contain handle_cpld_ip_reset script and rtfeatures udev rules

ni-jxie:
### Description

handle_cpld_ip_reset script is missing on Hardknott, and we need it for the long push button reset to reset network configuration on cRIO controllers. handle_cpld_ip_reset is a script that interact with rtfeatures in-tree driver which relies on udev rules for rtfeatures. So let's collect the rtfeatures related stuff together into its own package.

### Implementation

- Create new package ni-rtfeatures for handle_cpld_ip_reset script and rtfeatures udev rules (migrated from udev-extraconf)
- Add ni-rtfeatures to packagegroup-ni-base
- Simplify handle_cpld_ip_reset to remove fallback mechansim

### Testing

- Provisioned VM and made sure all the stuff is in the right place. For a VM, handle_cpld_ip_reset is basically a no-op;
- Tested on real cRIO 9049 and validate that pushing reset button > 5 seconds force the target to safe mode and then pushing reset button > 5 seconds again resets network configuration and force target to reboot to safe mode again.

@ni/rtos

### DIFF
--- a/recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
+++ b/recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
@@ -11,6 +11,6 @@ fi
 
 if [ "$IP_RESET_REQUESTED" == "y" ]; then
 	mkdir -p /tmp/ip_reset_dir
-	/usr/local/natinst/bin/ninetcfgutil pulldefault -d /tmp/ip_reset_dir -g primary -g secondary
+	ninetcfgutil pulldefault -d /tmp/ip_reset_dir -g primary -g secondary
 	rm -rf /tmp/ip_reset_dir
 fi
--- a/recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
+++ b/recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
@@ -1,25 +1,12 @@
 #!/bin/sh
 
-# We are going to create a udev script to mount nirtfeatures under /dev
-# to make us more robust to changes in the absolute path under /sys. However
-# to allow things to continue to work until everything makes its way through
-# the system or in case that udev script doesn't run for some reason we fall
-# back to trying to find ip_reset in a likely spot.
-
 IP_RESET_FILE="/dev/nirtfeatures/ip_reset"
-SMBIOS_HELPER="/usr/share/nisysinfo/smbios_helper"
-
-if [ ! -e $IP_RESET_FILE ]; then
-        [ "${VERBOSE}" != "no" ] && echo "Unable to find ip_reset at normal spot. Attempting to find dynamically."
-        IP_RESET_FILE=`find /sys/bus/acpi/drivers/nirtfeatures -maxdepth 2 -follow -name ip_reset`
-fi
 
 IP_RESET_REQUESTED=""
-if [ -n "$IP_RESET_FILE" ]; then
+if [ -e "$IP_RESET_FILE" ]; then
 	grep -q '1' $IP_RESET_FILE && IP_RESET_REQUESTED="y"
-elif [ -r $SMBIOS_HELPER ]; then
-	source $SMBIOS_HELPER
-	get_ipreset && IP_RESET_REQUESTED="y"
+else
+	[ "${VERBOSE}" != "no" ] && echo "Unable to find ip_reset at normal spot."
 fi
 
 if [ "$IP_RESET_REQUESTED" == "y" ]; then
--- a/recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
+++ b/recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+# We are going to create a udev script to mount nirtfeatures under /dev
+# to make us more robust to changes in the absolute path under /sys. However
+# to allow things to continue to work until everything makes its way through
+# the system or in case that udev script doesn't run for some reason we fall
+# back to trying to find ip_reset in a likely spot.
+
+IP_RESET_FILE="/dev/nirtfeatures/ip_reset"
+SMBIOS_HELPER="/usr/share/nisysinfo/smbios_helper"
+
+if [ ! -e $IP_RESET_FILE ]; then
+        [ "${VERBOSE}" != "no" ] && echo "Unable to find ip_reset at normal spot. Attempting to find dynamically."
+        IP_RESET_FILE=`find /sys/bus/acpi/drivers/nirtfeatures -maxdepth 2 -follow -name ip_reset`
+fi
+
+IP_RESET_REQUESTED=""
+if [ -n "$IP_RESET_FILE" ]; then
+	grep -q '1' $IP_RESET_FILE && IP_RESET_REQUESTED="y"
+elif [ -r $SMBIOS_HELPER ]; then
+	source $SMBIOS_HELPER
+	get_ipreset && IP_RESET_REQUESTED="y"
+fi
+
+if [ "$IP_RESET_REQUESTED" == "y" ]; then
+	mkdir -p /tmp/ip_reset_dir
+	/usr/local/natinst/bin/ninetcfgutil pulldefault -d /tmp/ip_reset_dir -g primary -g secondary
+	rm -rf /tmp/ip_reset_dir
+fi
--- a/recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
+++ b/recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
@@ -6,7 +6,7 @@ IP_RESET_REQUESTED=""
 if [ -e "$IP_RESET_FILE" ]; then
 	grep -q '1' $IP_RESET_FILE && IP_RESET_REQUESTED="y"
 else
-	[ "${VERBOSE}" != "no" ] && echo "Unable to find ip_reset at normal spot."
+	[ "${VERBOSE}" = "yes" ] && echo "Unable to find ip_reset at normal spot."
 fi
 
 if [ "$IP_RESET_REQUESTED" == "y" ]; then
--- a/recipes-ni/ni-rtfeatures/files/rtfeatures.rules
+++ b/recipes-ni/ni-rtfeatures/files/rtfeatures.rules
@@ -1,0 +1,5 @@
+# create symlink to establish a uniform path to nirtfeatures
+DRIVER=="nirtfeatures",RUN+="/bin/ln -sf /sys%p /dev/nirtfeatures"
+
+#give all users read/write access to CPLD-created /dev/input/event?
+SUBSYSTEM=="input", ENV{ID_PATH}=="acpi-NIC775D:??", KERNEL=="event?", MODE="0666"
--- a/recipes-ni/ni-rtfeatures/ni-rtfeatures_1.0.bb
+++ b/recipes-ni/ni-rtfeatures/ni-rtfeatures_1.0.bb
@@ -1,5 +1,5 @@
 SUMMARY = "rtfeatures user-space tools"
-DESCRIPTION = "Provides user-space tools to support rtfeatures."
+DESCRIPTION = "Provides user-space tools to support the nirtfeatures kernel module."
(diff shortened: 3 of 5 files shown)


recipes-ni/ni-hw-scripts/ni-hw-scripts-crio/init.d/handle_cpld_ip_reset
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+# We are going to create a udev script to mount nirtfeatures under /dev
```
ni-jxie, line 3: We do have an udev rule to generate symlink under /dev/nirtfeatures, so maybe we can simplify this script.

```
(safemode) admin@NI-cRIO-9049-01CEEE05:~# cat /etc/udev/rules.d/localextra.rules
...
# create symlink to establish a uniform path to nirtfeatures
DRIVER=="nirtfeatures",RUN+="/bin/ln -sf /sys%p /dev/nirtfeatures"

#give all users read/write access to CPLD-created /dev/input/event?
SUBSYSTEM=="input", ENV{ID_PATH}=="acpi-NIC775D:??", KERNEL=="event?", MODE="0666"
```

However, it does mention in the comment that this can also handle case where the udev script doesn't run and then we can fall back. So we might want to keep the fallback logic.

Thoughts?

ni-jxie: We discussed in our standup today that if we have good testing, we can simplify the script. So I will go ahead and simplify the script. Then I will test it out on real hardware.

recipes-ni/ni-hw-scripts/ni-hw-scripts-crio_1.0.bb
```diff
@@ -0,0 +1,54 @@
+SUMMARY = "Support scripts and utilities common to all NI hardware products"
+DESCRIPTION = "Support scripts and utilities for all NI hardware products which are supported by NI LinuxRT."
```
amstewart: Did you forget to update these for `-crio`?

recipes-ni/ni-hw-scripts/ni-hw-scripts-crio_1.0.bb
```diff
@@ -0,0 +1,54 @@
...
+
+
+PACKAGE_ARCH = "all"
+PACKAGES_remove += "${PN}-staticdev ${PN}-dev ${PN}-dbg"
+
+FILES_${PN} += "\
+	${sysconfdir}/init.d/handle_cpld_ip_reset \
+"
+
+RDEPENDS_${PN} += "\
```
amstewart: Should this also RDEPEND on whatever package provides `/usr/local/natinst/bin/ninetcfgutil`?

amstewart (changes requested):

recipes-ni/ni-hw-scripts/ni-hw-scripts-crio/init.d/handle_cpld_ip_reset
```diff
@@ -1,26 +1,9 @@
...
-        [ "${VERBOSE}" != "no" ] && echo "Unable to find ip_reset at normal spot. Attempting to find dynamically."
-        IP_RESET_FILE=`find /sys/bus/acpi/drivers/nirtfeatures -maxdepth 2 -follow -name ip_reset`
-fi
 
 IP_RESET_REQUESTED=""
-if [ -n "$IP_RESET_FILE" ]; then
-	grep -q '1' $IP_RESET_FILE && IP_RESET_REQUESTED="y"
-elif [ -r $SMBIOS_HELPER ]; then
-	source $SMBIOS_HELPER
-	get_ipreset && IP_RESET_REQUESTED="y"
```
ni-jxie: I remove the SMBIOS_HELPER fallback as well, because it does not work. get_ipreset actually expects an argument which is not provided by the call here.

recipes-ni/ni-hw-scripts/ni-hw-scripts-crio_1.0.bb
ni-jxie: Good catch.

recipes-ni/ni-hw-scripts/ni-hw-scripts-crio_1.0.bb
ni-jxie: Doh!

recipes-ni/ni-hw-scripts/ni-hw-scripts-crio/init.d/handle_cpld_ip_reset
ni-jxie: Right now if the udev rule is not there, then the script would fail with a file/directory not found error. Should we fail more gracefully?

amstewart: Yes? :)

Is it actually failure-case, if the udev rule is missing? I thought the raison d'etre for the initscript was to gracefully handle that case.

amstewart: Also; does the udev rule get installed by a package? Maybe this script should RECOMMEND that package.

recipes-ni/ni-hw-scripts/ni-hw-scripts-crio_1.0.bb
ni-jxie: Updated for cRIO.

recipes-ni/ni-hw-scripts/ni-hw-scripts-crio_1.0.bb
ni-jxie: Added dependency to ni-netcfgutil.

recipes-ni/ni-hw-scripts/ni-hw-scripts-crio/init.d/handle_cpld_ip_reset
ni-jxie: In the standup today, @gratian suggested that we should simplify the script if the udev rule is there and we can test on real hardware to validate that it does work. But having no fall-back does make me feel uneasy. I don't know enough about udev to know under what circumstance it would fail.

The udev rule is in `/etc/udev/rules.d/localextra.rules` which is from `udev-extraconf` package. Since the script would fail wihout the udev rule, maybe the relationship should be stronger than recommended?

recipes-ni/ni-rtfeatures/files/rtfeatures.rules
```diff
@@ -0,0 +1,5 @@
+# create symlink to establish a uniform path to nirtfeatures
+DRIVER=="nirtfeatures",RUN+="/bin/ln -sf /sys%p /dev/nirtfeatures"
+
+#give all users read/write access to CPLD-created /dev/input/event?
+SUBSYSTEM=="input", ENV{ID_PATH}=="acpi-NIC775D:??", KERNEL=="event?", MODE="0666"
```
ni-jxie, line 5: This is CPLD related, but not `rtfeatures` related. Maybe it does not belong here, but it does not feel like it belongs in `localextra.rules` either. I can be convinced to move it back to `localextra.rules`.

recipes-ni/ni-rtfeatures/ni-rtfeatures_1.0.bb
```diff
@@ -0,0 +1,62 @@
+SUMMARY = "rtfeatures user-space tools"
+DESCRIPTION = "Provides user-space tools to support rtfeatures."
```
amstewart: NIT: Maybe:
```DESCRIPTION = "Provides user-space tools to support the nirtfeatures kernel module."```

File: recipes-ni/ni-rtfeatures/ni-rtfeatures_1.0.bb [hunk 7]
```diff
@@ -0,0 +1,62 @@
...
+
+pkg_postrm_${PN} () {
+	if [ -n "$D" ]; then
+		OPT="-f -r $D"
+	else
+		OPT="-f"
+	fi
+
+	update-rc.d $OPT handle_cpld_ip_reset  remove
+}
```
amstewart: The `ni-hw-support-scritps*` recipes explicitly defined these functions because the OE-core `update-rc.d` bbclass doesn't support installing multiple initscripts within the same package (which the generic hw-support recipes needed to do).

Now that we have broken this feature out into its own recipe, I don't see us needed (or wanting) to install multiple initscripts, so lets prefer to use that bbclass instead.

recipes-ni/ni-rtfeatures/files/rtfeatures.rules
amstewart: Is it required for the other content in this package to work?

If not, I would agree that it would be better in a specific package; but maybe not this one?

recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+IP_RESET_FILE="/dev/nirtfeatures/ip_reset"
+
+IP_RESET_REQUESTED=""
+if [ -e "$IP_RESET_FILE" ]; then
+	grep -q '1' $IP_RESET_FILE && IP_RESET_REQUESTED="y"
+else
+	[ "${VERBOSE}" != "no" ] && echo "Unable to find ip_reset at normal spot."
```
amstewart: MINOR: I would prefer
```bash
[ "${verbose}" = yes ]
```
Since the variable is worded in the affirmative (as opposed to `NOT_QUIET`), we should interpret an unset `VERBOSE` as being `FALSE`.

recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
```diff
@@ -0,0 +1,16 @@
...
+fi
+
+if [ "$IP_RESET_REQUESTED" == "y" ]; then
+	mkdir -p /tmp/ip_reset_dir
+	/usr/local/natinst/bin/ninetcfgutil pulldefault -d /tmp/ip_reset_dir -g primary -g secondary
```
amstewart: I can't recall if `$PATH` is valid this early in boot. But if it is, I think we generally prefer to let `PATH` locate commands, in cases where the command binary isn't owned by the initscript package. Can you change this to just `ninetcfgutil ...`?

recipes-ni/ni-rtfeatures/files/rtfeatures.rules
billpittman: I generally group the CPLD and nirtfeatures together logically - so I don't have a problem moving this code into this recipe.

billpittman (approved):

shruthi-ravi (approved):
lgtm.

recipes-ni/ni-rtfeatures/files/rtfeatures.rules
ni-jxie: It is not required for rtfeatures to work, but keeping the grouping the CPLD and rtfeatures together makes sense.

recipes-ni/ni-rtfeatures/ni-rtfeatures_1.0.bb
ni-jxie: Done.

recipes-ni/ni-rtfeatures/ni-rtfeatures_1.0.bb
ni-jxie: That makes sense, inherit from `update-rc.d` instead.

recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
ni-jxie: Done.

recipes-ni/ni-rtfeatures/files/init.d/handle_cpld_ip_reset
ni-jxie: Relying on `$PATH` works, so update to call `ninetcfgutil ` instead.

amstewart (approved):
Updates all look good; thanks. :+1:

PR #350, opened 2022-03-03
linux-nilrt: Don't run postinst in safemode

chaitu236:
kernel package file's posinst fails when safemode is booted and /boot/runmode directory isn't present (which happens when runmode isn't installed), and opkg complains about this when installing new packages.

In any case, we shouldn't be running postinst in safemode. So return from postinst if in safemode.

### Testing
Added `[ -f /etc/natinst/safemode ] && exit 0` line to existing postinst on a safemode and verified running posinst manually doesn't fail like before.

### DIFF
--- a/recipes-kernel/linux/linux-nilrt.inc
+++ b/recipes-kernel/linux/linux-nilrt.inc
@@ -97,6 +97,11 @@ do_shared_workdir_append() {
 }
 
 pkg_postinst_ontarget_${KERNEL_PACKAGE_NAME}-base_append_x64 () {
+	# The postinst is intended to be run when new kernels are installed
+	# in runmode so that uninstalling them works correctly.
+	# It does not make sense to run this on safemode. So skip.
+	[ -f /etc/natinst/safemode ] && exit 0
+
 	cd "$D/${KERNEL_IMAGEDEST}"
 
 	if [ -f "${KERNEL_IMAGETYPE}" ]; then


recipes-kernel/linux/linux-nilrt.inc
```diff
@@ -97,6 +97,8 @@ do_shared_workdir_append() {
 }
 
 pkg_postinst_ontarget_${KERNEL_PACKAGE_NAME}-base_append_x64 () {
+	[ ! -f /etc/natinst/safemode  ] || exit 0
```
amstewart: Couple nits:
1. could you switch this to positive logic, for readability? eg.
    ```bash
    [ -f /etc/natinst/safemode ] && exit 0
    ```
2. Could you add a comment to the script explaining that the postinst is insane to run in safmode, so that future readers don't wonder why the code is getting skipped?

amstewart (changes requested):

recipes-kernel/linux/linux-nilrt.inc
chaitu236: Done

amstewart (approved):

shruthi-ravi (approved):

amstewart:
@chaitu236 Are you waiting for @gratian's review?

chaitu236:
> @chaitu236 Are you waiting for @gratian's review?

No, you can merge it.

PR #352, opened 2022-03-05
sysvinit-inittab: Bring back custom start_getty

chaitu236:
Serial console should be enabled/disabled based on the corresponding
setting in MAX.

These changes bring back the custom start_getty file that was previously
deleted. Some modifications were done to make it similar to OE's
start_getty.

### Testing
Built `sysvinit-inittab` recipe and verified the package contains these changes.

Manually copied `start_getty` from the package to a cRIO-9030 and verified that serial console output is enabled or disabled depending on the corresponding setting in MAX.

### DIFF
--- a/recipes-core/sysvinit/sysvinit-inittab/start_getty
+++ b/recipes-core/sysvinit/sysvinit-inittab/start_getty
@@ -1,0 +1,37 @@
+#!/bin/sh
+# Copyright (c) 2012-2013 National Instruments.
+# All rights reserved.
+# Spawn getty on the enabled, controlling console device. There will be at most
+# one of these. If there aren't any, don't spawn any.
+
+
+SPEED=""
+CONSOLE=""
+TERM=$3
+
+while read cons flags; do
+   case "$flags" in
+   *EC*) CONSOLE="$cons"
+         SPEED=$(stty speed < /dev/$CONSOLE)
+         echo Autodetected console "$CONSOLE" with baud rate "$SPEED"
+         break
+         ;;
+   esac
+done < /proc/consoles
+
+# If there aren't any detected consoles, exit
+[ -z "$CONSOLE" ] && exit 0
+
+# busybox' getty does this itself, util-linux' agetty needs extra help
+getty="/sbin/getty"
+case $(readlink -f "${getty}") in
+    */busybox*)
+        ;;
+    *)
+        if [ -x "/usr/bin/setsid" ] ; then
+            setsid="/usr/bin/setsid"
+        fi
+        ;;
+esac
+
+${setsid:-} ${getty} -L $SPEED $CONSOLE $TERM
--- a/recipes-core/sysvinit/sysvinit-inittab_2.%.bbappend
+++ b/recipes-core/sysvinit/sysvinit-inittab_2.%.bbappend
@@ -1,0 +1,3 @@
+# overwrite the standard OE start_getty with our ni-specific one; start_getty
+# is already added in SRC_URI so we just put our path first so OE fetches it.
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"


recipes-core/sysvinit/sysvinit-inittab/start_getty
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
```
shruthi-ravi, line 1: Can you mention the commit that deleted the file in the commit message? I'm curious to know why we deleted it before and why you are bringing it back now.

recipes-core/sysvinit/sysvinit-inittab/start_getty
```diff
@@ -0,0 +1,36 @@
...
+   *EC*) CONSOLE="$cons"
+         SPEED=$(stty speed < /dev/$CONSOLE)
+         echo Autodetected console "$CONSOLE" with baud rate "$SPEED"
+         break
+         ;;
+   esac
+done < /proc/consoles
+
+# If there aren't any detected consoles, exit
+[ -z "$CONSOLE" ] && exit 0
```
shruthi-ravi, line 23: It looks like the sumo version of the file defaults to tty2 in this case. Was that part no longer needed? It would've been nice to see the file as is resurrected, and then a commit on top of that modifying it.

recipes-core/sysvinit/sysvinit-inittab/start_getty
```diff
@@ -0,0 +1,36 @@
...
+    */busybox*)
+        ;;
+    *)
+        if [ -x "/usr/bin/setsid" ] ; then
+            setsid="/usr/bin/setsid"
+        fi
+        ;;
+esac
+
+${setsid:-} ${getty} -L $SPEED $CONSOLE $3
```
shruthi-ravi: What's in $3?

shruthi-ravi: Why are $1 and $2 not used if provided? Not sure if that's a supported use case though.

shruthi-ravi (commented):
Can you add the azdo link in the PR?

recipes-core/sysvinit/sysvinit-inittab/start_getty
chaitu236: Its https://github.com/ni/meta-nilrt/commit/f772bc9bad1ac9a49fa3f9cb22baab2ff011fedb. Added it to commit message. The testing at that time didn't uncover this issue and the code was thought to be unnecessary I believe.

recipes-core/sysvinit/sysvinit-inittab/start_getty
chaitu236: The commit that removed the file (https://github.com/ni/meta-nilrt/commit/f772bc9bad1ac9a49fa3f9cb22baab2ff011fedb) contained changes in another file (nilrt-runmode-system-image.postinst) that we don't want.

I could revert that commit and delete the change in that file .. but it seemed cleaner this way. Let me know if you prefer I do that.

recipes-core/sysvinit/sysvinit-inittab/start_getty
chaitu236: What arguments to call `start_getty` with are specified in [/etc/inittab](https://github.com/ni/openembedded-core/blob/nilrt/master/hardknott/meta/recipes-core/sysvinit/sysvinit-inittab_2.88dsf.bb#L32) - and for the [default script ](https://github.com/ni/openembedded-core/blob/nilrt/master/hardknott/meta/recipes-core/sysvinit/sysvinit-inittab/start_getty#L16) these arguments are used to invoke `/sbin/getty`.

In our case, we don't care about $2 and $1 which are the console name and its speed - this is because we are getting them from `/proc/consoles` (which is only set when console is enabled in MAX).

$3 is an optional arguement that specifies the terminal type and is set to `vt102` - we haven't used this in old version of `start_getty` but it seemed better to use it like upstream does.

recipes-core/sysvinit/sysvinit-inittab/start_getty
chaitu236: And regarding `tty2`, I haven't found anything that indicates its required - Bill also looked into this previously and [couldn't find](https://github.com/ni/meta-nilrt/pull/274#discussion_r751396054) anything about it either.

In console params we [only specify ttyS0](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-bsp/grub/grub/grub-runmode-bootimage.cfg#L3) so `tty2` doesn't even appear in `/proc/consoles` (`tty1` does appear in `/proc/consoles` even when not specified in console params but I think thats the default terminal that appears on a display connected to target).

recipes-core/sysvinit/sysvinit-inittab/start_getty
shruthi-ravi: Ok. Can you set $3 to a variable before using it? A variable name that explains what it is, for readability.

shruthi-ravi (approved):

recipes-core/sysvinit/sysvinit-inittab/start_getty
chaitu236: Ok, done

recipes-core/sysvinit/sysvinit-inittab/start_getty
chaitu236: I just noticed that `/etc/inittab` line for serial console specifies `respawn` which means `start_getty` will be restarted if it is accidentally killed (which I guess is what we want). So in the case when console is disabled in MAX, if we return early from the script, it will be restarted continuously and we end up with following log in `/var/log/messages`

> init: Id "S0" respawning too fast: disabled for 5 minutes

I suspect `tty2` was present previously to prevent this. I'll add it back.

amstewart (approved):

recipes-core/sysvinit/sysvinit-inittab/start_getty
chaitu236: Addressed this in https://github.com/ni/meta-nilrt/pull/353

PR #355, opened 2022-03-11
Strip the boot-attestation (TPM) components from NILRT hardknott

ni-jxie:
### Description

The boot-attestation (TPM) components from NILRT is dead code at this point because it is based on a pre-release fork of TPM2-tools and no longer being maintained. Strip all of it from NILRT so that we don't carry these dead code forward.

### Implementation

Strip boot-attestation logic from meta-nilrt layer.

### Testing

Provision a VM with tpm stuff removed, it still boots and everything seems to still work.


ni-jxie:
Should we test this on real hardware? Looking through the logic, it seems like if there is no TPM chip, a VM is just as good as real hardware.

bstreiff:
QEMU has support for virtualized TPMs which may be useful for testing. https://qemu.readthedocs.io/en/latest/specs/tpm.html https://en.opensuse.org/Software_TPM_Emulator_For_QEMU

(TPM is necessary for Win11 so there's a bunch of guides available if you look for that set of keywords)

amstewart:
I would prefer that we test on real hardware, rather than emulating a TPM with QEMU - simply because we haven't been testing with an emulated TPM thus far, and so I don't think we can be confident that the results are accurate.

recipes-extended/sudo/sudo_%.bbappend
```diff
@@ -3,8 +3,5 @@ do_install_append() {
 	if ${@oe.utils.conditional('DISTRO', 'nilrt-nxg', 'false', 'true', d)}; then
 		echo >>"${D}${sysconfdir}/sudoers" ""
 		echo >>"${D}${sysconfdir}/sudoers" "User_Alias TPM_QUOTERS = lvuser, webserv"
```
amstewart: I see you remove this whole file later in the patchset. Could you just incorporate the later commit as a fixup to the first?

recipes-ni/initscripts-nilrt/files/run-ptest
```diff
@@ -5,9 +5,4 @@
 ./test-nisetcommitratio-unit
 ./test-nisetcommitratio-system
 
-if [ "`uname -m`" == "x86_64" ]; then
-    ./test-niopendisks-init
-    ./test-niclosedisks-init
```
amstewart, line 10: >initscripts-nilrt: remove niclosedisks + niopendisks and related ptests 

Could you add a commit summary which makes it clear that these scripts served the measured boot feature, and are no longer needed? It isn't obvious from the script names.

recipes-core/initrdscripts/files/ni_provisioning.safemode
```diff
@@ -425,7 +406,6 @@ sanity_check()
 }
 
 check_all_used_binaries
-load_kernel_modules
```
amstewart, line 428: Did you remove this function because the provisioning scripts only need to load the `tpm_tis` module if they are going to setup measured boot? If so, can you make that clear in the commit message?

Otherwise, I'm not sure we can safely remove this part. It seems reasonable to me that we would load the `tpm_tis` module on devices which have a TPM (eg. TigerLake and other PXIe controllers).

recipes-ni/ni-systemformat/files/nisystemformat
```diff
@@ -49,14 +42,13 @@ SYSCFG_IMAGE_TGZ_FILE=$NIBOOT_DIR/system-config-image.tgz
...
-	[INVALID_ARCH]=6        # Running an unsupported processor architecture
-	[NO_TPM_HARDWARE]=7     # Could not detect a TPM chip
-	[NO_KEY_BACKUP_DEVICE]=8     # Partition to store backup key could not be found; perhaps it's mislabeled or not plugged in
+	[UNSPECIFIED]=1				# No reason specified
+	[UNKNOWN_ERROR]=2			# Unexpected exit code encountered
+	[INVALID_ARGUMENT]=3		# An invalid combination of arguments was specified
+	[INVALID_FSTYPE]=4			# An invalid fstype was specified
+	[ILLTIMED]=5            	# Called at an inappropriate time
+	[INVALID_ARCH]=6        	# Running an unsupported processor architecture
+	[NO_KEY_BACKUP_DEVICE]=7    # Partition to store backup key could not be found; perhaps it's mislabeled or not plugged in
```
amstewart: NIT: I think you used tab characters to align the comments here, and it is making it look weird and out-of-alignment. Consider using spaces instead.

recipes-ni/ni-systemformat/files/nisystemformat
```diff
@@ -252,8 +238,6 @@ usage()
 		-t	Filesystem type
 		-c	Format the config volume (otherwise rootfs + kernel)
 		-r	Relaunch the system webserver after format
-		-e	Encrypt volume using a new generated disk key stored in the TPM
-		-B	Backup encryption keys to the filesystem labeled <fslabel> (normally on USB storage)
 		-P	Comma separated set of PCR registers to seal disk keys
 		-L	Specifies dictionary lockout configuration:
 		        keep: Do not change current configuration
```
amstewart: You removed the `-P` and `-L` arguments from `getopts` (which seems reasonable), but they're still in the usage text.

recipes-bsp/grub/grub-efi_git.bbappend
```diff
@@ -2,7 +2,7 @@ require grub-nilrt.inc
 
 GRUB_BUILDIN += "smbios chain multiboot efi_uga font gfxterm gfxmenu terminal \
                 minicmd iorw echo reboot terminfo loopback memdisk tar help serial \
-                ls search_fs_uuid udf btrfs ntfs reiserfs xfs lvm ata tpm \
+                ls search_fs_uuid udf btrfs ntfs reiserfs xfs lvm ata \
```
amstewart, line 5: Is this specific to the measured boot feature? I imagine that we might want this `tpm` option in general.

recipes-core/packagegroups/packagegroup-ni-restoremode.bb
```diff
@@ -40,7 +40,6 @@ RDEPENDS_${PN} += "\
 "
 
 RRECOMMENDS_${PN}_x64 = "\
-	kernel-module-tpm-tis \
```
amstewart: I think the `tpm_tis` module is generally useful for targets which have TPMs (like PXIs) do we really want to remove it in all cases? 

@gratian ?

recipes-ni/initscripts-nilrt/files/populateconfig
```diff
@@ -25,7 +25,6 @@ then
 		fi
 	fi
 
-	mkdir -p /etc/natinst/share/tpm
```
amstewart, line 28: Can you do a codesearch and attempt to identify if this path is used by NI packages?

amstewart (changes requested):
Overall; the content looks good, pending some suggestions below.

I know its a pain, but could you take a pass through the commit messages and make sure that we're documenting the justification for each change. It isn't obvious in a lot of cases, and I expect that these commits are going to show up in future debugging.

recipes-ni/initscripts-nilrt/files/populateconfig
ni-jxie: It is only used by nilrtdiskcrypt.

recipes-extended/sudo/sudo_%.bbappend
ni-jxie: Done.

recipes-ni/initscripts-nilrt/files/run-ptest
ni-jxie: Done.

recipes-core/initrdscripts/files/ni_provisioning.safemode
ni-jxie: Yeah, I was assuming that the `tpm_tis` kernel module is only needed if we actually using the TPM chip. Even if the PXIe controllers does have TPM but if we don't support boot attestation, do we still need the kernel module? I am going to run some test on real hardware to see what happens.

recipes-bsp/grub/grub-efi_git.bbappend
ni-jxie: It was added in the same time period by Haris in Penguin CL:6218826 when we added the measured boot feature, but there was no explanation in the CL.

recipes-ni/ni-systemformat/files/nisystemformat
ni-jxie: Fixed.

recipes-ni/ni-systemformat/files/nisystemformat
ni-jxie: Fixed.

billpittman (approved):

recipes-core/packagegroups/packagegroup-ni-restoremode.bb
billpittman: I think keeping the module in would probably be better in the long run in case someone decides they want to use it.  I also don't imagine that its going to impact the RT nature of the product, so keeping it in seems safe.  We support using the TPM and an rng source, (CONFIG_HW_RANDOM_TPM), so it would be consistent to leave it in.

gratian: Yes, I agree let's keep the module (for the reasons Alex and Bill enumerated above). It's pretty small and it might be useful to somebody.

amstewart (approved):
@ni-jxie When you do a force-push; please create a patch-revision comment outlining what is different between this patch-rev and the last - even if you just addressed reviewer feedback.

amstewart:
@ni-jxie Because this is a large patchset, I'm going to wait for cjohnston's meta layer deprecation PRs to pass build before pulling this PR on its own. I expect that means I'll pull this later today or early tomorrow.

PR #362, opened 2022-03-21
Remove meta-measured as it is dead upstream

ni-jxie:
### Description

meta-measured layer is no longer being maintained upstream. Migrate the strongswan bbappend from meta-measured into meta-nilrt and remove meta-measured;

[Technical Debt 1564625](https://ni.visualstudio.com/DevCentral/_workitems/edit/1564625)

### Implementation
- Migrate strongswan and trousser to meta-nilrt
- Remove meta-measured from meta-nilrt

### Testing
Pending

### DIFF
--- a/recipes-support/strongswan/strongswan_5.%.bbappend
+++ b/recipes-support/strongswan/strongswan_5.%.bbappend
@@ -3,7 +3,54 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 SRC_URI += "file://ipsec.init \
 "
 
-PACKAGECONFIG += " eap-identity eap-mschapv2"
+PACKAGECONFIG += " \
+    aikgen \
+    eap-identity \
+    eap-mschapv2 \
+    imc-test \
+    imv-test \
+    imc-scanner \
+    imv-scanner \
+    imc-os \
+    imv-os \
+    imc-attestation \
+    imv-attestation \
+    imc-swima \
+    imv-swima \
+    tnc-ifmap \
+    tnc-imc \
+    tnc-imv \
+    tnc-pdp \
+    tnccs-11 \
+    tnccs-20 \
+    tnccs-dynamic \
+    tss-trousers \
+    "
+
+PACKAGECONFIG[aikgen] = "--enable-aikgen,--disable-aikgen,,"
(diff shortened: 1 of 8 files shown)


recipes-support/strongswan/strongswan_5.%.bbappend
```diff
@@ -3,7 +3,54 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 SRC_URI += "file://ipsec.init \
 "
```
shruthi-ravi, line 4: The commit message makes it sound like the strongswan recipe was owned by meta-measured. But it looks like the actual recipe is part of meta-openembedded, and meta-measured only added to it (using a bbappend). Can you modify the message to say that you migrated the strongswan bbappend instead?

recipes-support/trousers/trousers_0.3.15.bb
```diff
@@ -0,0 +1,8 @@
+# 0.3.14 unpacks to the current directory instead of to a sub-directory
```
shruthi-ravi: Is this version-specific comment still needed?

recipes-support/trousers/trousers_0.3.15.bb
```diff
@@ -0,0 +1,8 @@
...
+
+S = "${WORKDIR}/trousers-0.3.15"
```
shruthi-ravi: You can use the recipe variable here. I think they are `PN` and `PV`.

amstewart (approved):

recipes-support/strongswan/strongswan_5.%.bbappend
ni-jxie: Done.

ni-jxie:
### PATCH V1

- Remove unneeded version specific comment.
- Update trousers bb to use recipe variable.

recipes-support/trousers/trousers_0.3.15.bb
ni-jxie: Removed

recipes-support/trousers/trousers_0.3.15.bb
ni-jxie: Modified to use `PN` and `PV` instead.

shruthi-ravi (approved):

PR #364, opened 2022-03-22
cherry-pick: remove meta-measured as it is dead upstream

ni-jxie:
Cherry-pick into the release branch.

### DIFF
--- a/conf/bblayers.conf.sample
+++ b/conf/bblayers.conf.sample
@@ -8,7 +8,6 @@ BBFILES ?= ""
 BBLAYERS ?= " \
 	${GIT_REPODIR}/meta-cloud-services \
 	${GIT_REPODIR}/meta-cloud-services/meta-openstack \
-	${GIT_REPODIR}/meta-measured \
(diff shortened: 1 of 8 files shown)


amstewart:
The `nilrt/21.8/hardknott` branch was forked prior to merging #355. So I think you will have to cherry-pick that patchset first.

@shruthi-ravi

ni-jxie:
> 

Gotcha, #365

shruthi-ravi:
Would a merge be easier if we want all the changes? @amstewart

ni-jxie:
> > 
> 
> Gotcha, #365

amstewart:
@shruthi-ravi Conceptually, I think our intention is to scrutinize PRs which are submitted against the release branches while we perform go-final. If we're attempting to exercise that workflow, then I think we need to hold ourselves to submitting cherry-pick PRs for content which needs to build during that period. So while it would be easier for you or I to submit a merge, I think it would be violating our own policies.

PR #370, opened 2022-03-29
Making kdump work for nilrt

ni-jxie:
### Description
It would be nice if we get crashkernel working just by installing a few packages, this PR tries to achieve that.

1. `opkg install packagegroup-ni-debug-kernel`
2. `opkg install kexec kexec-tools makedumpfile kdump`
3. Reboot
4. `cat /sys/kernel/kexec_crash_loaded`, make sure it is 1
5. `echo c > /proc/sysrq-trigger` to trigger a crash
6. The dumpfile should be in `/var/crash/`, locate it. The OE crash does not process the dumpfile correctly. So you need to compile crash 8.0.0++ from source. Then do `crash vmcore vmlinux` to analyze the log.

### Implementation
Make nilrt specific modification to kdump.conf.

### Testing
It works on a 9049 4 GB RAM running Hardknott image, the compressed dumpfile is 3.4 GB.

### DIFF
--- a/recipes-kernel/kexec/kexec-tools/kdump.conf
+++ b/recipes-kernel/kexec/kexec-tools/kdump.conf
@@ -1,0 +1,19 @@
+#the kdump kernel version string.
+KDUMP_KVER="`uname -r`"
+
+#this will be passed to the kdump kernel as kdump kernel command line
+#only go to runlevel 3 to make boot faster
+#blacklist nikal so no NI driver module would be loaded
+KDUMP_CMDLINE="`cat /proc/cmdline` 3 usbcore.nousb nomodeset module_blacklist=nikal"
+
+#the kernel image for kdump
+KDUMP_KIMAGE="/boot/runmode/bzImage"
+
+#Where to save the vmcore
+KDUMP_VMCORE_PATH="/var/crash/`date +"%Y-%m-%d"`"
+
+#OE makdumpfile in OE cannot read the vmlinux image when booted in crashkernel
+# due a dwarf error so do not pass in the vmlinux and only compresss the image
+# reduce the image size by about (10~20%)
+#the arguments to makedumpfile
+MAKEDUMPFILE_ARGS="-c -D"
--- a/recipes-kernel/kexec/kexec-tools_2.0.21.bbappend
+++ b/recipes-kernel/kexec/kexec-tools_2.0.21.bbappend
@@ -1,0 +1,9 @@
+SUMMARY = "nilrt specific modification for kdump"
+FILESEXTRAPATHS_prepend := "${THISDIR}/kexec-tools:"
+
+SRC_URI += " file://kdump.conf \
+           "
+
+pkg_postinst_kdump() {
+    sed -i 's/set otherbootargs="\(.*\)"/set otherbootargs="\1 crashkernel=128M "/' /boot/runmode/bootimage.cfg
+}


recipes-kernel/kexec/kexec-tools/kdump.conf
```diff
@@ -0,0 +1,19 @@
+#the kdump kernel version string.
+KDUMP_KVER="`uname -r`"
+
+#this will be passed to the kdump kernel as kdump kernel command line
+#only go to runlevel 3 with nonet nousb nosound nomodeset to make boot faster
+#blacklist nikal so no NI driver module would be loaded
+KDUMP_CMDLINE="`cat /proc/cmdline` 3 nonet nousb nosound nomodeset module_blacklist=nikal"
```
ni-jxie, line 7: nonet nousb nosound and nomodeset did not actually do anything @gratian ?

recipes-kernel/kexec/kexec-tools_2.0.21.bbappend
```diff
@@ -0,0 +1,11 @@
+SUMMARY = "nilrt specific modification for kdump"
+FILESEXTRAPATHS_prepend := "${THISDIR}/kexec-tools:"
+
+SRC_URI += " file://kdump.conf \
+           "
+
+pkg_postinst_kdump() {
+#!/bin/sh
+
+sed -i 's/set otherbootargs="\(.*\)"/set otherbootargs="\1 crashkernel=128M "/' /boot/runmode/bootimage.cfg
```
ni-jxie, line 10: Redhat has a guideline for crashkernel reserved memory size: https://access.redhat.com/solutions/59432. But 120 MB seems to be enough for NI Linux RT targets with 4 GB of RAM. Might need to bump this for large memory targets.

amstewart:
>The OE crash does not process the dumpfile correctly. So you need to compile crash 8.0.0++ from source.

Is `crash` the only tool (or the only recommended) tool for evaluating those crash dumps? If so, we should probably prioritize fixing it for users.

ni-jxie:
Yeah, `crash` seems to be the only recommended tool, according to Debian, Ubuntu, Fedora and Redhat documentation. The next step would be to make the `crash` package from our feed working out of the box by upgrading it (maybe OE upstream has a newer one).

amstewart:
Would you mind filing a user story to do that?

amstewart (approved):

recipes-kernel/kexec/kexec-tools/kdump.conf
bstreiff: `nonet` and `nosound` don't exist in https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html. `nousb` might be `usbcore.nousb`. `nomodeset` should still be operational-- if you have that on you _should_ only get EFI framebuffer instead of i915, I think? But it's possible that those others might have been removed in earlier kernel versions.

bstreiff:
One of the observations I'd made about `makedumpfile` the last time I'd tried this was that it needs to have support for the kernel in question. For instance, with sumo, the `makedumpfile` there only supported up to 4.8.1, but we were on 4.14. Presumably we're not hitting that now because hardknott's `makedumpfile` is newer, but this might need to be a thing we need to check for as we keep updating kernel versions.

I don't know what the best way to verify that is though. The `makedumpfile` source has a `makedumpfile.h` with a `LATEST_VERSION` constant, so maybe a ptest there? Otherwise we'd need a whole crash-the-kernel test setup. that's probably out-of-scope for this PR though.

ni-jxie:
### PATCH V1

Update kernel boot parameter to drop the deprecated ones and use the more up-to-date ones

recipes-kernel/kexec/kexec-tools/kdump.conf
ni-jxie: Ah, updated documentation! Thanks, drop `nousb` ,`nonet`, and `nosound`. Add `usbcore.nousb` and keep `nomodeset`.

ni-jxie:
> Would you mind filing a user story to do that?

[Technical Debt 1904479](https://ni.visualstudio.com/DevCentral/_workitems/edit/1904479): Uprade crash in OE

ni-jxie:
> One of the observations I'd made about `makedumpfile` the last time I'd tried this was that it needs to have support for the kernel in question. For instance, with sumo, the `makedumpfile` there only supported up to 4.8.1, but we were on 4.14. Presumably we're not hitting that now because hardknott's `makedumpfile` is newer, but this might need to be a thing we need to check for as we keep updating kernel versions.
> 
> I don't know what the best way to verify that is though. The `makedumpfile` source has a `makedumpfile.h` with a `LATEST_VERSION` constant, so maybe a ptest there? Otherwise we'd need a whole crash-the-kernel test setup. that's probably out-of-scope for this PR though.

Excellent idea: [Technical Debt 1904487](https://ni.visualstudio.com/DevCentral/_workitems/edit/1904487): Add a ptest for makedumpfile to make sure that it works with the current kernel for nilrt

recipes-kernel/kexec/kexec-tools_2.0.21.bbappend
```diff
@@ -0,0 +1,11 @@
+SUMMARY = "nilrt specific modification for kdump"
+FILESEXTRAPATHS_prepend := "${THISDIR}/kexec-tools:"
+
+SRC_URI += " file://kdump.conf \
+           "
+
+pkg_postinst_kdump() {
+#!/bin/sh
```
bstreiff: I don't think this shebang line is used at all

ni-jxie:
### PATCH V2
-----------------------------------------------------------------------------------------------------------------------------------------------
Remove shebang line from pkg_postinst_kdump

recipes-kernel/kexec/kexec-tools_2.0.21.bbappend
ni-jxie: Removed.

gratian (approved):

bstreiff (approved):

PR #373, opened 2022-04-05
makedumpfile: add a ptest to check compatibility with running kernel

bstreiff:
makedumpfile needs to know some things about internal kernel structures,
which means that it has a supported kernel version range. Since the nilrt
layer uses a different (and sometimes newer) kernel version than the OE
upstream on which it is based, add a ptest so that we can have a check to
see if our current version of makedumpfile works on a kernel upgrade.

Fixes AzDO#1904487.

### DIFF
--- a/recipes-kernel/makedumpfile/files/makedumpfile-is-kernel-supported.c
+++ b/recipes-kernel/makedumpfile/files/makedumpfile-is-kernel-supported.c
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * makedumpfile-is-kernel-supported.c
+ *
+ * makedumpfile has a supported kernel version range, but doesn't expose a
+ * way to check the current kernel version against it... so this simple
+ * executable sucks in the header file in order to do so.
+ */
+
+#include <errno.h>
+#include <limits.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/utsname.h>
+
+// get KVER_MAJ_SHIFT, KVER_MIN_SHIFT, KERNEL_VERSION, OLDEST_VERSION, LATEST_VERSION:
+#include "makedumpfile.h"
+
+// inverse of the KERNEL_VERSION() macro
+#define VERSION_FMT "%d.%d.%d"
+#define VERSION_SPLIT(x) \
+	((x) >> KVER_MAJ_SHIFT) & ((1 << (32 - KVER_MAJ_SHIFT)) - 1), \
+	((x) >> KVER_MIN_SHIFT) & ((1 << (KVER_MAJ_SHIFT - KVER_MIN_SHIFT)) - 1), \
+	(x) & ((1 << KVER_MIN_SHIFT) - 1)
+
+#define PTEST_NAME "makedumpfile-is-kernel-supported"
+
+void ptest_pass() {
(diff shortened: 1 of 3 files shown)


bstreiff:
@ni/rtos

also spoilers: makedumpfile 1.6.8 in hardknott only claims compatibility up to linux 5.9.4, so this ptest fails at present.
```
BEGIN: /usr/lib/makedumpfile/ptest
FAIL: kernel version 5.10.100 not within supported range [2.6.15 .. 5.9.4]
```

recipes-kernel/makedumpfile/files/makedumpfile-is-kernel-supported.c
```diff
@@ -0,0 +1,81 @@
...
+
+int main(int argc, char *argv[])
+{
+	struct utsname utsname;
+	uint32_t version;
+	int err;
+
+	if (uname(&utsname)) {
+		err = errno;
+		fprintf(stderr, "FAIL: uname() failed to retrieve kernel information: %s\n",
```
amstewart: The string after `FAIL:` on a ptest status output line needs to contain the immutable test name and (optionally) subtest name; it can't contain descriptive information about the failure. You can print descriptive information on lines above or below the status line. eg.

```
ERROR: uname() failed to retrieve kernel information:...
FAIL: makedumpfile-is-kernel-supported
```

https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/3560/Yocto-PTests?anchor=test-result-lines

recipes-kernel/makedumpfile/files/makedumpfile-is-kernel-supported.c
```diff
@@ -0,0 +1,81 @@
...
+	}
+	
+	if ((version < OLDEST_VERSION) || (LATEST_VERSION < version)) {
+		fprintf(stderr, "FAIL: kernel version " VERSION_FMT " not within supported range [" VERSION_FMT " .. " VERSION_FMT "]\n",
+			VERSION_SPLIT(version),	VERSION_SPLIT(OLDEST_VERSION),
+			VERSION_SPLIT(LATEST_VERSION));
+		exit(ENOTSUP);
+	}
+
+	exit(0);
```
amstewart: I've found that most ptest-consuming frameworks don't care about test return codes, even if they should. So you generally always have to output a status line, even if the test passes. In addition, the test_name on the pass line must match with the string used on the fail lines. So I think should add here:

```
fprintf(stderr, "PASS: %s\n", "makedumpfile-is-kernel-supported")
```

https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/3560/Yocto-PTests?anchor=return-codes

amstewart (approved):

bstreiff:
v2 changes:
- if we're gonna mash together stdout/stderr anyway, just use `printf` throughout
- add a `PASS:` report, replace `FAIL:` lines with `ERROR:` and add a `FAIL:` report with test name

Output:
```
START: ptest-runner
2022-04-05T22:32
BEGIN: /usr/lib/makedumpfile/ptest
ERROR: kernel version 5.10.100 not within supported range [2.6.15 .. 5.9.4]
FAIL: makedumpfile-is-kernel-supported

ERROR: Exit status is 256
END: /usr/lib/makedumpfile/ptest
2022-04-05T22:32
STOP: ptest-runner
```

amstewart (approved):

gratian (approved):

PR #380, opened 2022-04-15
packagegroup-ni-internal-deps: Add rdma-core

bstreiff:
libnirdma requires rdma-core as a dependency; add it to the core feeds via packagegroup-ni-internal-deps.

Resolves AzDO#1926845.

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
@@ -33,3 +33,9 @@ RDEPENDS_${PN} += "\
 RDEPENDS_${PN}_append_x64 = "\
 	qtbase \
 "
+
+# ni-rdma, libnirdma
+# Contact: Eric Gross <eric.gross@ni.com>
+RDEPENDS_${PN} += "\
+	rdma-core \
+"


ericgross2 (approved):

recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
```diff
@@ -33,3 +33,8 @@ RDEPENDS_${PN} += "\
 RDEPENDS_${PN}_append_x64 = "\
 	qtbase \
 "
+
+# ni-rdma, libnirdma
```
amstewart, line 37: Can you add a technical contact here - someone who can speak to the dependency?

amstewart (changes requested):

recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
bstreiff: Sure. @ericgross2 just approved this review so I'm taking that as him volunteering to be that contact.

bstreiff: added in v2

amstewart:
@ericgross2 Do you agree with being the technical contact?

ericgross2:
@amstewart sure, I don't mind

amstewart (approved):

PR #381, opened 2022-04-18
ni-grpc-device: use modern recipe version

amstewart:
When we merged the `sumo` and `dunfell` meta-nilrt mainline refs to make `hardknott`, we accidentally incorporated both the *static-binary* and *source-build* versions of the `ni-grpc-device` recipe. The static-binary implementation was a hack, necessary to get `grpc-device` into sumo. The source-build version should be preferred in all cases. But because the static-binary version recipe file was strongly versioned (`_1.1.bb`), it was preferred by bitbake.

This PR removes the static-binary recipe in favor of the source-build version. It also fixes the do_configuration() task in the recipe - which is currently failing due to cmake not being able to find the gtest dependency.

# Testing
Recipe builds on my machine.


chaitu236 (approved):

recipes-ni/ni-grpc-device/files/run-ptest
```diff
@@ -1,43 +0,0 @@
-#!/bin/bash
```
chaitu236, line 1: Isn't the ptest useful anymore?

amstewart: The competing recipe versions used two different paths for the `run-ptest` file. The modern version uses the file from [here](https://github.com/ni/meta-nilrt/blob/b4f33d2bbe02c9f20847e1fb36c67a010d9b5084/recipes-ni/ni-grpc-device/files/ptest/run-ptest), which is identical to the version one-level-above, used by the static-binary recipe version.

The GH PR doesn't do a great job of showing that I have dropped one and not the other.

PR #383, opened 2022-04-20
sysconfig-settings: Add systemsettings directory

billpittman:
Create shared systemsettings directory with expected permissions
and ownership.

### Testing
```
0 ✓ bill@ravine /ssd/hardknott/nilrt/build $ ar -p tmp-glibc/deploy/ipk/core2-64/sysconfig-settings_1.0-*.ipk ./data.tar.xz |tar tJv |grep var
drwxr-xr-x root/root         0 2022-04-20 14:03 ./var/
drwxr-xr-x root/root         0 2022-04-20 14:03 ./var/local/
drwxrwxr-x lvuser/ni         0 2022-04-20 14:03 ./var/local/natinst/
drwxrwxr-x lvuser/ni         0 2022-04-20 14:03 ./var/local/natinst/systemsettings/

```

### DIFF
--- a/recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
+++ b/recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
@@ -48,9 +48,10 @@ FILES_${PN} = "${settingsdatadir}/consoleout.ini \
                ${uixmldir}/nilinuxrt.rtprotocol_enable.* \
                ${uixmldir}/nilinuxrt.rtapp_disable.* \
                ${uixmldir}/nilinuxrt.fpga_disable.* \
+               ${systemsettingsdir} \
 "
 
-DEPENDS += "shadow-native pseudo-native niacctbase"
+DEPENDS += "shadow-native pseudo-native niacctbase base-files-nilrt"
 RDEPENDS_${PN} += "niacctbase bash fw-printenv"
 
 # sysconfig-settings-ssh package
@@ -124,8 +125,10 @@ do_install () {
 
 	# Common interface for system settings (soft dip switches, etc.)
 	install -d -m 0775 ${D}${settingsdatadir}/
-	chown ${LVRT_USER}:${LVRT_GROUP} ${D}${settingsdatadir}/
 	install -m 0644 ${S}/systemsettings/* ${D}${settingsdatadir}/
+
+	# Create shared systemsettingsdir with appropriate permissions and ownership
+	install -d -m 0775 -o ${LVRT_USER} -g ${LVRT_GROUP} ${D}${systemsettingsdir}
 
 	install -d ${D}${sysconfdir}/init.d/
 	install -m 0755 ${WORKDIR}/nisetembeddeduixml ${D}${sysconfdir}/init.d


recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -127,6 +129,12 @@ do_install () {
 	chown ${LVRT_USER}:${LVRT_GROUP} ${D}${settingsdatadir}/
 	install -m 0644 ${S}/systemsettings/* ${D}${settingsdatadir}/
 
+	# Create shared systemsettingsdir with appropriate permissions and ownership
+	install -d -m 0775 ${D}/${systemsettingsbasedir}
+	chown ${LVRT_USER}:${LVRT_GROUP} ${D}/${systemsettingsbasedir}
```
amstewart: Since `/var/local/natinst` isn't owned by this package, I'm not sure we care to set the ownership here. Presumably, that directory is owned by [`base-files-nilrt`](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-ni/base-files-nilrt/base-files-nilrt_1.0.bb#L28), which this package should depend upon. So it should already exist and have whatever permissions are defined in `base-files-nilrt` when this IPK is unpacked.

amstewart (changes requested):

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
billpittman: Fair enough, I added the additional DEPEND and changed the recipe to use the -o and -g flags instead of chown-ing the directory after the fact.

amstewart: I don't think you need the `-o` or `-g` at all, but this is fine.

recipes-ni/sysconfig-settings/sysconfig-settings_1.0.bb
```diff
@@ -123,10 +124,12 @@ do_install () {
 	install -m 0644 ${S}/uixml/* ${D}${uixmldir}/
 
 	# Common interface for system settings (soft dip switches, etc.)
-	install -d -m 0775 ${D}${settingsdatadir}/
-	chown ${LVRT_USER}:${LVRT_GROUP} ${D}${settingsdatadir}/
+	install -d -m 0775 -o ${LVRT_USER} -g ${LVRT_GROUP} ${D}${settingsdatadir}/
```
bstreiff: I don't think this directory needs to be set ${LVRT_USER}:${LVRT_GROUP} at all-- it's package-local to us, and I don't believe lvuser ever needs to modify the contents.

billpittman: I'm honestly fine either way, but the sumo version was lvuser:ni and the original bug was that it had changed in hardknott.  I'm not sure what other packages get installed and need to read/write files in this directory, but if you're sure that 0:0 would work, I can change it.  The question is ... are you sure?

bstreiff: So there are two directories in play here.

    settingsdatadir = "${datadir}/${BPN}/systemsettings"
    systemsettingsdir = "${localstatedir}/local/natinst/systemsettings"

the first one did not exist in sumo, it was added in 633c3f2e072e.

(the latter one did and should probably be lvuser:ni for consistency)

billpittman: Oh sorry missed your point on that one.  Yea I'm totally fine removing the lvuser:ni on settingsdatadir.

bstreiff (approved):

amstewart (approved):

PR #396, opened 2022-05-24
[sumo] pkggrp-ni-runmode: remove tbtadm

amstewart:
When building the zynq safemode rootfs in the sumo mainline - only on the build machines - the opkg status will become contamindated with an entry for `tbtadm-lic`. It's status is `deinstall hold not-installed`, and its `VERSION` is set to an error string, apparently from the build machine environment.

This error is consistent on the build machines, but I cannot reproduce it outside of that environment, and it never affects the other rootfses. I suspect it may be related to [this opkg operation](https://github.com/ni/meta-nilrt/blob/840f2e5a38b579c26d6a585d917e8bef665a0d5b/recipes-core/images/niconsole-image-safe.bb#L17), but I don't have an easy way to test that theory on the build machines.

However, thunderbolt isn't actually used on any of the currently supported zynq targets. `tbtadm` was added in [this commit](https://github.com/ni/meta-nilrt/commit/e2653c9d05383b1b95ed3437a61076b408c14737) for PXI support. Since sumo only supports ARM, we can safely remove this package.

Natinst AZDO bug: [Bug 1956754](https://dev.azure.com/ni/DevCentral/_workitems/edit/1956754)

# Testing
* Rebuilt the `zynq*` make targets on my dev machine and confirmed that there are no references to `tbtadm` either in the opkg status file, or in the IPK feed.

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-extra.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-extra.bb
@@ -266,10 +266,6 @@ RDEPENDS_${PN} += "\
 "
 
 
-# meta-openembedded/meta-oe/recipes-bsp
-RDEPENDS_${PN} += "\
-	tbtadm \
-"
 # meta-openembedded/meta-oe/recipes-connectivity
 RDEPENDS_${PN} += "\
 	gammu \
--- a/recipes-core/packagegroups/packagegroup-ni-runmode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-runmode.bb
@@ -28,7 +28,6 @@ RDEPENDS_${PN} = "\
 	parted \
 	rtctl \
 	systemimageupdateinfo \
-	tbtadm \
 	util-linux-sfdisk \
 	vlan \
 	zip \


chaitu236:
Seeing that we're not sure why the problem occurs, is it possible to add some test to rootfs stage to detect this if it happens again?

amstewart:
Functionally, we could. But I'm not sure what we would test for exactly. I don't really want to get into the business of verifying the *correctness* of the `status` file fields - since even the `VERSION` string can fit so many different valid formats.

The opkg status of the system is functionally tested in different forms when (a) opkg installs the rootfs, (b) we do another opkg install at BSI installation, and (c) whenever MAX does a package install. I think those are all the tests that we would want to maintain - and they caught this issue.

chaitu236:
> since even the VERSION string can fit so many different valid formats.

Yes, it may not be straight forward but maybe we can check for strings such as "failed", "error", etc.. or maybe anything space separated?

> I don't really want to get into the business of verifying the correctness of the status file fields

Yes, it is not ideal but I don't like not being able to detect this issue if it happens in future.

> The opkg status of the system is functionally tested in different forms when ..

Can you clarify this? We only caught this because MAX crashed on a specific target, but once that is fixed in MAX to not crash on malformed versions, we won't be able to detect the issue - this is the case in other ARM targets right now where MAX doesn't crash.

shruthi-ravi (approved):
Pending test resolution with chaitu236.

PR #401, opened 2022-06-03
nilrt-grub-runmode: fix up spectre/meltdown mitigation cfgs

bstreiff:
The sumo release offered grub configuration fragments for toggling various spectre/meltdown-related kernel command line arguments, because some mitigations have performance impacts that could affect existing customer applications.

We have a documented interface for toggling these via nirtcfg; however, the nirtcfg implementation is hardcoded to expect the configurations at paths that are not under bootimage.cfg.d. Additionally, the fragments under bootimage.cfg.d were only being properly sourced with a grub that supports wildcard expansion, which some older provisioned targets do not have.

In order to handle both of these issues, move the configuration fragments back into the /boot/runmode directory directly, and also restore explicitly sourcing those fragments by name in bootimage.cfg instead of using wildcards for compatibility with older provisioning. This gets us back to feature parity with sumo.

### DIFF
--- a/recipes-bsp/grub/grub/grub-runmode-bootimage.cfg
+++ b/recipes-bsp/grub/grub/grub-runmode-bootimage.cfg
@@ -4,8 +4,17 @@ set consoleparam='console=tty0 console=ttyS0,115200n8'
 set kernel_path='/runmode/bzImage'
 
 set otherbootargs="rootwait rw usbcore.usbfs_memory_mb=0 consoleblank=0 rcu_nocbs=all "
-for cfg_d_file in /runmode/bootimage.cfg.d/*.cfg; do
-    source "$cfg_d_file"
+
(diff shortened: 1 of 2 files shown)


bstreiff:
@ni/rtos

amstewart (approved):

amstewart:
@bstreiff What did you do for testing here?

bstreiff:
I ran with this on a hardknott VM and verified that the various mitigation flags showed up in `/proc/cmdline`.

Also verified that the prescribed `nirtcfg --set section=SYSTEMSETTINGS,token=meltdown.mitigations.disabled,value="False"` interface still functioned (renamed `/boot/runmode/cve-2017-5754.cfg` -> `/boot/runmode/cve-2017-5754.cfg.disabled`).

gratian (approved):

PR #405, opened 2022-06-07
opkg: add patch to check if tmp_dir exists before deleting

shruthi-ravi:
For unknown reasons, the error message from the tmp_dir deletion gets embedded in the version string of *-lic packages within the opkg status file.
See bugs:
https://ni.visualstudio.com/DevCentral/_workitems/edit/1995116
https://ni.visualstudio.com/DevCentral/_workitems/edit/1956754

This patch is an attempt to unblock clients while I root cause the string issue.

I built the NIOpenEmbedded component with this change on the build server and verified that the version string of the gnupg-lic package is fixed in the runmode image. The gnupg-lic package bug is otherwise reproducible in every build on the build server.

### DIFF
--- a/recipes-devtools/opkg/files/0002-opkg-check-temp-dir-exists-before-delete.patch
+++ b/recipes-devtools/opkg/files/0002-opkg-check-temp-dir-exists-before-delete.patch
@@ -1,0 +1,32 @@
+From 7a8a38e6863d88de2869056b1cb53fc6de3d5f24 Mon Sep 17 00:00:00 2001
+From: Shruthi Ravichandran <shruthi.ravichandran@ni.com>
+Date: Tue, 7 Jun 2022 14:36:19 -0700
+Subject: [PATCH] opkg_conf: Check if tmp_dir exists before deleting
+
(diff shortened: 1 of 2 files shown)


recipes-devtools/opkg/files/0002-opkg-check-temp-dir-exists-before-delete.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/libopkg/opkg_conf.c b/libopkg/opkg_conf.c
```
amstewart, line 17: Could you add the commit message to the .patch?

https://github.com/ni/nilrt/blob/nilrt/master/hardknott/docs/CONTRIBUTING.md#patch-files

amstewart (approved):

recipes-devtools/opkg/files/0002-opkg-check-temp-dir-exists-before-delete.patch
shruthi-ravi: How about now?

amstewart (approved):

PR #421, opened 2022-06-29
nilrt.inc: Set NILRT_MAIN_FEED_VERSION to NILRT_FEED_NAME

chaitu236:
Set NILRT_MAIN_FEED_VERSION to NILRT_FEED_NAME so that it need not be
set manually every release.

Also move "require nilrt.inc" to end of nilrt.conf so that nilrt.inc can
use NILRT_FEED_NAME.

### Testing
- [x] Ran a local build without `--org` and saw the following error, which proves that these changes are working

    image

- [x] Build works fine with `--org`

### Note
Didn't remove `NILRT_MAIN_FEED_VERSION` (and replace its uses with `NILRT_FEED_NAME`) as an external customer wanting to build `nilrt/hardknott/master` will need to set `NILRT_MAIN_FEED_VERSION` to an older feed version.


conf/distro/nilrt.inc
```diff
@@ -142,7 +142,7 @@ NILRT_FEEDS_URI_RELEASE            ?= "${NILRT_FEEDS_URI}/${NILRT_FEED_NAME}"
 NILRT_MACHINE_FEED_URI_x64         ?= "${NILRT_FEEDS_URI_RELEASE}/x64"
 NILRT_MACHINE_FEED_URI_xilinx-zynq ?= "${NILRT_FEEDS_URI_RELEASE}/arm"
 
-NILRT_MAIN_FEED_VERSION            ?= "2021.8"
```
amstewart, line 145: The purpose of this variable was to let us provide a sane pre-release default version for the `ni-` feeds, that external builders could use before we publish the new feeds. I'm fine with this solution, but we should add a step to the hardknott README telling external builders to go find the latest release on `download.ni.com` and set this variable in their local.conf.

@chaitu236 Could you put up a PR for that.

amstewart (changes requested):

conf/distro/nilrt.inc
chaitu236: Done https://github.com/ni/nilrt/pull/162

chaitu236:
Please don't merge this until `ni-rt-exec-webservice` ends up in `ni-main`

shruthi-ravi (approved):

amstewart (approved):

chaitu236:
@amstewart/@shruthi-ravi this is ready to merge

PR #423, opened 2022-07-07
Remove GRUB_VERSION and associated variables

shruthi-ravi:
Fixes azdo [bug](https://ni.visualstudio.com/DevCentral/_workitems/edit/2033588).

### Testing

Built nilrt-recovery-media and successfully ran the provisioning scripts to generate a qemu safemode VM. Was also able to start and log into the VM.

### DIFF
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -2,8 +2,6 @@ MAINTAINER = "NI Linux Real-Time Maintainers <nilrt@ni.com>"
 
 TARGET_VENDOR = "-nilrt"
 
-GRUB_BRANCH = "nilrt/19.0"
-
 DISTRO_FEATURES += "${DISTRO_FEATURES_DEFAULT}"
 # Common features enabled on all NILRT flavours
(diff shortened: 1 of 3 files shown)


shruthi-ravi:
Btw, the GRUB_VERSION in sumo is also probably rubbish because GRUB_BRANCH is set to [main](https://github.com/ni/meta-nilrt/blob/715829f461d449f3926819e01e932cd993aab926/conf/distro/nilrt.inc#L5). I suspect that the version is not used in sumo either. I can clean that up too, if you agree @amstewart.

amstewart:
> Btw, the GRUB_VERSION in sumo is also probably rubbish because GRUB_BRANCH is set to [main](https://github.com/ni/meta-nilrt/blob/715829f461d449f3926819e01e932cd993aab926/conf/distro/nilrt.inc#L5). I suspect that the version is not used in sumo either. I can clean that up too, if you agree @amstewart.

Yeah; let's do it.

amstewart (approved):

PR #425, opened 2022-07-08
start_getty: Don't run getty on tty0 or tty1

chaitu236:
In /etc/inittab, we run getty on tty1 by default and want to run getty
on default serial console if "Enable Console Out" is selected in MAX.

getty shouldn't be run more than once on the same terminal. So on
targets where tty0 or tty1 end up being the preferred console, we don't
want to run getty on them because getty is already running on tty1 and
tty0 is an alias to active terminal (and may point to tty1).

Azure WI: [2055925](https://ni.visualstudio.com/DevCentral/_workitems/edit/2055925)

### Testing
Verified terminal on a Hyper-V VM works correctly.
Ran these changes on a cRIO-9049 without issues.

### DIFF
--- a/recipes-core/sysvinit/sysvinit-inittab/start_getty
+++ b/recipes-core/sysvinit/sysvinit-inittab/start_getty
@@ -10,6 +10,12 @@ CONSOLE=$2
 TERM=$3
 
 while read cons flags; do
+   # tty1 comes up by default in inittab and tty0 is an alias to active terminal.
+   # So don't select these.
+   if [ "$cons" = "tty0" ] || [ "$cons" = "tty1" ]; then
+      continue;
+   fi
+
    case "$flags" in
    *EC*) CONSOLE="$cons"
          SPEED=$(stty speed < /dev/$CONSOLE)


amstewart (commented):
Do you know why we're just now seeing this issue (in 22.8)?

chaitu236:
> Do you know why we're just now seeing this issue (in 22.8)?

The issue happens only on hardknott - on targets that don't have any serial devices and `consoles=tty0` is on `/proc/cmdline` (i.e., there is a cpld and console out is enabled in MAX) and where tty0 becomes the default console. This  probably rules out all targets but the VMs.

Before [this](https://github.com/ni/linux/commit/9f177ec9076481d310556620a075c8a9cbcbd9c1) change, in all 5.10 kernels **ttyS**s were created by default (regardless of whether the device has any serial ports). And ttyS0 is usually the preferred console over tty0/tty1 so we didn't see the issue before.

shruthi-ravi:
@chaitu236 so what happens when getty runs on more than one terminal? That isn't clear to me.

amstewart (approved):

gratian:
Can you double check things still work correctly after this change when using screen/keyboard (as opposed to serial) virtual terminals? In particular can you verify that with Embedded UI you get X showing on tty0 and a text console on tty1 (or maybe it's the other way around), and you can switch (with alt+ctrl+Fx keys) and log in?

I believe running getty on tty0/tty1 was there to support keyboard+screen terminals irrespective of the MAX console out setting.

chaitu236:
> @chaitu236 so what happens when getty runs on more than one terminal? That isn't clear to me.

Multiple instances of getty running on same tty will cause keyboard input events to go to the instance thats currently scheduled, so some keyevents will be processed by one and some by other, causing even logging into the system difficult.

chaitu236:
> I believe running getty on tty0/tty1 was there to support keyboard+screen terminals irrespective of the MAX console out setting.

In sumo, `start_getty` was replaced by `consolegetty` which had this line

> [ "$CONSOLE" = "" ] || [ "$CONSOLE" = "tty0" ] || [ "$CONSOLE" = "tty1" ] && CONSOLE="$DEFAULT_CONSOLE"

 in it which we didn't put in hardknott - all I'm doing in this change it is adding it back (in a slightly different way).

My understanding is that this script is only for the purposes of enabling consoleout because with console out disabled in MAX, kernel's cmdline won't have `consoles` set - which means `/proc/consoles` is unpopulated thereby causing this script (and the older `consolegetty` on sumo) to be a noop (getty is actually run on an unused terminal `tty2` in this case but only to keep getty alive and prevent init from restarting it if it were to die).

> Can you double check things still work correctly after this change when using screen/keyboard (as opposed to serial) virtual terminals? In particular can you verify that with Embedded UI you get X showing on tty0 and a text console on tty1 (or maybe it's the other way around), and you can switch (with alt+ctrl+Fx keys) and log in?
> 

With embedded UI enabled, I'm seeing weird behavior with or without this change when trying to switch with Ctrl+Alt+F*. Let's discuss more tomorrow.

shruthi-ravi (approved):
lgtm.

chaitu236:
>  With embedded UI enabled, I'm seeing weird behavior with or without this change when trying to switch with Ctrl+Alt+F*. Let's discuss more tomorrow.

Created [this work item](https://ni.visualstudio.com/DevCentral/_workitems/edit/2067053) to test this on real h/w like we discussed today.

@shruthi-ravi / @amstewart this can be merged.

PR #426, opened 2022-07-11
Cleanup and modernize udev-extraconf and niacctbase

amstewart:
This patchset performs the following substantial changes:
* generally cleanup the `udev-extraconf` recipe
* move the `ni` udev rules from `udev-extraconf` to a new subpackage`niacctbase-udev`, so that the `udev-extraconf` recipe no longer contains NI-specific logic.
* generally cleanup the `niacctbase` recipe.
* Bump the package versions of both `udev-extraconf` and `niacctbase` to `1.1`.
* Switch both recipes to use the new variable override syntax (`:`).
* Improve the package metadata variables in both recipes.

Nothing in this patchset is intended to change the runtime behavior of NILRT. This is just a major recipe refactor.

# Testing
Created a NIRLT VM from this patchset and compared the layout and permissions of the `/dev` tree, the udev rules, and the installed packages against the latest 22.8 export. The VMs do not generate *every* device that can appear on supported hardware, but I was able to verify that there are no changes in what is present on VMs. `niacctbase-udev` is now installed by default.


recipes-ni/niacctbase/niacctbase.bb
```diff
@@ -1,4 +1,7 @@
-DESCRIPTION = "Basic user and group settings for NI Linux RT applications."
+SUMMARY = "Common user and group settings for NI applications"
```
shruthi-ravi: Should it be `applications and devices` based on your most recent changes?

shruthi-ravi (approved):

recipes-ni/niacctbase/niacctbase.bb
amstewart: Done.

PR #431, opened 2022-07-14
glibc-tests: add rwlockbomb test

amstewart:
This PR:
1. Creates a new recipe `ptest-utils`, with a subpackage `ptest-utils-bash`, which contains the `ptest-format.sh` source file that is commonly used by NI-produced ptest recipes. Ptest subpackages can depend on this recipe as the source for this bash library.
2. Adds the `rwlockbomb` glibc-bug reproducing case to `glibc-tests-ptest`, as well as a ptest wrapper script for it.
3. Adds a `glibc-tests-src` subpackage, with the uncompiled `.c` and `.cpp` files from glibc-tests.
4. Applies some other typically-astewart refactoring to the `glibc-tests` recipe.

NOTE: the version of `rwlockbomb` in this PR has the bomb duration increased from 5 seconds to 10, because I found it slightly more reliable to produce the error.

The `test_rwlockbomb` wrapper works by setting the `sched_rt_runtime_us` ("RT throttle") to 95%, and then running the bomb. If the bomb outputs an alert that the high priority thread has stuck (the bug case), the wrapper will report that the test has failed. Otherwise, the bug did not occur and the test is considered to have passed.

# Testing
* Installed `glibc-tests-ptest` to a 22.8 VM (safemode).
```
(safemode) admin@NI-cRIO-903x-VM-8c4556e3:/usr/src/debug/glibc-tests# ptest-runner
START: ptest-runner
2022-07-14T06:38
BEGIN: /usr/lib/glibc-tests/ptest
<snip>floating_point test output lines</snip>
PASS: test_floating_point
PASS: test_oom_handling
PASS: test_overcomit_memory.sh
PASS: test_overcomit_ratio.sh
INFO: Setting sched_rd_runtime_us to 950000
thread 1 seems stuck (nv=1153503), are we throttled?
<snip>~50 similar lines</snip>
thread 4 seems stuck (nv=4508059), are we throttled?
high priority thread 3 done
high priority thread 2 done
high priority thread 1 done
high priority thread 4 done
low priority thread done
finish
FAIL: test_rwlockbomb
INFO: Setting sched_rd_runtime_us to -1
FAIL: test_rwlockbomb.sh
PASS: test_shmem
PASS: test_stack_touch
END: /usr/lib/glibc-tests/ptest
2022-07-14T06:39
STOP: ptest-runner
```
* The output from `test_rwlockbomb.sh` when run on a system with only one core:
```
(safemode) admin@NI-cRIO-903x-VM-fe012444:/usr/lib/glibc-tests/ptest# ./test_rwlockbomb.sh 
ERROR: rwlockbomb test requires more than 1 processor core.
(safemode) admin@NI-cRIO-903x-VM-fe012444:/usr/lib/glibc-tests/ptest# echo $?
99
```
* Installed `glibc-tests-src` and confirmed that the `.c` and `.cpp` files are installed to `/usr/src/debug/glibc-tests/`, and the user can compile them.


bstreiff (approved):

recipes-devtools/ptest-utils/ptest-utils.bb
```diff
@@ -0,0 +1,30 @@
+SUMMARY = "Yocto ptest helper utilities"
+DESCRIPTION = "Libraries, scripts, and source files which help developers to write and emit valid yocto ptests."
+AUTHOR = "NI RTOS <rtos@ni.com>"
+HOMEPAGE = "https://github.com/ni/meta-nilrt"
+BUGTRACKER = "https://github.com/ni/meta-nilrt/issues"
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/BSD-3-Clause;md5=550794465ba0ec5312d6919e203a55f9"
+
+PV = "1.4"
```
shruthi-ravi, line 9: What's with this specific version number?

recipes-devtools/ptest-utils/ptest-utils/ptest-format.sh
```diff
@@ -0,0 +1,82 @@
...
+# Utility script containing common functions for ptest-compatible test scripts.
+# USAGE:
+# 1) Source this file at the top of your ptest script.
+# 2) Set ptest_test to the name of your test; recommended:
+#    $(basename -s ".sh" "$0")
+# 3) Define your subtests in functions, which call ptest_pass when they pass and
+#    ptest_fail otherwise (FAIL is DEFAULT)
+# 4) Prior to calling your test function, call ptest_change_subtest
+# 5) Call your test function.
+# 6) Call ptest_report immediately after the function returns.
```
shruthi-ravi, line 13: It may be better to provide a sample ptest that uses this, that we can add to as we add more scripts to this package.

shruthi-ravi (commented):
How often does the test reproduce the issue? That isn't clear to me.

recipes-devtools/ptest-utils/ptest-utils.bb
amstewart: It's the next-minor rev of the latest version of the `ptest-format.sh` script - which is statically-included in several [other recipes](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-core/util-linux/files/ptest-format.sh). I figured it was a good place to start the versioning of this package.

recipes-devtools/ptest-utils/ptest-utils/ptest-format.sh
amstewart: Better than the comments? I agree. After we merge this PR I can make a work item to create a sample ptest in this recipe.

Is that sufficient?

amstewart:
@shruthi-ravi 
>How often does the test reproduce the issue? That isn't clear to me.

With the previous timeout (5 seconds), I personally ran it ~25 times and it detected the error ~20 of those times. With the longer timeout (10 seconds), I ran it ~10 times and it never failed to detect.

If we find these settings to lack the power we need (ie. it is flaky in the ATS), I figure we can simply adjust the test as-needed.

recipes-devtools/ptest-utils/ptest-utils/ptest-format.sh
shruthi-ravi: Yes, works for me.

shruthi-ravi (approved):

amstewart:
@shruthi-ravi 
Created NI AZDO: [#2077406](https://dev.azure.com/ni/DevCentral/_workitems/edit/2077406)

PR #433, opened 2022-07-16
linux-nilrt-next: Upgrade to 5.16 kernel

shruthi-ravi:
Built the recipe.

AzDO WI [1792637](https://ni.visualstudio.com/DevCentral/_boards/board/t/RTOS/Work%20Items/?workitem=1792637)


amstewart (approved):

gratian (changes requested):
Let's hold off on pulling this. 5.16-rt is not being maintained upstream anymore and I think I would like to update to something newer. I also want to avoid confusion with client teams which are currently using packagegroup-ni-next-kernel to test 5.15.

I have a 5.19-rt branch that is almost ready for review. I can bump this recipe at that time.  Sorry for the confusion.

PR #435, opened 2022-07-20
libpam: add NI-specific pamd.d/other file

amstewart:
NILRT prefers to handle the default PAM authentication behavior
differently from OE-core upstream. Previously, NILRT maintained an
out-of-stream commit to set the values of the `pam.d/other` file. This
commit was reverted from OE-core.

Move the NI-specific `other` file to meta-nilrt as-is.

Natinst AZDO ID: [#1580104](https://dev.azure.com/ni/DevCentral/_workitems/edit/1580104)

# Testing
* Rebuilt the `libpam` package with this change, and verified that the `libpam-runtime:/etc/pam.d/other` conffile has the NI-specific values.

# Maintainership
* Depends on OE-core [#64](https://github.com/ni/openembedded-core/pull/64).

### DIFF
--- a/recipes-extended/pam/libpam/pam.d/other
+++ b/recipes-extended/pam/libpam/pam.d/other
@@ -1,0 +1,27 @@
+#
+# /etc/pam.d/other - specify the PAM fallback behaviour
+#
+# Note that this file is used for any unspecified service; for example
+#if /etc/pam.d/cron  specifies no session modules but cron calls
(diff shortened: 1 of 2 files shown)


shruthi-ravi (approved):
Same as the other PR, verify login works as expected in the image.

amstewart:
# Testing
* Built a new VM with this change and booted into safemode. Getty still allows login via `admin`  as expected, and disallows login via anything else.

PR #437, opened 2022-07-29
tbtadm: add recipe from NI meta-oe fork

bstreiff:
tbtadm does not seem to be receiving updates, its mailing list is offline, and in other distros it seems that bolt is preferred; thus tbtadm seems unsuitable to upstream into meta-oe.

At the same time, we still have customer documentation that refers to tbtadm; this change adds the recipe under meta-nilrt until such time as we can migrate over to bolt (which also does not have an OE recipe at this time).

Minor changes to update the github URI and the new bitbake variable syntax have been merged.

### DIFF
--- a/recipes-bsp/tbtadm/tbtadm_0.9.3.bb
+++ b/recipes-bsp/tbtadm/tbtadm_0.9.3.bb
@@ -1,0 +1,22 @@
+SUMMARY = "Thunderbolt user-space management tool"
+DESCRIPTION = "tbtadm provides convenient way to interact with Thunderbolt kernel module, approve the connection of Thunderbolt devices, handle the ACL for auto-connecting devices and more."
+HOMEPAGE = "https://github.com/intel/thunderbolt-software-user-space"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://COPYING;md5=565a8f58cc6472cf9c3ee9453d7e5aa4"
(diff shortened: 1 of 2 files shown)


recipes-bsp/tbtadm/tbtadm_0.9.3.bb
```diff
@@ -0,0 +1,22 @@
...
+LIC_FILES_CHKSUM = "file://COPYING;md5=565a8f58cc6472cf9c3ee9453d7e5aa4"
+
+DEPENDS = "cmake-native udev boost"
+RDEPENDS:${PN} = "boost-filesystem"
+
+inherit cmake
+
+SRCREV = "fe0fa2237b971ec8baae36b785d98a772684e5e7"
+SRC_URI = "git://github.com/intel/thunderbolt-software-user-space.git;protocol=https \
+           file://0001-tbtadm-Disable-manpage-target.patch \
```
amstewart, line 15: Can we get rid of this .patch and just pack the man pages into a `tbtadm-docs` subpackage?

bstreiff: Not without creating a recipe for [txt2tags](https://txt2tags.org/), which is a requirement of their build process for manpages.

amstewart (approved):

chaitu236 (approved):

PR #438, opened 2022-08-03
grub-safemode-bootimage: disable initramfs_async

amstewart:
Asynchronous unpacking of the initramfs was added to linux kernel 5.15
by upstream linux.git commit e7cb072eb988e46295512617c39d004f9e1c26f8.

The change introduces a significant hang during boot (on the order of 90
seconds) on some NI targets, which is undesirable.

Disable this feature using the `initramfs_async=0` kernel commandline
option.

Natinst AZDO: [#2095204](https://dev.azure.com/ni/DevCentral/_workitems/edit/2095204)

# Testing
* Tested this command line option on several cRIOs and confirmed that it resolves the observed boot hang.
* Built the `nilrt-grub-safemode` recipe with this change and confirmed that the `bootimage.cfg` contains the correct value.

### DIFF
--- a/recipes-bsp/grub/grub/grub-safemode-bootimage.cfg
+++ b/recipes-bsp/grub/grub/grub-safemode-bootimage.cfg
@@ -1,6 +1,6 @@
 set consoleparam='console=tty0 console=ttyS0,115200n8'
 set kernel_path='/.safe/bzImage'
 set rootfs_path='ramdisk_size=populate-ramdisk-size root=/dev/ram rootfstype=ext2'
-set otherbootargs='rw consoleblank=0'
+set otherbootargs='initramfs_async=0 rw consoleblank=0'
 set usb_gadget_args="g_ether.idVendor=${USBVendorID} g_ether.idProduct=${USBProductID} g_ether.iProduct=${USBProduct}[${hostname}] g_ether.iSerialNumber=${SerialNum} g_ether.dev_addr=${usbgadgetethaddr} g_ether.bcdDevice=${USBDevice}"
 set ramdisk_path='/.safe/ramdisk.xz'


shruthi-ravi (approved):
Would you mind adding just a little more detail on why that configuration causes a hang in the commit message?

gratian (approved):

amstewart:
I completed some additional testing on PXI chassis - with some... strange results.

| Type | Cores | async=1 (s) | async=0 (s)
|:-:|:-:|:-:|:-:|
| 8880 | 8 | 3 | 4
| 8861 | 4 | 1 | 0
| 8821 | 2 | 2 | 0

The `async` columns are the approximate observed delay in boot while the initramfs is being unpacked (higher number = more bad). Strangely, all of the PXI controllers were very performant, compared to the cRIOs - even the 8821, which only has 2 cores. Also, the 8-core controller was the only one which actually *improved* by using asynchronous unpacking - but maybe not by an amount which is practically significant.

In any case, @gratian and I agree that it isn't worth our time to chase down the details of which NI hardware benefits from - or can afford to use - the async unpacking feature. And the benefits of using it seem marginal, at best. We're OK with disabling it across the platform.

amstewart:
> Would you mind adding just a little more detail on why that configuration causes a hang in the commit message?

The short answer is - I don't know. It might have something to do with the details of the HDD IO on cRIO controllers. It might be aggravated by core-contention on systems with few cores. And it could be aggravated by the fact that we have a large number of plugin-modules in the kernel (as the upstream linux commit suggests).

Since we have observed this behavior on multiple hardware types, and it is consistent, I think we have enough info to say that it is harmful. But I'm not convinced it is worth our time to track down exactly why it occurs.

PR #440, opened 2022-08-13
[hardknott] recipes-devtools/ audit

amstewart:
This PR audits the `meta-nilrt:recipes-devtools/` directory contents for quality and correctness.

Changes:
* `breakpad` : removed; superfluous to configuration already in OE-core.
* `gptfdisk` : removed; the splitting done in this recipe isn't valuable enough to maintain out of stream.
* `mkinitcpio` : detritus from the old measured boot implementation which should be dropped.
* `mtd-utils-tests` : removed; debugging tool for ARM hardware
* `opkg` : removed deprecated distribution logic
* Inconsequential syntax and styling changes throughout.

NI AZDO: [#1945386](https://dev.azure.com/ni/DevCentral/_workitems/edit/1945386)

# Testing
* Built the core feeds with these changes and confirmed that nothing breaks.


recipes-core/packagegroups/packagegroup-ni-base.bb
```diff
@@ -48,7 +48,7 @@ RDEPENDS_${PN} += "\
 	eudev \
 	fw-printenv \
 	glibc-gconv-utf-16 \
-	gptfdisk-sgdisk \
```
shruthi-ravi, line 51: How much space does it save?

shruthi-ravi (approved):

recipes-core/packagegroups/packagegroup-ni-base.bb
amstewart: `cgdisk` = 196 KiB
`gdisk` = 208 KiB
`fixparts` = 72 KiB
`sgdisk` = 196 KiB

So previously, we were just installing `sgdisk` at 196 KiB, now we would install all utilities at 672 KiB.

PR #446, opened 2022-08-16
[hardknott] recipes-ni/ audit

amstewart:
This PR audits the recipes in the `meta-nilrt:recipes-ni/` directory.

* Generally renames recipes whose entire source is within meta-nilrt from `recipe_1.0.bb` to `recipe.bb`, and adds an explicit `PV = ` assignment within the recipes. It is very unlikely that we would ever have multiple, contemporaneous versions of these recipes needing to be distinguished by version number. Using an explicit `PV` variable lets devs more-easily rev the package version without introducing renames into the git history.
* Syntax and styling changes throughout.
* `base-files-nilrt` : removed deprecated ARM logic
* `initscripts-nilrt` : removed `nilrt-nxg` and non-x64 logic
* `nirtcfg-tests` : fixed broken do_install() (missing SRC_URI declaration)
* `ni-rtfeatures` : removed an unneeded `DEPENDS`
* `nisdbootconfig` : dropped; this was an ARM-only helper script to let MTD devices boot from SD cards.
* `ni-smbios-helper` : added missing `RDEPENDS` for utilities that are used by scripts in this package.
* `ni-systemformat` : added missing `RDEPENDS` for utilities that are used by scripts in this package.
* `ni-systemimage` : added missing `RDEPENDS` for utilities that are used by scripts in this package.
* `expand-disk` : dropped; this recipe has not been built or used by anything in a long time.
* `sysconfig-settings` : greatly simplified the `SRC_URI` and `FILES` assignments so that each file doesn't need to be individually specified.
    * Visually checked the packages produced by the simplified recipe and confirmed that all the source files are in the correct subpackages.

NI AZDO: [#1945386](https://dev.azure.com/ni/DevCentral/_workitems/edit/1945386)

# Testing
* Built each package as modifications were made and confirmed that they all still built.
* Where I wasn't confident my syntax changes were inconsequential, I verified the changes specifically.
* The core packagefeed still builds.


recipes-ni/sysconfig-settings/sysconfig-settings.bb
```diff
@@ -0,0 +1,116 @@
...
+
+PV = "2.0"
+
+
+
+SRC_URI = "\
+	file://nisetembeddeduixml \
+	file://systemsettings \
+	file://uixml \
+"
```
amstewart, line 18: @bstreiff I tried looking around and didn't see any commentary on the original sysconfig-settings PR as to why we were declaring each source file individually. Was there a reason?

amstewart:
CLOSED: this PR has been merged into my feature branch at [`github.com/amstewart/meta-nilrt:dev/hardknott/audit/master`](https://github.com/amstewart/meta-nilrt/tree/dev/hardknott/audit/master). It will be merged into the hardknott mainline later.

PR #447, opened 2022-08-16
docs: add layer styleguide

amstewart:
Add a first draft of a meta-nilrt layer styleguide, to coordinate syntax
and styling.

This styleguide includes some existing style sections from the [nilrt.git CONTRIBUTING guide](https://github.com/ni/nilrt/blob/a8e494d6181a7f858e0cd4385c68e40c0a7e06d3/docs/CONTRIBUTING.md?plain=1#L41). I'll submit a PR to remove it from that document.


docs/styleguide.md
```diff
@@ -0,0 +1,198 @@
...
+
+* `Upstream-Status: Inappropriate [$rationale]` ([ex](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-support/curl/curl/0005-Add-nicurl-wrapper-functions.patch)) For when the `.patch` change is specific to NI LinuxRT and would not be desired (or has been rejected) by upstream.
+* `Upstream-Status: Not Submitted [$rationale]` ([ex](https://github.com/ni/meta-nilrt/blob/nilrt/master/sumo/recipes-gnome/florence/files/0004-Add-option-for-automatic-bring-to-top.patch)) For when the `.patch` file is being included in NILRT before being submitted to its upstream project. This should only occur if the `.patch` is needed for an immediate NILRT release and there is no time to get it reviewed upstream beforehand. Or if - as in the case of the example - the upstream mailing list is dead.
+* `Upstream-Status: Submitted [$upstream_mailing_list_link]` ([ex](https://github.com/ni/meta-nilrt/blob/1e6453ebca8735de96eaaf3bc931d22998c8dfb3/recipes-support/curl/curl/0014-Fixup-lib1529-test.patch)) For when the `.patch` has been submitted to the upstream project's mailing list, but needs to be pulled into NILRT prior to final upstream approval.
```
bstreiff: nit: some upstream projects don't use mailing lists, patch submissions happen via a web PR system. Would we have (for example) a GitHub or GitLab PR URL as the "upstream mailing list link" in that situation?

amstewart: Yes. I'll add some clarification to the doc.

amstewart:
# Patch V2
* rebased to the hardknott mainline
* [Implemented feedback](https://github.com/ni/meta-nilrt/pull/447/commits/fc8dc83397d8a24bd7e83e76ca9b7d5d449010c7) from @bstreiff .

bstreiff (approved):
"accomidate upstreames" could use a spellcheck pass but I assume you'll probably squash that into the other commit.

docs/styleguide.md
```diff
@@ -0,0 +1,198 @@
...
+
+# General
+
+**Capitalize NI LinuxRT and NILRT**
+
+When referring to "NI LinuxRT" in recipes and documentation, or its abbreviated form: "NILRT", capitalize the N-I-L-R-and-T. The exception to this rule is when speaking about the "meta-nilrt" layer itself, or the "nilrt" git repository - where it is recommended you call it "nilrt.git".
+
+**Indentation**
+
+Use **tabs** for indentation and **spaces** for alignment in all places - *except* within python task functions, where spaces should be used for both.
```
shruthi-ravi, line 12: Why the deviation here?

docs/styleguide.md
```diff
@@ -130,9 +130,9 @@ Contributors are free to use a single newline to create logical breaks between d
 At a minimum, you should add an additional message trailer declaring the [upstream status](https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines#Patch_Header_Recommendations:_Upstream-Status) of the .patch file at the time of your OE commit. These status lines are crucial to helping the project maintainers properly upgrade and rebase recipes. Common status lines are:
```
shruthi-ravi, line 130: Typos in the commit message.

docs/styleguide.md
amstewart: Because I believe strongly that soft tabs are degenerate, and also upstream isn't actually consistent about using soft tabs.

docs/styleguide.md
amstewart: Fixed.

amstewart:
# Patch V3
* Audited the commit messages and styleguide body for spelling and grammar.

docs/styleguide.md
shruthi-ravi: It sounds like a personal opinion. Even if upstream doesn't enforce the soft tabs, if it is in their style guide, I think we should follow it. It'll be less of a hassle for developers who follow the style guide and wish to contribute to multiple layers.

amstewart: Like I said, upstream doesn't actually enforce this style and you can find many examples of both in upstream. Is this a blocker for you, because I feel strongly about it, but I don't want to go through the effort of selling it to upstream to change on the wiki?

shruthi-ravi: Sorry, I missed this response. I do feel strongly about enforcing this policy since I think it is not sustainable. I think we should only care about consistency within a file or write a linter to do the work for us. But since you've already made all of those changes based on what you felt, I'm not going to block the way.

PR #449, opened 2022-08-19
[hardknott] Add logging to system log for pstore saves.

SparkingSpork:
Currently, pstore-save only saves to dmesg. Add
logging to syslog as well for support purposes.

AzDO Work Item: https://dev.azure.com/ni/DevCentral/_workitems/edit/1991389/

## Testing:
- Forced a crash with magic sysrq and confirmed that the new item appeared in the logs:
```
admin@rtos-ci-nilrt-b:~# dmesg | grep pstore
[    0.121952] pstore: Registered efi as persistent store backend
[    0.122569] pstore: Using crash dump compression: deflate
[    1.143127] pstore-save: Saving prior dmesg from crash to /var/lib/pstore/20220819_204926
admin@rtos-ci-nilrt-b:~# cat /var/log/messages | grep pstore
2022-08-19T20:50:04.921+00:00 rtos-ci-nilrt-b kernel: [    0.121952] pstore: Registered efi as persistent store backend
2022-08-19T20:50:04.921+00:00 rtos-ci-nilrt-b kernel: [    0.122569] pstore: Using crash dump compression: deflate
2022-08-19T20:50:04.921+00:00 rtos-ci-nilrt-b [    1.143127] pstore-save: Saving prior dmesg from crash to /var/lib/pstore/20220819_204926
```
- Ran the ptests and confirmed they passed:
``` 
START: ptest-runner
2022-08-19T20:51
BEGIN: /usr/lib/pstore-save/ptest
PASS: pstore_save_ptest
END: /usr/lib/pstore-save/ptest
2022-08-19T20:51
STOP: ptest-runner
````

### DIFF
--- a/recipes-ni/pstore-save/files/pstore-save
+++ b/recipes-ni/pstore-save/files/pstore-save
@@ -46,6 +46,7 @@ done
 
 function log_msg () {
     echo "$SCRIPTNAME: $@"
+    logger "$SCRIPTNAME: $@"
 }
 
 function log_error () {

### CODE REVIEW

File: recipes-ni/pstore-save/files/pstore-save [hunk 1]
```diff
@@ -46,6 +46,7 @@ done
 
 function log_msg () {
     echo "$SCRIPTNAME: $@"
+    logger "$SCRIPTNAME: $@"
```
SparkingSpork, line 49: Default log level is `user.notice`. I'm assuming that's sufficient.

amstewart: Does this actually do anything during boot? At `S09`, the `pstore-save` [initscript runs](https://github.com/ni/meta-nilrt/blob/4402ef086bb2d9b4fdfd47845dbcd0100dbfe211/recipes-ni/pstore-save/pstore-save_1.0.bb#L19) and [calls the `pstore-save` bin script](https://github.com/ni/meta-nilrt/blob/4402ef086bb2d9b4fdfd47845dbcd0100dbfe211/recipes-ni/pstore-save/files/initscript#L36), and redirects stdout to `/dev/ksmg`. So I think the `echo` line above here is what is getting logged to the kernel log.

The `syslog` daemon starts at `S20`, so I don't think the `logger` call would actually do anything during boot - which is the same issue that I had in #450, and why I had to split the initscripts for the `ni-rtfeatures` package.

SparkingSpork: I see entries in both syslog and dmesg after manually crashing a target. Does `/dev/kmsg` get output to `/var/log/messages` as well?

SparkingSpork: After some more poking around, I think this change is unnecessary. Since we're logging to the kernel ring buffer during boot, `syslogd` already grabs the log item and copies it to `/var/log/messages` when the daemon starts. Abandoning this PR.

SparkingSpork:
Abandoning. Anything written to the kernel ring buffer during boot appears to be copied to the system log when `syslogd` starts.

PR #450, opened 2022-08-19
[hardknott] Display the CPLD reset_source to users during boot.

amstewart:
We have had recent customer escalations where knowing the reset source on a cRIO controller has been critical. This information is available on most Linux RT cRIOs by reading the appropriate CPLD register. It was useful for customer troubleshooting to create a log that recorded this reset source at system startup. Having this kind of logging implemented as a feature could make troubleshooting unexpected system reboots much easier since we'll have a record of previous reset sources and can narrow down root causes quickly from there. 

This patchset adds a new initscript to the `ni-rtfeatures` package - appropriately called `/etc/init.d/ni-rtfeatures`. The new initscript reads the `/dev/rtfeatures/reset_source` sysfs entry, interprets the value, and then prints its interpretation to the boot log, and to the message log.

This patchset also generally cleans up the `ni-rtfeatures` recipe to conform to modern styling.

**Example Console Output**
```
Starting NIAuth: done                                                           
Starting OpenBSD Secure Shell server: sshd                                      
done.                                                                           
Starting Advanced Configuration and Power Interface daemon: acpid.              
<46>Aug 22 16:23:33 ni-rtfeatures: reset_source=processor  # Reset from MAX or command line
Starting NI Service Locator: done                                               
```
**Example /var/log/messages Output**
```
2022-08-22T16:23:33.000+00:00 NI-cRIO-9042-01E4AFBF ni-rtfeatures: reset_source=processor  # Reset from MAX or command line                                 
```

Natinst AZDO: https://dev.azure.com/ni/DevCentral/_workitems/edit/1004889

# Testing
* [x] Installed this package to a 9042 and verified that the `reset_source` is now printed to both console and to `/var/log/messages`.
* [x] Verified the same behavior on a crio-903x.
* [ ] Verified the same behavior on a crio-905x.
* [x] Tested the script output with the various possible (and impossible) values of `reset_source`.


amstewart:
I'll post samples of what the output looks like on Monday.

recipes-ni/ni-rtfeatures/ni-rtfeatures.bb
```diff
@@ -0,0 +1,66 @@
...
+}
+
+pkg_postinst:${PN} () {
+	if [ -n "$D" ]; then
+		OPT="-r $D"
+	else
+		OPT=""
+	fi
+	update-rc.d $OPT handle_cpld_ip_reset start 6 1 3 4 5 .
+	update-rc.d $OPT ni-rtfeatures start 20 1 3 4 5 .
```
amstewart, line 39: I would have rolled these scripts into one, but there is a race issue with syslog. `handle_cpld_ip_reset` wants to start before `sshd` - so <09. `ni-rtfeatures` needs to start after `syslog` - so >19. I thought for a bit about potentially moving sshd upward, or syslog downward, but eventually decided against it.

But note to future devs: that is why this initscript is split.

recipes-ni/ni-rtfeatures/files/ni-rtfeatures.initd
```diff
@@ -0,0 +1,70 @@
...
+
+DEV_NIRTFEATURES="/dev/nirtfeatures"
+
+
+# Log messages to the console and to the system message buffer.
+log () {
+	local level=$1
+	local msg=${@:2}
+	local logger_stderr=""
+	[ "$VERBOSE" ] && logger_stderr="--stderr"
```
amstewart: @SparkingSpork 
Is it desirable to hide the console output if the user doesn't have `VERBOSE=yes` during boot? Or should we show it all the time?

SparkingSpork: I think showing this all the time is reasonable. It's a single line and is useful information regardless of if there's detailed debugging going on or not.

amstewart: Removed the unused `logger_stderr` logic in latest commits.

SparkingSpork (approved):

shruthi-ravi (approved):

PR #463, opened 2022-09-20
niconfigpersistentlogs: add init script to configure persistent logs

jpautler:
As installed, /var/log is a symlink to /var/volatile/log which is located on a tmpfs mount. As a result, logs are lost on each reboot. This change provides a simple way for users to change logs to be persistent across reboots for debugging purposes and to change back again.


recipes-core/packagegroups/packagegroup-ni-runmode.bb
```diff
@@ -23,6 +23,7 @@ RDEPENDS_${PN} = "\
 	librtpi \
 	linux-firmware-radeon \
 	lldpd \
+	niconfigpersistentlogs \
```
amstewart: Can you add a hyphen here, like `ni-configpersistentlogs`? That's the new package naming style for NI packages.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
```diff
@@ -0,0 +1,83 @@
...
+
+set -uo pipefail
+
+error_and_die() {
+set +u
+    if [ -n "$1" ]; then
+        echo >&2 "$1"
+    fi
+set -u
+    echo >&2 "ERROR"
```
amstewart: Why not like: `echo >&2 "ERROR: $1"`?

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
```diff
@@ -0,0 +1,83 @@
...
+    exit 1
+}
+
+FILE_PATH_VOLATILES_CONFIG="/etc/default/volatiles/00_core"
```
amstewart: I can't recall which package owns this file, but `ni-configpersistentlogs` should have an `RDEPENDS` on it.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
```diff
@@ -0,0 +1,83 @@
...
+PERSISTENT_DIR_VAR_LOG_LINE="d root root 0755 /var/log none"
+VOLATILE_SYMLINK_VAR_LOG_LINE="l root root 0755 /var/log /var/volatile/log"
+
+ENABLE_PERSISTENT_LOGS_INI_VAL=$(/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=PersistentLogs.enabled,value="false" | tr "[:upper:]" "[:lower:]")
```
amstewart: New guidance is to rely on `PATH` to evaluate your binaries. So `$(nirtcfg ...)` is preferred.

recipes-ni/niconfigpersistentlogs/niconfigpersistentlogs_1.0.bb
```diff
@@ -0,0 +1,27 @@
+SUMMARY = "NI persistent logs configuration utility"
+DESCRIPTION = "Installs the niconfigpersistentlogs utility"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+SECTION = "base"
+
+DEPENDS += "shadow-native pseudo-native niacctbase update-rc.d-native"
```
amstewart: I don't think you need any of these `DEPENDS`.

* pseudo-native isn't required for any recent OE recipes, I don't think.
* shadow-native and niacctbase are only needed if you're using the labview user account or writing to its /home directory.
* update-rc.d-native will be included in the DEPENDS when you inherit `update-rc.d.bbclass`.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
```diff
@@ -0,0 +1,83 @@
...
+fi
+
+if [ "$RM_VAR_LOG" = true ]; then
+    if ! rm -rf /var/log
+    then
+        error_and_die
+    fi
+fi
+
+set +u
```
amstewart: You keep setting and un-setting `-u`, but I don't think the script has any real opportunities to have an undefined variable. It doesn't take any user input other that what is defined in `ni-rt.ini`, and that is sanitized by its own logic. So I think you can just remove the `-u` setting.

Am I missing something?

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
```diff
@@ -0,0 +1,83 @@
...
+        error_and_die
+        fi
+    elif [ "$EXIT_CODE" = 1 ]; then
+        error_and_die "Could not locate expected /var/log entry in $FILE_PATH_VOLATILES_CONFIG."
+    else
+        error_and_die
+    fi
+else
+    error_and_die
+fi
```
amstewart: Rather than `sed`ing a file which is owned by another package, what if we just removed the `/var/log` declaration from `00_core` and used a new runparts file in the `/etc/default/volatiles` directory which is wholely owned by this recipe? Then this script's logic is just to swap out the contents of its runparts file wholesale.

amstewart (changes requested):

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
mike-petersen-ni: If we want a one-liner in that format without unsetting the `-u` flag, this would be better:

    echo >&2 "ERROR${1+: $1}"

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
```diff
@@ -0,0 +1,83 @@
...
+
+if ! rm -rf /etc/volatile.cache
+then
+    error_and_die
+fi
+
+if [ "$RM_VAR_LOG" = true ]; then
```
shruthi-ravi: What happens if the user does not reboot after this? Is it going to leave the system in a bad state wrt the logging? Should that logic be part of an init script?
And sorry, I forgot what you said about the following: Do you intend this script to be only run as an initscript? So would the advised workflow be:
Make the ni-rt.ini change
Reboot for the init script to make the file changes
Reboot for the changes to take effect 
Or have I completely misunderstood?

SparkingSpork: My understanding is that this script runs before the script that populates the volatiles and other directories. So only the initial reboot is necessary. Jeff can correct me if I'm wrong, but here is what I understand to happen:
User changes ini -> User reboots -> this script runs -> populate-volatile.sh runs and will recreate the directories
Note that all of this will happen prior to syslogd starting so there shouldn't be anything logging to /var/log at this point in the boot process

recipes-core/packagegroups/packagegroup-ni-runmode.bb
jpautler: Done.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
jpautler: I had been printing the ERROR line separately because in some cases a call to a utility (grep, sed) would directly write its own output to stderr. I realized it would be better to just capture the stderr output of those calls directly and pass them along to error_and_die instead.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
jpautler: It is the initscripts package. I added it to RDEPENDS.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
jpautler: Done.

recipes-ni/niconfigpersistentlogs/niconfigpersistentlogs_1.0.bb
jpautler: Thanks for the explanation. I had meant to ask about this and forgot. I had just copied that line from similar recipes.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
jpautler: I've been in a habit of always using -u to catch potential issues during development and at least give a good error message if somethings slips through and ships. I usually don't have to toggle it off and back on much. But in this case it's probably more trouble than it's worth. I removed it.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
jpautler: We had some offline discussion and decided to leave the current approach intact. Moving this line out to a separate runparts file had its own complications:

1. There are ordering dependencies with other entries in 00_core, such as those defining /var/volatile/log and /var/log/wtmp. So we would have to move at least one of those out as well. There's also a potential issue that other upstream changes to 00_core could add additional dependencies over time that we'd have to deal with.

2. By moving the /var/log line out of 00_core and into a separate runparts file in the ni-configpersistentlogs package, we would be making ni-configpersistentlogs a required package to have /var/log defined on a system. We don't really want ni-configpersistentlogs to be required. Specifically, we don't want it to be part of a safemode image where the entire rootfs is RAM-based and the whole idea of persistent logs doesn't really make sense.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
jpautler: That explanation is correct. This script is only intended to be run via init. Only one reboot is needed and it should happen after ni-rt.ini is changed.

shruthi-ravi (approved):
Can you squash your commits when it is ready to be merged?

recipes-ni/ni-configpersistentlogs/ni-configpersistentlogs_1.0.bb
```diff
@@ -0,0 +1,28 @@
+SUMMARY = "NI persistent logs configuration utility"
```
amstewart: Sorry; I should have noted this before. Since all of the source for `ni-configpersistentlogs` lives in OE directly, encoding the version name into the recipe filename doesn't do much. It just means that, if a future dev does want to rev the package version, they'll have to do a git-rename operation, which can complicate the git history.

Instead, you can just rename the recipe file to `ni-configpersistentlogs.bb` and add a `PV = "1.0"` line in the recipe. Then devs can just update that variable.

recipes-ni/niconfigpersistentlogs/files/niconfigpersistentlogs
amstewart: Did you change it? It looks the same to me.

recipes-ni/ni-configpersistentlogs/ni-configpersistentlogs_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+	install -d ${D}${sysconfdir}/init.d/
+	install -m 0755 ${S}/ni-configpersistentlogs ${D}${sysconfdir}/init.d/
+}
+
+
+FILES_${PN} += "\
+	${sysconfdir}/init.d/ni-configpersistentlogs \
+"
+
+RDEPENDS_${PN} += "bash initscripts"
```
amstewart: Could you change these to use the new override syntax. `FILES:${PN} += ...` `RDEPENDS:${PN} += ...`

amstewart (changes requested):

amstewart (approved):

PR #467, opened 2022-09-29
[hardknott] audit 2-1: use new override syntax

amstewart:
OE upstream is deprecating the underscore-syntax for declaring variable overrides, in favor of the less-ambiguous colon-syntax. Convert all instance of the underscore-syntax to the colon.

This commit is not intended to change behavior.

---

This PR uses a slightly modified version of the [`convert-overrides.py`](https://git.openembedded.org/openembedded-core/tree/scripts/contrib/convert-overrides.py) script from OE-core upstream, to semi-automatically cover the variable override syntax to conform to the new *colon-style*.

This change is optional for hardknott, but will be a requirement for future rebases. So we may as well do it now.

# Testing
* Manually reviewed each change that the script applied, and fixed the few places that the script made the wrong decision.
* Compared the resulting layer state with the previous manual audit we performed. This script caught a few [spots we missed](https://github.com/amstewart/meta-nilrt/blob/dev/hardknott/audit/master/recipes-xfce/xserver-xfce-init/xserver-xfce-init_1.0.bb#L43) in the manual audit.
* Ran `bitbake -p world` and verified that everything still parses correctly.


mike-petersen-ni (approved):
Looks good. Thanks for splitting up these changes. 👍

recipes-kernel/linux/linux-nilrt.inc
```diff
@@ -47,14 +47,14 @@ SRCREV ?= "${AUTOREV}"
...
+LIC_FILES_CHKSUM:xilinx-zynq = "file://COPYING;md5=d7810fab7487fb0aad327b76f1be7cd7"
 
 # This checks ${PV} == version from kernel sources which our PV/AUTOREV breaks, so skip it.
 KERNEL_VERSION_SANITY_SKIP="1"
 
-FILESEXTRAPATHS_prepend := "${THISDIR}/${KBUILD_FRAGMENTS_LOCATION}:"
-KBUILD_DEFCONFIG_x64 = "nati_x86_64_defconfig"
-KBUILD_DEFCONFIG_armv7a = "nati_zynq_defconfig"
+FILESEXTRAPATHS:prepend := "${THISDIR}/${KBUILD_FRAGMENTS_LOCATION}:"
+KBUILD_DEFCONFIG:x64 = "nati:x86_64_defconfig"
```
shruthi-ravi: Should the value have not been replaced?

shruthi-ravi (approved):

recipes-kernel/linux/linux-nilrt.inc
amstewart: I knew I was going to miss one. :(

Will fix.

amstewart:
# Patch v2
* Fixed `KBUILD_DEFCONFIG:x64` assignment in linux-nilrt.inc. @shruthi-ravi

PR #468, opened 2022-09-29
[hardknott] audit 2-2: rename local recipes

amstewart:
remove versions from local recipe filenames
    
    For recipes whose entire source is storied in OE layers, it isn't
    necessary or desirable to declare the recipe version in the filename.
    Doing so means that, if a developer makes a change to the local source,
    and wants to rev the recipe version, he will have to commit a git-move
    operation - which complicates the git history.
    
    Instead, allow for the package version to be defined within the recipe
    source itself.
    
    Since all of the recipes touched by this commit are version "1.0", and
    that is the default package version, it isn't necessary to explicitly
    define the variable in the recipe source just yet. Devs can do that when
    they want to change it from 1.0.

----

This PR also includes a fixup to the `nirtcfg-tests` recipe to make it build.

# Testing
* Confirmed that every recipe touched by this commit still builds.

# Maintainership
* This PR is based on #467.


recipes-core/initrdscripts/init-nilrt-ramfs.bb
```diff
@@ -2,6 +2,10 @@ SUMMARY = "NILRT initramfs init script"
 LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
+
+PV = "1.0"
```
mike-petersen-ni, line 6: These four files:

    recipes-core/initrdscripts/init-nilrt-ramfs.bb
    recipes-core/initrdscripts/init-restore-mode.bb
    recipes-xfce/xfce-nilrt-settings/xfce-nilrt-settings.bb
    recipes-xfce/xserver-xfce-init/xserver-xfce-init.bb

each declare their versions to be 1.0, though the commit mentions that that is the default and simply renames most recipes.
Are these four files special in some way?

mike-petersen-ni (approved):

recipes-core/initrdscripts/init-nilrt-ramfs.bb
amstewart: TBH because I started with those files and added the PV variable. Then realized that it didn't matter and would take 3 times as long to do it to all the files, so I just stopped. I can go back and remove them if you have some objection to them being defined, but it doesn't make a technical difference.

shruthi-ravi (approved):

amstewart:
# Patch V2
* Rebased on `nilrt/master/hardknott`. No changes.

PR #472, opened 2022-10-01
nilrt-academic: Move 'require nilrt.inc' to EOF

chaitu236:
Move 'require nilrt.inc' to end of file for better readability.
This doesn't affect functionality.

WI: [2163365](https://dev.azure.com/ni/DevCentral/_workitems/edit/2163365)

### Testing
Built locally

### DIFF
--- a/conf/distro/nilrt-academic.conf
+++ b/conf/distro/nilrt-academic.conf
@@ -1,5 +1,3 @@
-require nilrt.inc
-
 DISTRO_NAME = "NI Linux Real-Time - Academic"
 
 DISTRO_VERSION = "8.14"
@@ -33,3 +31,5 @@ IMAGE_LINGUAS ?= "en-us en-us.iso-8859-1"
 # own release namespace.
 NILRT_FEEDS_URI_RELEASE = "${NILRT_FEEDS_URI}/academic/${NILRT_FEED_NAME}"
 PACKAGE_FEED_URIS = "${NILRT_FEEDS_URI_RELEASE}"
+
+require nilrt.inc


amstewart:
I don't think I agree that this is "better readability". In BB recipes, [recipe `includes`](https://github.com/ni/openembedded-core/blob/c848b4b7d024b88ef041507f5f727bca5c954a81/meta/recipes-kernel/linux/linux-yocto_5.10.bb#L3) and class `inherits` are both usually [pretty early](https://www.openembedded.org/wiki/Styleguide#Ordering_and_grouping) in the extending file. And I think it is intuitive that the `nilrt-academic.conf` file *extends* the `nilrt.inc` includes file, so the includes is defined first.

What's your objection to having it at the top?

chaitu236:
> I don't think I agree that this is "better readability". In BB recipes, [recipe `includes`](https://github.com/ni/openembedded-core/blob/c848b4b7d024b88ef041507f5f727bca5c954a81/meta/recipes-kernel/linux/linux-yocto_5.10.bb#L3) and class `inherits` are both usually [pretty early](https://www.openembedded.org/wiki/Styleguide#Ordering_and_grouping) in the extending file. And I think it is intuitive that the `nilrt-academic.conf` file _extends_ the `nilrt.inc` includes file, so the includes is defined first.
> 
> What's your objection to having it at the top?

Since `nilrt.inc` needs variables defined in `nilrt-academic.conf` to expand some of its variables, it made more sense to me that definition should precede use.

But it doesn't matter in this case, so I'll drop this.

PR #476, opened 2022-10-12
kernel-tests: rewrite test_kernel_security

bstreiff:
This test had a few problems:

1) It looked for the individual kernel cmdline options, but we've since
   changed to using the catch-all `mitigations` option, which this test
   didn't know about.

2) It looked for specific mitigation states, utilizing an (incomplete)
   database based on device model code-- but some model codes (like PXI)
   have a wide range of CPU options which may require different sets of
   mitigations (which themselves might change with future kernels as new
   CPU vulnerabilities are discovered)

Instead of trying to pick out each individual mitigation feature, lets trust that the kernel `mitigations` option does what it says it does, and have the test just confirm that we're setting that appropriately.

The test will also print out selected /proc/cpuinfo fields, the current mitigation state, and kernel cmdline on an informational basis.


bstreiff:
sample output (on a VM, so not directly matching any specific target, but enough to get the idea):
```
INFO: cpu under test:
INFO:     vendor_id: AuthenticAMD
INFO:     cpu family: 23
INFO:     model: 49
INFO:     model name: AMD EPYC-Rome Processor
INFO:     stepping: 0
INFO:     microcode: 0x1000065
INFO:     bugs: sysret_ss_attrs null_seg spectre_v1 spectre_v2 spec_store_bypass retbleed
INFO: vulnerability status:
INFO:     itlb_multihit: Not affected
INFO:     l1tf: Not affected
INFO:     mds: Not affected
INFO:     meltdown: Not affected
INFO:     mmio_stale_data: Not affected
INFO:     retbleed: Vulnerable
INFO:     spec_store_bypass: Vulnerable
INFO:     spectre_v1: Vulnerable: __user pointer sanitization and usercopy barriers only; no swapgs barriers
INFO:     spectre_v2: Vulnerable, IBPB: disabled, STIBP: disabled, PBRSB-eIBRS: Not affected
INFO:     srbds: Not affected
INFO:     tsx_async_abort: Not affected
INFO: kernel cmdline:
INFO:     BOOT_IMAGE=/runmode/bzImage root=PARTUUID=fa084617-e575-4ca6-a0cb-682e2c85b43f rootwait rw usbcore.usbfs_memory_mb=0 consoleblank=0 rcu_nocbs=all mitigations=off debug console=tty0 console=ttyS0,115200n8 sys_reset=false
PASS: test_kernel_security
```

gratian (approved):

amstewart (approved):

PR #479, opened 2022-10-13
kernel-tests: make i915 test skip on non-i915 targets

bstreiff:
Change the i915 firmware test to enumerate PCI devices looking for video adapters, and skip the test unless we determine that there are Intel video adapters in the system.

This approach lets us clean up a few other things in the test:

- We don't need to get the vendor name from DMI to confirm we're on a NI controller, since there's nothing NI-specific about "look for Intel video devices on PCI"

- We don't need to parse lsmod output, we can instead look in sysfs to see if the i915 driver is attached to anything, instead of using the "users" count, which was 0 when I checked it on a cRIO-9041. (This also makes the test work if we ever decided for some reason to configure i915 as a built-in, although I don't know why we would.)

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-tests-files/test_i915_firmware.sh
+++ b/recipes-kernel/kernel-tests/kernel-tests-files/test_i915_firmware.sh
@@ -5,6 +5,9 @@
 source $(dirname "$0")/ptest-format.sh
 
 ptest_test=$(basename "$0" ".sh")  # test_i915_firmware.sh
+ptest_rc=0
+
+declare -xr SYSFS_PCI_DEVS="/sys/bus/pci/devices"
 
 # Skips this test if we are on an ARM device.
 function check_arch () {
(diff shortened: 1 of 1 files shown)


amstewart (approved):

recipes-kernel/kernel-tests/kernel-tests-files/test_i915_firmware.sh
```diff
@@ -41,43 +44,50 @@ function check_files () {
...
-	fi
+			# get the vendor id
+			vendor_id=$(cat $SYSFS_PCI_DEVS/$dev/vendor)
+			if [ "${vendor_id}" = "0x8086" ]; then
+				has_intel_video_device=1
+			fi
+
+			driver_name=$(basename $(realpath $SYSFS_PCI_DEVS/$dev/driver))
+			if [ "${driver_name}" = "i915" ]; then
+				has_i915_driver_loaded=1
```
shruthi-ravi, line 69: Do you need to continue looking at the remaining devices once you find the i915 driver?

shruthi-ravi (approved):

recipes-kernel/kernel-tests/kernel-tests-files/test_i915_firmware.sh
bstreiff: Nope, we do not. My initial stab at this loop didn't incorporate the driver check because I didn't expect to remove `check_module`, but yeah I think now once we satisfy the precondition of "we have at least one GPU device with i915 loaded" we can stop looking.

PR #483, opened 2022-10-14
[hardknott] audit 2-5: recipes-ni changes

amstewart:
This PR adds some missing runtime dependencies to recipes in the `recipes-ni/` directory. It also moves the `fw-printenv` reciep out of the `ni-utils/` subdirectory, into its own.

# Testing
* `fw-printenv` still builds (which verifies the above changes, since they don't substantially impact recipe logic.)
* The other changes in this PR are adding common packages to `RDEPENDS`, so I expect that they will all reliably install on a running target.


recipes-ni/ni-systemformat/ni-systemformat.bb
```diff
@@ -42,4 +42,14 @@ inherit update-rc.d
...
+	bash \
+	niacctbase \
+"
+# nisystemformat rdeps
+RDEPENDS:${PN} += "\
+	coreutils \
+	ni-netcfgutil \
+	util-linux-lsblk \
+"
+RDEPENDS:${PN}:append:x64 = " e2fsprogs-mke2fs"
```
shruthi-ravi: Why a separate x64 list here?

shruthi-ravi (approved):

recipes-ni/ni-systemformat/ni-systemformat.bb
amstewart: The `nisystemformat` script [uses the `mkfs.ext4` binary](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-ni/ni-systemformat/files/nisystemformat#L172), but only when the system architecture is x64 (for the obvious reasons). So the runtime dependency technically only exists on x64.

shruthi-ravi: I thought we were removing x64-specific code with the assumption that we support this image only on x64. Is that not the case here?

amstewart: I originally didn't expect that `e2fsprogs-mke2fs` would exist in the ARM feeds, but I suppose we don't build those in hardknott anyway, so it doesn't matter. Will fix.

amstewart: Fixed in https://github.com/ni/meta-nilrt/pull/483/commits/57a560de271c9ac09ee990368e98a9080279226c.

PR #485, opened 2022-10-17
kernel-tests: Add test to diff dmesg log

chaitu236:
Add a new test that diffs the dmesg log with a previous version obtained from a database and uploads the latest log to this database.

### Testing
- [x] Built `kernel-tests` recipe, ensured `kernel-tests-ptest.ipk` is installable
- [x] Ran with current and next kernels on a VM.
- [x] Ran on actual h/w with the db populated with old logs.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
+++ b/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
@@ -1,0 +1,185 @@
+import argparse
+from bson.binary import Binary
+import datetime
+import difflib
+import mmap
+import os
+import pymongo
+import re
+import subprocess
+import sys
+
+class DB:
+    mongo_db_name = 'rtos'
+    mongo_db_collection_name = 'dmesg_logs'
+
+    def __init__(self, server, user, password):
+        mongo_client = pymongo.MongoClient("mongodb://{user}:{pwd}@{server}".format(
+            user=args.user, pwd=args.password, server=args.server
+            ),
+            w=1)
+        db = mongo_client[self.mongo_db_name]
+        self.dmesg_logs_collection = db[self.mongo_db_collection_name]
+
+    def insert(self, data):
+        return self.dmesg_logs_collection.insert_one(data)
+
+    def find(self, query):
+        return self.dmesg_logs_collection.find(query)
+
+    def count_documents(self, query):
+        return self.dmesg_logs_collection.count_documents(query)
+
+def run_cmd(cmd):
+    output = subprocess.check_output(cmd).strip()
+    return output.decode('utf-8')
+
+def get_device_desc():
+    cmdoutput = run_cmd(['fw_printenv', 'DeviceDesc'])
+    return cmdoutput.split('=')[1]
+
+class KernelVersion:
+    def __init__(self):
+        self.uname = run_cmd(['uname', '-r'])
+        kernel_version = re.search(r'((\d+\.\d+)\.\d+-rt\d+)(-next)?', self.uname)
+        self.full = kernel_version.group(1)
+        self.major_minor = kernel_version.group(2)
+        self.type = 'next' if kernel_version.group(3) else 'current'
+
+def get_architecture():
+    return run_cmd(['uname', '-m'])
+
+def get_os_version():
+    with open('/etc/os-release', 'rb') as file, mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ) as mfile:
+        build_id = re.search(br'BUILD_ID=\"(.*)\"', mfile)
+        return build_id.group(1).decode()
+
+def get_dmesg_log():
+    dmesg_log = run_cmd(['dmesg'])
+    return dmesg_log
+
+def upload_log(db, dmesg_log):
+    data = {}
+    kernel_version = KernelVersion()
+    data['kernel_version_full'] = kernel_version.full
+    data['kernel_version_major_minor'] = kernel_version.major_minor
+    data['kernel_type'] = kernel_version.type
+    data['device_desc'] = get_device_desc()
+    data['architecture'] = get_architecture()
+    data['os_version'] = get_os_version()
+    data['date'] = str(datetime.datetime.now())
+
+    header = ''
+    for key, val in data.items():
+        header = header + '# {}: {}\n'.format(key, val)
+
+    data['dmesg_log'] = header + dmesg_log
+
+    record = db.insert(data)
+    print('INFO: Uploaded current dmesg log from {} with _id {}'.format(data['date'], record.inserted_id))
+
+def strip_headers(dmesg_log):
+    output = ''
+    for line in dmesg_log.splitlines():
+        if not line.startswith('#'):
+            output = output + line + '\n'
+    return output
+
+def get_old_dmesg_log(db):
+    query = {}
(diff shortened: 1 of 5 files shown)


recipes-kernel/kernel-tests/kernel-tests.bb
```diff
@@ -2,7 +2,7 @@ SUMMARY = "Linux kernel-specific tests"
 HOMEPAGE = "https://kernel.org"
 SECTION = "tests"
 LICENSE = "GPLv2 & GPLv2+"
-LIC_FILES_CHKSUM = "file://run-ptest;md5=0c0d48c031233bf837da6482d67473ae"
+LIC_FILES_CHKSUM = "file://run-ptest;md5=7c0a0ce46f137996a104bdfa677dbb44"
```
bstreiff: This doesn't seem correct; the `run-ptest` file doesn't contain license text at all.

Should this just be `"file://${COMMON_LICENSE_DIR}/GPL-2.0-or-later;md5=fed54355545ffd980b814dab4a3b312c`?

(This is probably logically a separate commit.)

bstreiff (changes requested):

recipes-kernel/kernel-tests/kernel-tests.bb
```diff
@@ -82,4 +86,8 @@ do_install_ptest:append:xilinx-zynqhf() {
     cp ${WORKDIR}/required_kernel_modules.arm ${D}${PTEST_PATH}/required_kernel_modules
 }
 
+pkg_postinst_ontarget:${PN}-ptest:append() {
+    python3 -m pip install pymongo
```
amstewart: Why aren't we just distributing mongodb from the distro feeds? Licensing?

recipes-kernel/kernel-tests/kernel-tests.bb
chaitu236: Fixed

recipes-kernel/kernel-tests/kernel-tests.bb
chaitu236: I hadn't realized that the feeds already contained `python3-pymongo`. Updated the PR to use it instead.

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
```diff
@@ -0,0 +1,157 @@
...
+def get_old_dmesg_log(db):
+    query = {}
+    query['kernel_version'] = {'$ne': get_kernel_version()}
+    query['kernel_type'] = 'next' if is_next_kernel() else 'current'
+    query['device_desc'] = get_device_desc()
+    results = db.find(query).sort('date', pymongo.DESCENDING).limit(1)
+    if not results:
+        return None
+
+    for result in results:
```
gratian: This for loop reads weird to me since we're returning with the first 'result' in 'results' i.e. do we need the loop?

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
```diff
@@ -0,0 +1,157 @@
...
+def prepare_log_for_diff(log):
+    log = strip_timestamps(log)
+    log = strip_known_differences(log)
+    log = log.splitlines()
+    log.sort()
+    return log
+
+def diff_logs(current_log, old_log):
+    current_log = prepare_log_for_diff(current_log)
+    old_log = prepare_log_for_diff(old_log)
```
gratian, line 153: Is the test going to fail if we don't have a 'old_log' to compare against? Maybe that's OK for the first test run.

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
```diff
@@ -0,0 +1,157 @@
...
+    output = ''
+    for line in dmesg_log.splitlines():
+        if not line.startswith('#'):
+            output = output + line + '\n'
+    return output
+
+def get_old_dmesg_log(db):
+    query = {}
+    query['kernel_version'] = {'$ne': get_kernel_version()}
+    query['kernel_type'] = 'next' if is_next_kernel() else 'current'
```
gratian: In theory you should be able to compare against logs from '-next' kernels too. For example when we upgrade the current 5.15 kernel to the 'next' 6.0 one we should be able to use the '-next' logs already stored for 6.0 because of '-next' testing.

I'm also wondering if the query should be restricted to logs from the same 'major.minor' set as the current kernel. We might get failures just because dmesg output changes between major kernel versions.

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
chaitu236: `results` is a generator and if we don't want to use the loop method can use `next(results)` but this fails if there are no results - so it would require putting a try catch block which ends up complicating the logic even more. This seemed the simplest option to me.

But if someone knows a better way, I'm happy to change this.

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
chaitu236: Yes, if there isn't an old log, `old_log` is empty string and the test will fail as diff will be non-empty.

I'm planning to seed the db with 22.5 BSI's kernel log. But the test will likely fail - at least the first time anyway as there are so many changes and it is not possible to put all those in allow list.

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
chaitu236: I'll try to make this more intelligent and pick existing `major.minor` log if it isn't too involved.

Don't the `major.minor` versions change every couple of months? I think it would still be valuable to diff against any old log if the one with same `major.minor` isn't available - the test will fail but we can still manually vet the diff from the test output.

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
gratian: OK, I think it's fine.

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
gratian: I see your point  - for 'next' it will change every couple of months, for the stable shipping kernel is less often but yes it would still be valuable to see the diff.

chaitu236: Updated to prefer last log with same kernel major.minor version, but if not then it'll pick last log of same kernel type (current/next).

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
chaitu236: Modified the implementation to account for the new 'prefer same major.minor' logic and got rid of this loop.

gratian (approved):

bstreiff (approved):

amstewart (approved):

PR #492, opened 2022-11-01
packagegroups: Add back removed tests to the smoke test packagegroup

jeminor:
These tests are fast and give us good feedback, no reason not to just have them be part of our smoke testing.

Signed-off-by: James Minor <james.minor@ni.com>


shruthi-ravi (approved):
We need to ensure the kernel-test events on cats-slave are deleted when the functional ptest events are created.

amstewart (approved):

PR #499, opened 2022-11-18
[kirkstone] syslog-ng: Update conf file to use recipe version

SparkingSpork:
kirkstone is now shipping syslog-ng version 3.36.1. The conf file was still version 3.31 which caused a warning to print at boot. This updates the version tag in the conf file to use the recipe version to resolve the warning. This should prevent the need to update the recipe every release.

AzDO Work Item: https://dev.azure.com/ni/DevCentral/_workitems/edit/2212522

## Testing:
Build base system image locally with the change and installed to a target. Confirmed the warning no longer appears.

### DIFF
--- a/recipes-support/syslog-ng/files/syslog-ng.conf.sysvinit
+++ b/recipes-support/syslog-ng/files/syslog-ng.conf.sysvinit
@@ -1,4 +1,4 @@
-@version: 3.31
+@version: @VERSION@
 # syslog-ng config file for NI RT targets
 
 options {
--- a/recipes-support/syslog-ng/syslog-ng_3.%.bbappend
+++ b/recipes-support/syslog-ng/syslog-ng_3.%.bbappend
@@ -28,6 +28,13 @@ do_install:append (){
    install -m 0644 ${WORKDIR}/logrotate.d-auth.conf ${D}${sysconfdir}/logrotate.d/auth.conf
    install -m 0644 ${WORKDIR}/logrotate.d-cron.conf ${D}${sysconfdir}/logrotate.d/cron.conf
    install -m 0644 ${WORKDIR}/logrotate.d-messages.conf ${D}${sysconfdir}/logrotate.d/messages.conf
+
+   # NILRT has a custom conf file for sysvinit systems.
+   # If using sysvinit, update the conf file with the current version.
+   if ${@bb.utils.contains('DISTRO_FEATURES','systemd','false','true',d)}; then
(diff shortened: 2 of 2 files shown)


recipes-support/syslog-ng/files/syslog-ng.conf.sysvinit
```diff
@@ -1,4 +1,4 @@
-@version: 3.31
+@version: 3.36
```
shruthi-ravi: Is it possible to have a placeholder version text here, which gets replaced with the recipe version at build time? Then, we wouldn't have to update this every release.

SparkingSpork: I like that idea. I'll take a stab at it.

recipes-support/syslog-ng/syslog-ng_3.%.bbappend
```diff
@@ -28,6 +28,12 @@ do_install:append (){
    install -m 0644 ${WORKDIR}/logrotate.d-auth.conf ${D}${sysconfdir}/logrotate.d/auth.conf
    install -m 0644 ${WORKDIR}/logrotate.d-cron.conf ${D}${sysconfdir}/logrotate.d/cron.conf
    install -m 0644 ${WORKDIR}/logrotate.d-messages.conf ${D}${sysconfdir}/logrotate.d/messages.conf
+
+   # NILRT has a custom conf file for sysvinit systems.
+   # If using sysvinit, update the conf file with the current version.
+   if ${@bb.utils.contains('DISTRO_FEATURES','systemd','false','true',d)}; then
+      sed -i -e "s/@VERSION@/$(echo ${PV} | grep -oE ^[0-9]+.[0-9]+)/g" ${D}${sysconfdir}/${BPN}/${BPN}.conf
```
SparkingSpork, line 36: `syslog-ng` expects the version string to be `MAJ.MIN` only. Any suggestions on a better way to extract that from ${PV} are welcome.

shruthi-ravi (approved):

recipes-support/syslog-ng/syslog-ng_3.%.bbappend
shruthi-ravi: I don't know of a better way. But if you stick to this method, I'd suggest adding a comment. It took me a minute to figure out why you were doing it, and that there was a .patch version you didn't want.

SparkingSpork: Done.

amstewart (approved):

PR #503, opened 2022-12-01
nilrt.conf: squash meta-security warning 

dylan-matthew-turner:
As per the meta-security README for kirkstone, various bbs require that the "security" feature be enabled now.

This appears to be a change from hardknott, so it makes sense to now enable it

### DIFF
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -15,6 +15,8 @@ VIRTUAL-RUNTIME_xserver_common = "xserver-common"
 
 VIRTUAL-RUNTIME_mountpoint ?= "busybox"
 
+SKIP_META_SECURITY_SANITY_CHECK = "1"
+
 # on older NILRT distro flavors the kernel is installed in non-standard paths
 # for backward compatibility
 KERNEL_IMAGEDEST = "boot/runmode"


amstewart:
What are all the recipe changes that are affected by enabling this DISTRO_FEATURE? Are any of the recipes that we're including in the distro going to change?

dylan-matthew-turner:
@amstewart 

This is what it says in the README:

```
The bbappend files for some recipes (e.g. linux-yocto) in this layer need
to have 'security' in DISTRO_FEATURES to have effect.
```

There is no list I could find of recipes, but using `grep` to find references to the feature, it seems to only be relevant to "sanity-meta-security" which may just be to generate the warning:

```
WARNING: You have included the meta-security layer, but 'security' has not been enabled in your DISTRO_FEATURES. Some bbappend files and preferred version setting may not take effect. See the meta-security README for details on enabling security support.
```

If it's not affecting many packages, we may just want to disable the feature requirement instead via adding `SKIP_META_SECURITY_SANITY_CHECK = 1`

amstewart:
Grepping the whole sources/ dir for `DISTRO_FEATURES.*security` provides one result-of-interest.
```
meta-security/recipes-kernel/linux/linux-yocto_5.%.bbappend:1:require ${@bb.utils.contains('DISTRO_FEATURES', 'security', '${BPN}_security.inc', '', d)}
```

[That `linux-yocto_security.inc`](https://github.com/ni/meta-security/blob/c79262a30bd385f5dbb009ef8704a1a01644528e/recipes-kernel/linux/linux-yocto_security.inc) file enables some kernel features like `apparmor` and `smack`, neither of which we want. But that would only be in the case that we actually used the `linux-yocto` recipe, instead of overwriting it wholesale with the `linux-nilrt` recipes in meta-nilrt. So even if we enable the `security` distro feature, I don't think the above kernel recipe logic would be executed.

That being said, since there doesn't seem to be an issue with keeping the distro feature off, I would prefer to keep it that way. Is setting `SKIP_META_SECURITY_SANITY_CHECK` the only thing we need to do to squash the warning?

dylan-matthew-turner:
I just tried it (though I had to do SKIP_META_SECURITY_SANITY_CHECK = "1" not just 1) and it appears to remove the warning

amstewart (approved):

PR #504, opened 2022-12-07
Add irq_work priority and fix kthread_test_priority.sh test

chaitu236:
kthread_test_priority.sh expects threads starting with "irq" to have priority 15. In older kernels (i.e., 4.14) this was true as only "irq/x" threads were present. But on newer kernels (5.10, 5.15, etc), there are also "irq_work/x" threads which have priority 1 which fails the test.

So fixed the test to check for "irq/" instead of "irq", added priority for irq_work in rtgroups and added a test for irq_work.

### Testing
The test is passing on a cRIO-9042 with current and next kernels with these changes.

### DIFF
--- a/recipes-rt/rt-tests/files/kthread_test_priority.sh
+++ b/recipes-rt/rt-tests/files/kthread_test_priority.sh
@@ -43,7 +43,7 @@ check_prio_for_task ksoftirqd/0 FIFO 8
 ptest_report
 
 ptest_change_subtest 3 irq
-check_prio_for_task irq FIFO 15
+check_prio_for_task irq/ FIFO 15
 ptest_report
 
 exit $rc_first
--- a/recipes-rt/rt-tests/files/kthread_test_priority.sh
+++ b/recipes-rt/rt-tests/files/kthread_test_priority.sh
@@ -46,4 +46,8 @@ ptest_change_subtest 3 irq
 check_prio_for_task irq/ FIFO 15
 ptest_report
 
+ptest_change_subtest 4 irq_work
+check_prio_for_task irq_work/ FIFO 12
+ptest_report
+
 exit $rc_first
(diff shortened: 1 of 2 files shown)


amstewart (approved):

recipes-rt/rt-tests/files/kthread_test_priority.sh
```diff
@@ -43,7 +43,7 @@ check_prio_for_task ksoftirqd/0 FIFO 8
 ptest_report
 
 ptest_change_subtest 3 irq
-check_prio_for_task irq FIFO 15
+check_prio_for_task irq/ FIFO 15
```
gratian, line 46: Should we also add a check for the priority of 'irq_work' threads since we are setting an explicit priority for them now?

chaitu236: Yes I'm already checking for irq_work a few lines below.

gratian: Huh, I must be looking at the wrong thing or the github UI doesn't show it. The next line down for me is `ptest_report` followed by `exit $rc_first`. Anyway if it's there then never mind.

gratian (approved):

recipes-rt/rt-tests/files/kthread_test_priority.sh
chaitu236: Huh, thats weird. Here's how it looks like for me

image

gratian: Maybe I was looking at the wrong commit. Anyway I see it now and looks good.

shruthi-ravi:
@chaitu236 can you cherry-pick these changes to the release branch?

chaitu236:
> @chaitu236 can you cherry-pick these changes to the release branch?

https://github.com/ni/meta-nilrt/pull/511

chaitu236:
> > @chaitu236 can you cherry-pick these changes to the release branch?
> 
> #511

Sorry, I realized we only want these changes in master for now as we want this to undergo client testing.

PR #505, opened 2022-12-07
util-linux: busybox dependency to squash warning

dylan-matthew-turner:
Getting a warning where util-linux-hwclock (which is RPROVIDES by util-linux) depends on busybox-hwclock (RDEPENDS by busybox):

```
WARNING: util-linux-2.37.4-r0 do_package_qa: QA Issue: util-linux-hwclock rdepends on busybox-hwclock, but it isn't a build dependency, missing busybox in DEPENDS or PACKAGECONFIG? [build-deps]
```

Adding a dependency here squashes warning

### DIFF
--- a/recipes-core/util-linux/util-linux_2.%.bbappend
+++ b/recipes-core/util-linux/util-linux_2.%.bbappend
@@ -1,6 +1,6 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
-DEPENDS:append:class-target = " shadow-native pseudo-native"
+DEPENDS:append:class-target = " shadow-native pseudo-native busybox"
 
 RDEPENDS:${PN}-hwclock:append = " niacctbase busybox-hwclock"
 RDEPENDS:${PN}-ptest += "${PN}-nilrt-ptest"


amstewart:
That doesn't seem right. `util-linux-hwclock` should replace `busybox-hwclock`, not depend on it. What is the reason for the dependency? Is it the `/etc/init.d/hwclock` initscript?

dylan-matthew-turner:
It is due to the util-linux_2.%.bbappend file containing: `RDEPENDS:${PN}-hwclock:append = " niacctbase busybox-hwclock"`

dylan-matthew-turner:
I'll try removing the `busybox-hwclock` dependency as an alternative

amstewart:
Yeah; I think that's for the `/etc/init.d/hwclock` script. Maybe we can't get rid of the RDEPENDS, but I'm still not sure that the `:class-target` is the correct override for the `DEPENDS`.

dylan-matthew-turner:
Removing `busybox-hwclock` from the `RDEPENDS` doesn't appear to cause any errors when running `bitbake util-linux`

amstewart:
It wouldn't; it's a runtime dependency.

amstewart (approved):

amstewart:
We discussed this PR offline, and I'm now in agreement.

PR #464, opened 2022-09-21
[sumo] linux: Staple kernel recipes to 22.8 branches and hashes.

SparkingSpork:
Stapling the sumo master branch to the 22.8 ni/linux release branches to reduce long-term overhead.

## Testing: 
Build linux-nilrt locally.

PR #508, opened 2022-12-07
[academic] linux: set 23.0 release KBRANCH

amstewart:
Service for the 23Q1 release.

### DIFF
--- a/recipes-kernel/linux/linux-nilrt-debug_git.bb
+++ b/recipes-kernel/linux/linux-nilrt-debug_git.bb
@@ -1,5 +1,5 @@
 DESCRIPTION = "NILRT linux kernel debug build"
-NI_RELEASE_VERSION = "master"
+NI_RELEASE_VERSION = "23.0"
 LINUX_VERSION = "5.10"
 LINUX_VERSION_xilinx-zynq = "4.14"
 LINUX_VERSION_qemu-zynq7 = "4.14"
--- a/recipes-kernel/linux/linux-nilrt-next_git.bb
+++ b/recipes-kernel/linux/linux-nilrt-next_git.bb
@@ -1,5 +1,5 @@
 DESCRIPTION = "NILRT linux kernel next development build"
-NI_RELEASE_VERSION = "master"
+NI_RELEASE_VERSION = "23.0"
 LINUX_VERSION = "5.15"
 LINUX_KERNEL_TYPE = "next"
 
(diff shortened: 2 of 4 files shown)


recipes-kernel/linux/linux-nilrt-debug_git.bb
```diff
@@ -1,5 +1,5 @@
 DESCRIPTION = "NILRT linux kernel debug build"
-NI_RELEASE_VERSION = "master"
+NI_RELEASE_VERSION = "23.0"
 LINUX_VERSION = "5.10"
 LINUX_VERSION_xilinx-zynq = "4.14"
 LINUX_VERSION_qemu-zynq7 = "4.14"
```
shruthi-ravi, line 5: SRCREV is uncommented below and set to a hash. What happens when both that and AUTOREV are set? The SRCREV is usually commented out. Not sure why they aren't in all of the sumo files.

amstewart: Oh; that's right. We were intending to keep the academic distro using the 4.14 master branch, but with a stapled commit - since we don't expect to add commits to that mainline.

If we're still OK with that plan, I can just drop this PR.

shruthi-ravi: I'll leave it to @gratian. Did you intend that with the sumo release branch as well?

amstewart: No; just academic.

amstewart:
We [permanently stapled the ARM mainlines intentionally](https://github.com/ni/meta-nilrt/pull/464), so this PR isn't needed.

recipes-kernel/linux/linux-nilrt-debug_git.bb
gratian: The academic 4.14 kernel branch: https://github.com/ni/linux/commits/nilrt-academic/master/4.14 does have a couple of extra commits added recently.

Not sure if we want to leave recipes pointing to master and update the hash or create a 23.0 kernel branch.

amstewart: We spoke about this offiline, but for web: the academic meta-nilrt branches are stapled to a kernel master ref commit which is current with the changes gratian linked above. So, there is no need to make a new kernel branch for it.

PR #513, opened 2022-12-17
Drop rauc support

dylan-matthew-turner:
Rauc is creating a warning:
```
WARNING: rauc-1.8-r0 do_install: Please overwrite example ca.cert.pem with a project specific one, or set the RAUC_KEYRING_FILE variable with your file! 
```
Since we are not currently using its functionality, we should drop it.

These commits prevent rauc from being built


jeminor:
Please hold off on this commit for a bit until after you've talked with me.

amstewart (approved):

PR #517, opened 2022-12-29
[hardknott] Revert "Revert "opkg: populate keys even if creation data is in the future""

SparkingSpork:
This reverts commit ccae7a71997dc5e25c42dc98d9fae3d9d751b5e4.

AzDO#2251378

opkg-key currently calls gpg with the --no-options flag, which results in /etc/opg/gpg/gpg.conf not being applied with the ignore-time-conflict flag on NI devices. This patch previously applied the flag directly in opkg-key so temporarily reverting it until a long-term fix can be identified.

AzDO Work Item: https://dev.azure.com/ni/DevCentral/_workitems/edit/2251378/

## Testing: 
Built and installed a new BSI with the changes and confirmed that opkg update worked. Messages about keys being created in the future are still printed but the actual `opkg-key populate` call succeeds now and `opkg update` works on initial boot. 

## Note:
This will need cherry-picked into the 23.0 branch and we'll need to update out release build to use the new commit.

### DIFF
--- a/recipes-devtools/opkg/files/0001-opkg-key-add-keys-even-if-creation-date-is-in-the-fu.patch
+++ b/recipes-devtools/opkg/files/0001-opkg-key-add-keys-even-if-creation-date-is-in-the-fu.patch
@@ -1,0 +1,31 @@
+From e33cbfc0c983fda5c4efa13444efb8584dde7d2b Mon Sep 17 00:00:00 2001
+From: Alejandro del Castillo <alejandro.delcastillo@ni.com>
+Date: Fri, 22 Feb 2019 11:35:04 -0600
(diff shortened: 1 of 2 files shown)


shruthi-ravi:
@SparkingSpork can you change the target branch for the PR to the release branch instead? We decided to carry this patch only in the release branch for now. 
Also, can you add an azdo bug tag to the commit description?

SparkingSpork:
> @SparkingSpork can you change the target branch for the PR to the release branch instead? We decided to carry this patch only in the release branch for now.
Also, can you add an azdo bug tag to the commit description?

Done.

shruthi-ravi (approved):

PR #521, opened 2023-01-05
Add patch to un-uniquify elfutils ptest output

mike-petersen-ni:
As noted in [AB#2232487](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2232487), the PID-uniquified file name here causes our test review system to think it's a new failure for each run.
This change adds a patch to not print out  that unique name.

(Moved from ni/openembedded-core#86.)


chaitu236 (approved):

amstewart (approved):

amstewart:

PR #523, opened 2023-01-12
[kirkstone] Set the feed name to 2023Q2

mike-petersen-ni:
This was set to "next" in 23a181e8663d3df2670d050d5cd0190c810e8d8e but should be set to the usual Y+Q format so that the correct feeds can be pulled.

[AB#2105165](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2105165)

### DIFF
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -4,7 +4,7 @@ DISTRO_VERSION = "10.0"
 
 DISTRO_CODENAME = "${LAYERSERIES_COMPAT_meta-nilrt}"
 
-NILRT_FEED_NAME ?= "next"
+NILRT_FEED_NAME ?= "2023Q2"
 
 DISTRO_FEATURES:append:x64 = "\
         x11 \


conf/distro/nilrt.conf
```diff
@@ -4,7 +4,7 @@ DISTRO_VERSION = "10.0"
 
 DISTRO_CODENAME = "${LAYERSERIES_COMPAT_meta-nilrt}"
 
-NILRT_FEED_NAME ?= "next"
+NILRT_FEED_NAME ?= "2023Q2"
```
chaitu236, line 7: Shouldn't this be "next/2022Q2" so that devs building locally can access the feeds

mike-petersen-ni: When I pinged @amstewart he said it should be like hardknott. Maybe I misunderstood; I had thought that the changes in AzDO to add the `next/` prefix would be enough since nilrt.inc here is using download.ni.com instead of nickdanger.

chaitu236: `ni-org.conf` (included when `--org` arg is used for `ni-oe-init-build-env`) sets feeds to nickdanger.

mike-petersen-ni: Should the `next/` prefix should be added in that file, then?

amstewart: The server domain set by the org.conf files isn't relevant here. None of that logic is changing for the kirkstone mainline.

This variable only sets the default feed namespace in the OE-built images. So it is largely inconsequential for local builders, because they will be building the feed contents locally and would have to write their own opkg feed configs from scratch, if they want to do a local package installation.

The build servers [set this value](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/rtos/includes/bitbakeRules.mak&version=GBmain&line=105&lineEnd=106&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) to what is coordinated through the AZDO `REPO_EXTERNAL_VERSION_NOSPACE` variable. So it is also overwritten there.

The more I think about this variable, the more I think we might just want to deprecate it. @mike-petersen-ni Maybe we just drop this PR for now and let it stay on `next`.

PR #524, opened 2023-01-19
kernel-tests: handle onboard serial devices connected via ISA bridge

bstreiff:
Serial drivers enumerated through the platform_driver mechanism instead of the legacy PNP mechanism can show up as children of a PCI-based ISA bridge (for example, on cRIO-904x, the RS-485 UART shows up as a child of the "Atom E3900 Series Low Pin Count Interface" device). Refactor the "onboard" check into a separate function and consider those devices to be "onboard" too.


amstewart (approved):

amstewart:

PR #525, opened 2023-01-26
[kirkstone] kernel-tests: Include OS version in dmesg log compare criteria.

SparkingSpork:
Currently, if there are two OS versions running in the ATS the test_kernel_dmesg_diff test will use the latest dmesg log in the database regardless of which OS version it came from. This can result in failures due to changes between the OS versions when there are overlaps.

This change updates the test python script to prefer logs from the same OS version when available.

[AB#2279632](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2279632)

## Testing:
Installed to a cRIO-9053 and hand-tested by running the ptest. I edited the database entry created by the first run by hand to test different scenarios such as:
- No os_version_major_minor
- Matching and mismatching os_version_major_minor
- Same and different kernel versions

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
+++ b/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
@@ -59,13 +59,15 @@ class KernelVersion:
         self.major_minor = kernel_version.group(2)
         self.type = 'next' if kernel_version.group(3) else 'current'
 
+class OsVersion:
+    def __init__(self):
+        with open('/etc/os-release', 'rb') as file, mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ) as mfile:
+            build_id = re.search(br'BUILD_ID=\"((\d+\.\d+).*)\"', mfile)
+            self.full = build_id.group(1).decode()
+            self.major_minor = build_id.group(2).decode()
+
 def get_architecture():
     return run_cmd(['uname', '-m'])
-
-def get_os_version():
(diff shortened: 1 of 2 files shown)


gratian (approved):

SparkingSpork:
After speaking with @chaitu236, we decided to flip the logic to prefer dropping the os_version check before dropping the kernel_version check. This is technically a kernel test so that should be the preference. I've updated the commit accordingly.

chaitu236 (approved):

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
```diff
@@ -59,27 +59,31 @@ def __init__(self):
         self.major_minor = kernel_version.group(2)
         self.type = 'next' if kernel_version.group(3) else 'current'
 
+class OsVersion:
+    def __init__(self):
+        with open('/etc/os-release', 'rb') as file, mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ) as mfile:
```
amstewart, line 64: (MINOR) Could you [add an RDEPENDS](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/recipes-kernel/kernel-tests/kernel-tests.bb#L14) on `os-release`?

SparkingSpork: Done.

amstewart (approved):

PR #531, opened 2023-02-10
[kirkstone] kernel-tests: Use semantic version comparison in dmesg-diff.

SparkingSpork:
The comparison to check the kernel version in the dmesg log database was actually using a lexicographical compare. This resulted in the comparison mistakenly using newer kernel versions to compare to instead of older.

By adding logic to parse the kernel semantic versioning scheme, this change allows for the comparison to behave as intended.

[AB#2282279](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2282279)

## Note to Reviewers:
Before this is cherry-picked to hardknott, the new fields will be manually added to each of the most recent dmesg logs in the database for each of the most recent kernels.

## Testing:
Hand tested locally and by hand modifying the database. This included testing changing which field in the kernel version was higher than/equal to/less than.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
+++ b/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
@@ -51,13 +51,24 @@ def get_device_desc():
     cmdoutput = run_cmd(['fw_printenv', 'DeviceDesc'])
     return cmdoutput.split('=')[1]
 
-class KernelVersion:
+class SemanticVersion:
+    def __init__(self, versionString):
+        # Regex below copied from https://semver.org/.
+        version = re.search(r'^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$', versionString)
+        self.full = version.group()
+        self.major = int(version.group('major'))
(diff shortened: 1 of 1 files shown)


recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
```diff
@@ -106,9 +117,10 @@ def get_old_dmesg_log(db, logger):
     query = {}
     kernel_version = KernelVersion()
     os_version = OsVersion()
-    query['kernel_version_full'] = {'$lt': kernel_version.full}
+    query['kernel_version'] = { '$lt': kernel_version.version_dict }
```
SparkingSpork, line 120: This works because of the order mongodb queries perform comparisons. It will actually go through the fields in order and compare each one to decide if the next one needs evaluated or not.

amstewart (approved):
I'll let you merge it this time. :)

chaitu236 (approved):

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
```diff
@@ -51,13 +51,24 @@ def get_device_desc():
...
+class KernelVersion(SemanticVersion):
     def __init__(self):
-        self.uname = run_cmd(['uname', '-r'])
-        kernel_version = re.search(r'((\d+\.\d+)\.\d+-rt\d+)(-next)?', self.uname)
-        self.full = kernel_version.group(1)
-        self.major_minor = kernel_version.group(2)
-        self.type = 'next' if kernel_version.group(3) else 'current'
+        version =run_cmd(['uname', '-r'])
+        SemanticVersion.__init__(self, version)
+        self.type = 'next' if self.build_metadata else 'current'
```
jpautler, line 69: Assuming that the NO_HZ kernel puts something unique in this field, then I guess this logic wouldn't work on that kernel or any other kernel variants we might dream up in the future. It's not a requirement now, so I don't think we necessarily need to change anything. Just wanted to point it out.

jpautler (approved):
Pending your review of my one comment, I approve.

PR #534, opened 2023-02-17
ni-base-system-image-tests: Add filesystem permissions ptest

mike-petersen-ni:
As part of [AB#2232369](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2232369), we want to notice when filesystem permissions change. Writing this as a ptest improves maintainability over the alternative of creating a standalone test, and allows future improvements (e.g. for isolation) to be easily utilized.

This ptest is not added to the ptest packagegroup, as the installation of other packages will alter its results.

# Behavior of the new test
This test is similar to the dmesg diff test at `recipes-kernel/kernel-tests`.

This test collects a manifest using `find`, using its `-printf` expression to generate a table in tab-separated-value format. To compare the tables, they are first converted to dictionaries and compared in Python, rather than using `diff` as in the dmesg diff test.

## Determining the previous manifest
Unlike the dmesg diff test, the kernel version is irrelevant. Instead, the OS version (retrieved from `/etc/os-release`) is used to fulfill the purpose. Since multiple variants may have similar versions (e.g. hardknott and kirkstone are now both building for 23.3.0), the codename is also extracted and used to distinguish between versions.

The version of the previous manifest must:
 - have major/minor/patch components that are lexically less than the current version
 - have a phase of "final"

It will be necessary to manually run this test on the last release version to create such a manifest to compare against.

## Directories examined
As seen in `fs_permissions_diff.py`:

```py
search_dirs = ['/bin', '/boot', '/etc', '/lib', '/lib64', '/sbin', '/usr', '/var']
omit_dirs = ['/lib/modules', '/var/cache', '/var/run', '/var/tmp', '/var/volatile']
```

Let me know if these lists should change (e.g. exclude `/etc/natinst/share`, include `/home` or `/tmp`).

## Handling of symbolic links
The test does not resolve symbolic links as their targets, though it does record the link target.  
Changing a link to/from a non-link will be noted as a difference, as that is part of the mode.  
Targets of links are not compared.

## Handling of added/removed paths
All paths present in the previous manifest but missing in the current one (or vice versa) are printed in the log and fail the test.

# Miscellaneous changes
 - dmesg diff test:
   - Removed unused import
   - Modified strip_headers to avoid bad asymptotic behavior

# Testing
 - [x] Built with `bitbake ni-base-system-image-tests`
 - [x] Installed on a NILRT VM with `opkg install ni-base-system-image-tests-ptest`
 - Ran `ptest-runner ni-base-system-image-tests` in various scenarios
   - [x] Verified that the manifest was uploaded to the Mongo database
   - Reran after changing content in `os-release` to...
      - [x] a later version; verified that the previous run was diffed against
      - [x] a different codename; verified that the previous run was *not* diffed against
    - [x] After changing permissions (mode, user, or group) on `/bin`, the test fails
    - [x] After adding files (equivalent: running without a previous manifest), the test fails
    - [x] After removing files, the test fails

# Merging
I think it is appropriate to cherry-pick this change into nilrt/master/hardknott.

### DIFF
--- a/recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
+++ b/recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
@@ -1,0 +1,240 @@
+import argparse
+import datetime
+import difflib
+import hashlib
+import mmap
+import os
+import pymongo
+import re
+import subprocess
+import sys
+
+class DB:
+    mongo_db_name = 'rtos'
+    mongo_db_collection_name = 'dmesg_logs'
+
+    def __init__(self, server, user, password):
+        mongo_client = pymongo.MongoClient("mongodb://{user}:{pwd}@{server}".format(
+            user=user, pwd=password, server=server
+            ),
+            w=1)
+        db = mongo_client[self.mongo_db_name]
+        self.dmesg_logs_collection = db[self.mongo_db_collection_name]
+
+    def insert(self, data):
+        return self.dmesg_logs_collection.insert_one(data)
+
+    def find(self, query):
+        return self.dmesg_logs_collection.find(query)
+
+    def count_documents(self, query):
+        return self.dmesg_logs_collection.count_documents(query)
+
+class Logger:
+    logs = []
+    def log(self, log):
+        self.logs.append(log)
+
+    def first_log(self, log):
+        self.logs.insert(0, log)
+
+    def report(self):
+        for log in self.logs:
+            print(log)
+
+def run_cmd(cmd):
+    output = subprocess.check_output(cmd).strip()
+    return output.decode('utf-8')
+
+def get_device_desc():
+    cmdoutput = run_cmd(['fw_printenv', 'DeviceDesc'])
+    return cmdoutput.split('=')[1]
+
+class SemanticVersion:
+    def __init__(self, versionString):
+        # Regex below copied from https://semver.org/.
+        version = re.search(r'^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$', versionString)
+        self.full = version.group()
+        self.major = int(version.group('major'))
+        self.minor = int(version.group('minor'))
+        self.patch = int(version.group('patch'))
+        self.prerelease = version.group('prerelease')
+        self.build_metadata = version.group('buildmetadata')
+
+class KernelVersion(SemanticVersion):
+    def __init__(self):
+        version =run_cmd(['uname', '-r'])
+        SemanticVersion.__init__(self, version)
+        self.type = 'next' if self.build_metadata else 'current'
+        prerelease_version = int(re.search(r'0|[1-9]\d*', self.prerelease).group())
+        self.version_dict = { 'major': self.major, 'minor': self.minor, "patch": self.patch, 'prerelease': prerelease_version }
+
+class OsVersion:
+    def __init__(self):
+        with open('/etc/os-release', 'rb') as file, mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ) as mfile:
+            build_id = re.search(br'BUILD_ID=\"((\d+\.\d+).*)\"', mfile)
+            self.full = build_id.group(1).decode()
+            self.major_minor = build_id.group(2).decode()
+
+def get_architecture():
+    return run_cmd(['uname', '-m'])
+
+def get_dmesg_log():
+    dmesg_log = run_cmd(['dmesg'])
+    return dmesg_log
+
+def upload_log(db, dmesg_log, logger):
+    data = {}
+    kernel_version = KernelVersion()
+    os_version = OsVersion()
+    data['kernel_version_full'] = kernel_version.full
+    data['kernel_version'] = kernel_version.version_dict
+    data['kernel_type'] = kernel_version.type
+    data['device_desc'] = get_device_desc()
+    data['architecture'] = get_architecture()
+    data['os_version'] = os_version.full
+    data['os_version_major_minor'] = os_version.major_minor
+    data['date'] = str(datetime.datetime.now())
+
+    header = ''
+    for key, val in data.items():
+        header = header + '# {}: {}\n'.format(key, val)
+
+    data['dmesg_log'] = header + dmesg_log
+
+    record = db.insert(data)
+    logger.log('INFO: Uploaded dmesg log of "{}" kernel {} from {} with _id {}'.format(data['kernel_type'], data['kernel_version_full'], data['date'], record.inserted_id))
+
+def strip_headers(dmesg_log):
+    return ''.join(line + '\n' for line in dmesg_log.splitlines() if not line.startswith('#'))
+
+def get_old_dmesg_log(db, logger):
+    query = {}
+    kernel_version = KernelVersion()
+    os_version = OsVersion()
+    query['kernel_version'] = { '$lt': kernel_version.version_dict }
+    query['device_desc'] = get_device_desc()
+    query['kernel_version.major'] = kernel_version.major
+    query['kernel_version.minor'] = kernel_version.minor
+    query['os_version_major_minor'] = os_version.major_minor
+
+    if db.count_documents(query):
+        results = db.find(query).sort('date', pymongo.DESCENDING).limit(1)
+    else:
+        # If there are no results, there may not be a run from the current os version.
+        # Remove that requirement.
+        del query['os_version_major_minor']
+        logger.log('INFO: No prior logs from this OS Version found. Using previous version.')
+
+        if db.count_documents(query):
+            results = db.find(query).sort('date', pymongo.DESCENDING).limit(1)
+        else:
+            # If there still aren't any results matching the kernel major minor version, broaden the
+            # search but limit it to same kernel type
+            del query['kernel_version.major']
+            del query['kernel_version.minor']
+            query['kernel_type'] = kernel_version.type
+
+            if db.count_documents(query):
+                results = db.find(query).sort('date', pymongo.DESCENDING).limit(1)
+            else:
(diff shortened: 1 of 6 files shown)


amstewart (approved):

chaitu236 (approved):
The changes look good but I don't understand this comment

> A squash merge is likely simplest for maintainers. The merge commit in the early history connects the histories across the file copies, but this means a rebase would require manual intervention (to avoid accidentally removing the files copied from).

Why would a rebase require manual intervention?

mike-petersen-ni:
> Why would a rebase require manual intervention?

It's an artifact of how I preserved the history across the file copies with a merge commit. Rebasing doesn't handle merge commits too well; it'd be necessary to do an interactive rebase that looks like this at the beginning:

```
label base
pick 93c79f67 Move nirtcfg-tests.bb from recipes-ni/nirtcfg-tests
pick a2ff13e7 Move kernel_dmesg_diff.py from recipes-kernel/kernel-tests
pick 0b00982e Move ptest-format.sh from recipes-kernel/kernel-tests
pick 4acfe422 Move test_kernel_dmesg_diff.sh from recipes-kernel/kernel-tests
label side
reset base
merge side
break # to manually un-delete the files with git checkout HEAD^ -- <files ...>
pick 1a84f40c Use the correct files in the recipe and tests
```

You could attempt to use `git rebase origin/nilrt/master/kirkstone --onto origin/nilrt/master/hardknott`, but you'd get a conflict later on since the files had been renamed instead of copied.

---

If y'all want to preserve my commit history here, but don't want to pull in this messy stuff, I could fix this problem by squashing the copies. I imagine we'd probably also have me edit the commit titles to have that "location: " prefix that's common in this repo.

chaitu236:
> If y'all want to preserve my commit history here, but don't want to pull in this messy stuff, I could fix this problem by squashing the copies. I imagine we'd probably also have me edit the commit titles to have that "location: " prefix that's common in this repo.

Ok, I'll defer to Alex on how we want to handle this.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -0,0 +1,215 @@
...
+            'patch': self.patch,
+            'phase': { 'd': 0, 'a': 1, 'b': 2, 'f': 3 }[self.phase],
+            'build': self.build
+        }
+
+fs_manifest_format = '%p\t%M\t%u\t%g\t%l'
+fs_manifest_columns = ['path', 'mode', 'user', 'group', 'link_target']
+def get_fs_manifest():
+    search_dirs = ['/bin', '/boot', '/etc', '/lib', '/lib64', '/sbin', '/usr', '/var']
+    omit_dirs = ['/var/cache', '/var/run', '/var/tmp', '/var/volatile']
```
chaitu236: It just occurred to me that kernel modules directory `/lib/modules/<kernelversion>` will have a new name on every update. Will this cause false positives? Do we need to omit that directory? I'm not sure if there is any benefit to indexing that directory.

mike-petersen-ni: We could omit `/lib/modules` if we wanted to. As mentioned in the PR text added and removed files (and by extension, renames like this) are only logged, and won't fail the test. Let me know of your preference.

chaitu236: If we keep it, all the files within that directory will be logged right? That'd be too much noise so I'd prefer we not keep it.

mike-petersen-ni: Makes sense. I'll push that change.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -0,0 +1,215 @@
...
+import subprocess
+import sys
+
+class DB:
+    mongo_db_name = 'rtos'
+    mongo_db_collection_name = 'fs_permissions'
+
+    def __init__(self, server, user, password):
+        mongo_client = pymongo.MongoClient("mongodb://{user}:{pwd}@{server}".format(
+            user=args.user, pwd=args.password, server=args.server
```
jpautler: Did you meant to use "args." for these values? Assuming not, it will be a little confusing that the format string placeholder names are identical to the constructor parameter names, although I think it probably works anyway.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -0,0 +1,215 @@
...
+        logger.log('INFO: Starting diff')
+        for path in different_paths:
+            logger.log('\n'.join([path,
+                                  f'\told: {detail_string(old_manifest[path])}',
+                                  f'\tnew: {detail_string(current_manifest[path])}']))
+        logger.log('INFO: End of diff')
+        return False
+
+    logger.log('INFO: Empty diff')
+    return True
```
jpautler: Do you want this function to return False if there are **any** differences (including paths added or removed)? Currently, it seems to only returns False if a path is different.

recipes-ni/ni-base-system-image-tests/ni-base-system-image-tests.bb
```diff
@@ -0,0 +1,34 @@
...
+    install -m 0644 ${WORKDIR}/ptest-format.sh        ${D}${PTEST_PATH}
+    install -m 0755 ${WORKDIR}/run-ptest              ${D}${PTEST_PATH}
+    install -m 0644 ${WORKDIR}/fs_permissions_diff.py ${D}${PTEST_PATH}
+    install -m 0755 ${WORKDIR}/test_*.sh              ${D}${PTEST_PATH}
+}
+
+# We only want to build the -ptest package
+PACKAGES:remove = "${PN}-dev ${PN}-staticdev ${PN}-dbg"
+
+RDEPENDS:${PN}-ptest:append = " bash python3-pymongo"
```
jpautler, line 34: I thought maybe there should be a dependency on ptest-runner here, but I don't see it in the recipes for other ptest packages. Looks like it is included as a dependency in the packagegroups that install our ptests. Since this test is not part of a packagegroup, you might need to include the ptest-runner package in the optional_opkg_packages event variable so that it is present on your system in order to run the test.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
mike-petersen-ni: Copied from original; I'll fix both with next push.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
mike-petersen-ni: Do we want the test to fail for added/removed paths? I don't know if it'll provide much value to get a test failure from that.

amstewart:
>

amstewart:
> > If y'all want to preserve my commit history here, but don't want to pull in this messy stuff, I could fix this problem by squashing the copies. I imagine we'd probably also have me edit the commit titles to have that "location: " prefix that's common in this repo.
> 
> Ok, I'll defer to Alex on how we want to handle this.

Yeah; I'd prefer to squash this PR down into a smaller number of commits. @mike-petersen-ni Could you take care of that.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
jpautler: Good point. Maybe a good question for @amstewart?

mike-petersen-ni: Just pushed change to consider adds and removes as failures.

recipes-ni/ni-base-system-image-tests/ni-base-system-image-tests.bb
mike-petersen-ni: Addressed in the PR for the event.

mike-petersen-ni:
# Patch v5(?)

Conversed with @chaitu236 out-of-band; rebased to squash commits down to just three: fixup, copy, new work.

chaitu236:
Cherry picked to hardknott

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -0,0 +1,239 @@
...
+        logger.log('INFO: No prior fs permissions from this OS Version found. Using previous version.')
+
+        if db.count_documents(query):
+            results = db.find(query).sort('date', pymongo.DESCENDING).limit(1)
+        else:
+            # Keep this log in first line to help streak indexer group results. Hashes will be populated at end.
+            logger.first_log('INFO: fs_permissions_diff: {} against <empty> '.format(os_version.full))
+
+            logger.log('INFO: No suitable previous fs permissions found')
+            return ''
```
mike-petersen-ni, line 147: Conversation with @jpautler on improving output for readers of diff output.

 - Consider having a double-diff (analogue: git's three-way merge) against a basis (version < maj.min.pat.0) and recent (version < maj.min.pat.bld) to consistently report differences, but also point out new ones.
  - Attempt to match codename if possible, but if not possible, relax that constraint (and assume there's only one alternative, or equivalently. highest version found will be in the "previous" codename)
  - Try sorting by full version instead of date

```suggestion
   os_version = OsVersion()
    query = {
        'os_version_codename': os_version.codename,
        'os_version': {
            '$lt': {
                'major': os_version.major,
                'minor': os_version.minor,
                'patch': os_version.patch,
                'build': 0
            }
        }
    }

    if db.count_documents(query):
        # TODO sort change
        results = db.find(query).sort('date', pymongo.DESCENDING).limit(1)
    else:
        # If there are no results, there may not be a run from the current os codename.
        # Remove that requirement.
        del query['os_version_codename']
        logger.log('INFO: No prior fs permissions from this codename found. Relaxing that constraint.')

        if db.count_documents(query):
            # TODO sort change
            results = db.find(query).sort('date', pymongo.DESCENDING).limit(1)
        else:
            # Keep this log in first line to help streak indexer group results. Hashes will be populated at end.
            logger.first_log('INFO: fs_permissions_diff: {} against <empty> '.format(os_version.full))

            logger.log('INFO: No suitable previous fs permissions found')
            return ''
```

Also make sure that `os_version.phase` isn't uploaded to MongoDB. (Do this before enabling event so that $lt doesn't have undesirable behavior with key count stuff.)

mike-petersen-ni: Possible printout representation: similar to git's 3-way merge with two columns of +/-/space

PR #535, opened 2023-02-22
opkg-keyrings: Add "nilrt-feed-2023.gpg" pubkey

mike-petersen-ni:
This is the regular "two year" cadence update for the signing key (which was skipped last time).

This should be cherry-picked into hardknott and sumo.

# Testing
 - [x] booted VM built by [`build.vms.sh`](https://github.com/ni/nilrt/blob/nilrt/master/kirkstone/scripts/pipelines/build.vms.sh) and ran `opkg update`


amstewart (approved):

mike-petersen-ni:
@amstewart, should this be cherry-picked into all active branches?

amstewart:
Yeah; I can do that.

amstewart:
@mike-petersen-ni There's a conflict when cherry-picking to the sumo mainline. Can you resolve that and put up a PR?

Also, we should cherry-pick this to the `nilrt-academic/master/sumo` mainline as well, and pass that by the new FRC owners. Can you make a PR for that as well?

PR #539, opened 2023-02-27
ni-base-system-image-tests: Perform a three-way diff

mike-petersen-ni:
As mentioned in a comment on #534, during a discussion with @jpautler we determined that we had some contradictory requirements for this test's output:
 - only list the changes since the latest M.m.p.b version (the previous run), to make test results easier to interpret
 - list all the changes since the latest M.m.p version, to avoid accidentally ignoring changes since the last "release" (which is not *released* per se, but rather the last triplet-version before a version bump)

We assessed whether we wanted to make a custom streak indexer, which could allow each path to be reported as its own test case. This was undesirable due to the probable large amount of noise and difficulty of correlating failures.

# Changes
Two manifests are retrieved from the database. (The query may find the same manifest twice, which is expected on occasion.)
They are referred to as the "basis" and "recent" manifests, with the basis manifest being at least as old as the recent manifest.

The manifests are diffed with set operations and filters, as before. However, each of the three manifests is diffed against the other two (so the diffing work has technically tripled, though the test still completes in seconds):

         current                      23.3.0d11
          ^  ^                          ^  ^
         /    \        e.g.            /    \
        v      v                      v      v
    basis <--> recent        23.0.0f100 <--> 23.3.0d10

Doing this allows all three diffs to be considered and included in the output, to meet the needs of the test.
These diffs are referred to in the code as "net" (basis <-> current), "seen" (basis <-> recent), and "unseen" (recent <-> current).
Only *net* changes cause test failures.

## Output format
Two diff summaries are printed. The second is a three-way diff, and the first is a two way diff with the "unseen" changes (essentially the subset where the relevant column has a change for the three way diff).
For both, a legend is printed first, and as indicated, the first columns of each line are used to indicate the type of difference for each path. Subsequent lines give detail about mode, user, and group.

As with the previous test, files without any differences are not printed, though now there must be no differences in any of the three diffs.

The ordering of the differences as they are printed is in order of whether they are affected by certain diffs. First "unseen" differences are shown, since they are new to the run and more likely to be interesting; then "net" differences are shown, since they are the overall differences that we'd care about for release checks; then "seen" differences are shown for completeness.  
Paths are sorted within these groups.

Here are some snippets of output as an example:

1. Oops, `/bin` is group-writable and someone renamed `/bin/vi` (**test failed**):
```
###### Differences since last test run: current vs recent ######
(Note that these are not necessarily responsible for test failure; see full diff below)
LEGEND:
c: current changed relative to recent
+: current has a path that recent does not
-: recent has a path that current does not
###### begin last-run diff ######
c /bin
  current mode = drwxrwxr-x, user = admin, group = administrators
  recent  mode = drwxr-xr-x, user = admin, group = administrators
- /bin/vi
  recent  mode = lrwxrwxrwx, user = admin, group = administrators
+ /bin/vi.bak
  current mode = lrwxrwxrwx, user = admin, group = administrators
###### end last-run diff ######


###### Full three-way diff: current vs basis, recent vs basis, current vs recent ######
(Note that only the first column is responsible for test failure)
LEGEND:
c: X changed relative to Y
+: X has a path that Y does not
-: Y has a path that X does not
*   X=current Y=basis  (new issues since the last run for previous MAJOR.MINOR.PATCH)
 *  X=recent  Y=basis  (issues seen in the last run)
  * X=current Y=recent (new issues since the last run, probably for this MAJOR.MINOR.PATCH)
###### begin full diff ######
c c /bin
    current mode = drwxrwxr-x, user = admin, group = administrators
    basis   mode = drwxr-xr-x, user = admin, group = administrators
    recent  mode = drwxr-xr-x, user = admin, group = administrators
- - /bin/vi
    basis   mode = lrwxrwxrwx, user = admin, group = administrators
    recent  mode = lrwxrwxrwx, user = admin, group = administrators
+ + /bin/vi.bak
    current mode = lrwxrwxrwx, user = admin, group = administrators
###### end full diff ######
```
2. That issue was fixed (**test passed**):
```
###### Differences since last test run: current vs recent ######
(Note that these are not necessarily responsible for test failure; see full diff below)
LEGEND:
c: current changed relative to recent
+: current has a path that recent does not
-: recent has a path that current does not
###### begin last-run diff ######
c /bin
  current mode = drwxr-xr-x, user = admin, group = administrators
  recent  mode = drwxrwxr-x, user = admin, group = administrators
+ /bin/vi
  current mode = lrwxrwxrwx, user = admin, group = administrators
- /bin/vi.bak
  recent  mode = lrwxrwxrwx, user = admin, group = administrators
###### end unseen diff ######


###### Full three-way diff: current vs basis, recent vs basis, current vs recent ######
(Note that only the first column is responsible for test failure)
LEGEND:
c: X changed relative to Y
+: X has a path that Y does not
-: Y has a path that X does not
*   X=current Y=basis  (new issues since the last run for previous MAJOR.MINOR.PATCH)
 *  X=recent  Y=basis  (issues seen in the last run)
  * X=current Y=recent (new issues since the last run, probably for this MAJOR.MINOR.PATCH)
###### begin full diff ######
 cc /bin
    current mode = drwxr-xr-x, user = admin, group = administrators
    basis   mode = drwxr-xr-x, user = admin, group = administrators
    recent  mode = drwxrwxr-x, user = admin, group = administrators
 -+ /bin/vi
    current mode = lrwxrwxrwx, user = admin, group = administrators
    basis   mode = lrwxrwxrwx, user = admin, group = administrators
 +- /bin/vi.bak
    recent  mode = lrwxrwxrwx, user = admin, group = administrators
###### end full diff ######
```

# Testing
 - Locating previous manifests:
   - [x] basis query does not find manifest with current M.m.p
   - [x] recent query finds manifest with current M.m.p if one exists
   - [x] recent query finds same manifest as basis query if none exist for current M.m.p
   - [x] recent query does not find manifest for identical M.m.p.b if the the test is rerun
   - [x] basis query finds manifest with another codename if none exist with current codename
   - [x] recent query finds manifest with another codename if none exist with current codename
 - Pass/fail:
   - [x] test fails when net diff is nonempty
   - [x] test passes when net diff is empty, even if other diffs are found

# Merging
This should be cherry-picked into nilrt/master/hardknott, as with #534. I'll rebase upon request.

### DIFF
--- a/recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
+++ b/recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
@@ -50,8 +50,6 @@ def run_cmd(cmd):
     return output.decode('utf-8')
 
 class OsVersion:
-    phase_number = { 'd': 0, 'a': 1, 'b': 2, 'f': 3 }
-
     def __init__(self):
         with open('/etc/os-release', 'rb') as file, mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ) as mfile:
             build_id = re.search(br'BUILD_ID=\"((\d+\.\d+).*)\"', mfile)
@@ -70,7 +68,6 @@ class OsVersion:
             'major': self.major,
             'minor': self.minor,
             'patch': self.patch,
-            'phase': self.phase_number[self.phase],
             'build': self.build
         }
 
@@ -93,12 +90,14 @@ def get_fs_manifest():
     )
     return fs_manifest
 
+def log_version_info(logger, label, codename, full_version, date, db_id):
+    logger.log(f'INFO: {label} = {codename} {full_version} from {date} with _id {db_id}')
+
 def upload_manifest(db, fs_manifest, logger):
     data = {}
     os_version = OsVersion()
     data['os_version_codename'] = os_version.codename
     data['os_version_full'] = os_version.full
-    data['os_version_major_minor'] = os_version.major_minor
     data['os_version'] = os_version.version_dict
     data['date'] = str(datetime.datetime.now())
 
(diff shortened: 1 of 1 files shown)


mike-petersen-ni:
# Patch draft-v2
Rebased to reduce commit count.

 - avoid stripping command output (messes with the hash: in-memory manifest differs from uploaded manifest by a newline)
 - rewrote `Logger`'s `first_log` into `prefix_log`
 - print codename in "finding previous manifest" logging
 - sort paths before diffing
 - move recent hash to second line (not useful for streak indexer)
 - improve human-parseability of diff output
 - additionally provide simple two-way diff limited to just "unseen" changes
 - add a note to the log to help people know which column of the diff is actually responsible for test failures

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -109,110 +105,208 @@ def upload_manifest(db, fs_manifest, logger):
...
+    def run_query(query):
+        nonlocal db
+        nonlocal logger
+
+        sort_order = [('os_version', pymongo.DESCENDING), ('date', pymongo.DESCENDING)]
+
+        if db.count_documents(query):
+            results = db.find(query).sort(sort_order).limit(1)
+        else:
+            # If there are no results, there may not be a run from the current os version.
```
jpautler: Nit, I think you mean "there **must** not be a run from the current os **codename**".

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -109,110 +105,208 @@ def upload_manifest(db, fs_manifest, logger):
...
+
+def diff_manifests(current_manifest, basis_manifest, recent_manifest, logger):
+    hash_and_prep = lambda manifest: (hashlib.md5(manifest.encode('utf-8')).hexdigest(),
+                                      prepare_manifest_for_diff(manifest))
+    current_hash, current_manifest = hash_and_prep(current_manifest)
+    basis_hash, basis_manifest = hash_and_prep(basis_manifest)
+    recent_hash, recent_manifest = hash_and_prep(recent_manifest)
+
+    # Add hashes to first line of log to allow review queue to distinguish runs,
+    logger.prefix_logs[0] += f'(basis={basis_hash}, current={current_hash})'
```
jpautler: The first line is getting long. To keep the failure mode strings in dashboard reasonable, we only take the first 100 characters of the first line of output. By my estimate, you're probably up around 90 characters not including the two actual hash values. If we think we need to change that limit, we should probably discuss.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -109,110 +105,208 @@ def upload_manifest(db, fs_manifest, logger):
...
+                                                             f"mode = {manifest[path]['mode']}",
+                                                             f"user = {manifest[path]['user']}",
+                                                             f"group = {manifest[path]['group']}"]))
+    basis_detail =   detail_string('basis  ', basis_manifest)
+    recent_detail =  detail_string('recent ', recent_manifest)
+    current_detail = detail_string('current', current_manifest)
+
+    op_markers = {'changed': 'c', 'added': '+', 'removed': '-'}
+
+    logger.log('###### Unseen two-way diff: current vs recent ######')
```
jpautler: Having read the description of this PR, I get what "unseen" means. But for someone else looking at just the ATS results it might be less clear. Could we describe this section as "Differences since last test run: current vs recent"? Does that seem clearer but still reasonably accurate?

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -109,110 +105,208 @@ def upload_manifest(db, fs_manifest, logger):
...
+    logger.log('(Note that these are not necessarily responsible for test failure; see full diff below)')
+    logger.log('LEGEND:')
+    logger.log('c: current changed relative to recent')
+    logger.log('+: current has a path that recent does not')
+    logger.log('-: recent has a path that current does not')
+    logger.log('###### begin unseen diff ######')
```
jpautler: If we change the "unseen" nomenclature, we should change it here and in the matching end comment as well.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -109,110 +105,208 @@ def upload_manifest(db, fs_manifest, logger):
...
+
+            if db.count_documents(query):
+                results = db.find(query).sort(sort_order).limit(1)
+            else:
+                logger.log('INFO: No suitable previous fs permissions found')
+                return '', '<none>'
+
+        result = next(results)
+
+        logger.log('INFO: Using previous fs permissions of OS {} {} from {} with _id {}'.format(result['os_version_codename'], result['os_version_full'], result['date'], result['_id']))
```
jpautler: Could we replace the output generated from this line with something similar to this:

basis   = 23.0.0.f147 from 1/27/2023 with _id = 432987123
recent  = 23.3.0.d102 from 2/27/2023 with _id = 432987234
current = 23.3.0.d103 from 3/1/2023 with _id = 432987235

Maybe place this output as close as possible to the start of the diffs so the reader has more context about what "basis/recent/current" mean in the diff output.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -109,110 +105,208 @@ def upload_manifest(db, fs_manifest, logger):
...
+            msg_lines += [f'{current_detail(path)}', f'{recent_detail(path)}']
+        elif 'added' == difference['unseen']:
+            msg_lines += [f'{current_detail(path)}']
+        elif 'removed' == difference['unseen']:
+            msg_lines += [f'{recent_detail(path)}']
+        logger.log('\n  '.join(msg_lines))
+
+    logger.log('###### end unseen diff ######')
+
+    logger.log('###### Full three-way diff: current vs basis, recent vs basis, current vs recent ######')
```
jpautler, line 285: Could we insert a newline or two above this to create a visible separation?

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
mike-petersen-ni: Pushed change: new first line looks like

    INFO: fs_permissions_diff: 23.3.2d10000 against 23.3.1d10000 291d6685ecf0fea7fbd4744ea8509f91

(94 characters)

jpautler (approved):
Looks good!

amstewart (changes requested):
Looks good to me.

Could you squash the PR commits back into the main commits prior to merge?

mike-petersen-ni:
# Patch v2

Squashed to prepare for merge.

amstewart (approved):

amstewart:

PR #540, opened 2023-02-28
Enable initramfs to apply BIOS fixes

dylan-matthew-turner:
We want to be able to apply BIOS fixes to released hardware without requiring users to modify their bootloader.

This PR includes commits for:
- recipes for creating, installing, and adding ACPI patches to the initramfs
- recipe that installs a script for starting the initramfs
- connecting to packagefeed-ni-core
- DSDT patch for 903x (patches for other models can easily be added later, and they are not affected by not having one)


amstewart:
We're abandoning this change in favor of keeping the kernel workaround; right?

dylan-matthew-turner:
Correct

PR #542, opened 2023-03-01
recipes-core/packagegroups: Add ni-base-system-image-tests, remove packagegroup-ni-ptest

mike-petersen-ni:
We don't want to add the ni-base-system-image-tests package to one of the ptest packagegroups (which would alter the test results since it inspects the filesystem), but it is a package that should be getting built.

Codesearch & teammates suggest that packagegroup-ni-ptest is unused, and ought to be deleted. Folding that into this PR.

This should be cherry-picked into nilrt/master/hardknott, as with #534.


amstewart (approved):
Looks good; thanks.

jpautler (approved):

amstewart:

PR #543, opened 2023-03-02
kernel-tests: add test_kernel_serial_devices to run-ptest

bstreiff:
This test was added to the package, but never added to run-ptest, so it wasn't being run.

Fixes: f08425c203aa ("kernel-tests: add a test to ensure we haven't renumbered serial ports")


jpautler (approved):

amstewart (approved):

amstewart:
Cherry picked to `nilrt/master/kirkstone`.

PR #544, opened 2023-03-03
glibc-tests, test_floating_point: Fix CPU affinity and avoid CPU 0 if possible

gratian:
Based on code comments, the current implementation attempts to run the test with no CPU affinity followed by running on CPU 0 and CPU 1. However the code never calls sched_setaffinity() so the benchmark ends up running three times in a row without CPU affinity being set.

Additionally because this test runs as SCHED_FIFO with priority 50 it can create (disk) irq timeouts when it ends up monopolizing CPU 0.

Fix the CPU affinity setting and only run the test once on CPU 1 if on an SMP system. Alternatively if only one CPU is available run at reduced RT priority SCHED_FIFO/1 to avoid starving important interrupt threads which
by default on NILRT are affinitized to CPU 0.

Also remove dead code and do a bit of clean-up while we're at it.

### Testing

-  built and tested on cRIO-905x; no irq timeouts observed after 100 test runs whereas it was causing irq timeouts in less then 10 iterations before.

### DIFF
--- a/recipes-core/glibc-tests/glibc-tests/test_floating_point.cpp
+++ b/recipes-core/glibc-tests/glibc-tests/test_floating_point.cpp
@@ -20,8 +20,6 @@
 #include <sys/utsname.h>
 #include <sys/mman.h>
 
-#define gettid() syscall(__NR_gettid)
-
 #define USEC_PER_SEC		1000000
 #define NSEC_PER_SEC		1000000000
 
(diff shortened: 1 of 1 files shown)


gratian:
Changes since V1:

- rebased on latest nilrt/master/hardknott

amstewart (approved):

amstewart:
Cherry-picked to kirkstone.

PR #545, opened 2023-03-06
rt-tests: Remove sub-test test_throttling_jitter

gratian:
The RT throttling ptest contains two sub-tests:

  - first the sysctl setting 'kernel.sched_rt_runtime_us' is checked for the expected -1 value which signifies that RT throttling is disabled in the kernel.

  - the second sub-test attempts to measure jitter as an indirect indication that RT throttling is disabled.

However there are several issues with the second sub-test as indicated by the comments in 'test_throttling_jitter.c':

  - the test runs at the maximum RT priority available and pinned to CPU0. This will result in the target becoming unresponsive and result in various important kernel workloads timing out. This is especially true for CPU 0 which on NILRT services irq threads. It can result in subsequent tests failing, the root file system being re-mounted read-only, confusing error messages in the kernel log etc.

  - The threshold values used to detect a failure are arbitrary, vary between hardware variants, and have not been updated in a long while.

  - Due to the short duration of this test and lack of representative background workload this test is not a representative measure of real-time performance.

Additionally, since this test was introduced we have added better coverage for long duration real-time kernel performance tests (i.e. meta-nilrt://recipes-kernel/kernel-tests/kernel-performance-tests.bb). These tests are expected to catch any RT throttling issues.

Remove the problematic and redundant sub-test. Keep the sysctl setting check which has independent value.

### Testing

- built rt-tests package(s).
- verified by installing and running the rt-tests ptest on a cRIO-905x that only the sysctl portion of the throttling test runs and there are no errors observed.


gratian:
Changes since V1:

- rebased on latest origin/nilrt/master/hardknott

amstewart (changes requested):
Do you also need to drop this test from the [`run-ptest`](https://github.com/ni/meta-nilrt/blob/7eae40c8c864e35dc811fa44e98736473ef57bb0/recipes-rt/rt-tests/files/run-ptest#L4) script?

amstewart:
Oh wait, you're only removing the subtest. Nevermind.

amstewart (approved):

amstewart:
Cherry-picked to kirkstone.

PR #546, opened 2023-03-06
kernel-tests: Refactor hrtimers ptest

gratian:
The high-resolution timers kernel ptest is attempting to validate that the kernel has high resolution timers enabled by repeatedly calling clock_gettime() in a loop at maximum RT SCHED_FIFO priority while pinned to CPU0. The test computes a jitter value and uses this value to guess if the high-resolution timer support is enabled in the kernel based on an arbitrary jitter threshold.

This test is susceptible to fail sporadically if CPU0, which on NILRT services interrupts, gets a couple of interrupts in the middle of the test loop. Additionally because this test spins on CPU0 at high priority it can cause timeouts in important kernel threads and cause problems for subsequent tests.

Refactor the test to directly check CONFIG_HIGH_RES_TIMERS is enabled in the kernel configuration and check the correct clock source is configured (TSC for x86_64 machines).

### Testing

- bitbake kernel-tests
- cRIO-9056: opkg install kernel-tests-ptest
- ptest-runner kernel-tests  (refactored test passes)


gratian:
Changes since V1:

- rebased on latest origin/nilrt/master/hardknott

amstewart (approved):

gratian:
Changes since V2:

- rebased on latest origin/nilrt/master/hardknott

recipes-kernel/kernel-tests/kernel-tests-files/test_kernel_hrtimers.sh
```diff
@@ -1,11 +1,30 @@
...
+    ptest_pass
+else
+    ptest_fail
 fi
+ptest_report
 
-echo "PASS: kernel hrtimers clock latency"
+# test the current clock source configuration
+machine=$(uname -m)
+ptest_change_subtest 2 "test clock source configuration for $machine"
```
jpautler, line 17: I think it's a bit out of the ordinary to make ptest test case names change dynamically. Our test results already display the target-under-test prominently along side test case failures, so I don't think it's necessary to have the target name here. If you really want the target name in the test output, I'd be inclined to include it outside the pass/fail line and instead use a separate info line, like "INFO: TSC clock source not used on target [name]".

gratian: @jpautler the $machine variable here is not the target name - it is the machine type returned by uname. For all x64 targets that variable will expand to 'x86_64'.
I can take it out if you consider it unnecessary - I guess you can indirectly work out the machine type from the target name. Really the reason is there is more for future proofing if we add support for arm64 or something like that.

jpautler: Ah, I see. Then I withdraw my comment. I think it's fine as-is.

jpautler (approved):

bstreiff (approved):

amstewart:
Cherry-picked to kirkstone.

PR #547, opened 2023-03-07
packagegroup-ni-core: Add packagegroup-ni-ptest-smoke

mike-petersen-ni:
Removing the overall ptest package group removed the transitive dependency from the core packagegroup to the ptest-smoke packagegroup, preventing it from building and getting published to our feeds.

Fixes: b8a8d05a0074

This should be cherry-picked to nilrt/master/hardknott, as with #542 (which caused the issue).


amstewart (approved):

amstewart:
Cherry-picked to hardknott.

PR #550, opened 2023-03-15
[kirkstone] ni-base-system-image-tests: Add skyline cache to ignore list.

SparkingSpork:
niskyline places a file with a UUID name at a location that was not filtered out from the list of directories checked during file permissions tests. Since the UUID is different each run, the name change causes a test failure.

This change adds the Cache directory containing the problematic file to the ignore list.

[AB#2333297](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2333297)

## Testing:
Built the changes into an ipk locally and installed on a PXIe-8880. After running the test, looked at the output and confirmed the flagged difference was that the files in the Cache directory were missing (as opposed to new, different ones being present in the diff)


mike-petersen-ni (approved):
Should be cherry-picked to `nilrt/master/hardknott`, as with #534 which created the test.

amstewart (approved):

amstewart:

PR #554, opened 2023-03-23
Staple 23.3 hardknott recipe

zraul123:


zraul123 (commented):
@amstewart Your last stapling (for 23.0) didn't include a commit for SRCREV, so I'm not sure what changed. Do we not need SRCREV to be set anymore?

mike-petersen-ni:
@zraul123

> Do we not need SRCREV to be set anymore?

The RTOS team has discussed this, and we have decided to leave them as AUTOREV: the branches are already versioned; pinning to a commit is unnecessary.

amstewart (approved):
@mike-petersen-ni is correct. Setting the branch ref is sufficient for the release.

PR #555, opened 2023-03-23
Staple salt recipe in sumo release

zraul123:


zraul123:
@amstewart Same question as in https://github.com/ni/meta-nilrt/pull/554

amstewart (approved):

PR #557, opened 2023-03-23
Test known file permissions directly instead of diffing everything in the base system image

dylan-matthew-turner:
## Reason:

There are many files in the base system image that have known permissions.

When we do a diff test on the file permissions, we have to store those permissions and retrieve them from a data base and do a diff, when really, we could do a static test for many files that would be faster, have cleaner output, and require less maintenance.

## Implementation:

- Add new test (based on existing test) that checks a file permission manifest created with `find` against a csv file containing expected permissions
  + The test fails when permissions do not match expected
  + The test reports ALL differences, including if an expected path doesn't exist
- Create list of paths with known permissions in a dictionary that resembles the file tree
  + The value at each path is either a dictionary itself or a function taking the path and the stat result of the file it matches and handles checking the permissions
  + There are many built-in permissions functions like checking for system files and system directories, so the maintenance burden is lesser
  + Paths can resolve bash expressions like `$(...)` as well as globs. All paths that match will be checked against the permission
  + Paths are overridden by deeper expressions in the tree to allow exceptions. If you provide `*` to check all files but provide an exception for `some-other-file` later, the expected permission will be updated.
- Run new test along with the previous test
- Update list of installed files in the recipe

## Testing

I built and deployed a run mode in a virtual machine and tested the new known-permissions part of the test under the following conditions:
- Normal (passing) circumstances
- With a file that would be globbed deep in the tree, `/lib/modules/$(uname -r)/build/scripts/atomic/check-atomic.sh`, having wrong permissions for mode, user, group to verify the errors trigger
- Putting a directory where a file should be and vice-versa to verify it errors correctly
- Putting a link where a file should be and vice-versa to verify it errors correctly
- Changing a link to the wrong value to verify it errors correctly
- Combining user and group to be wrong to check they both output errors

### DIFF
--- a/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
+++ b/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
@@ -45,13 +45,6 @@ def known_permissions_tree():
                     '.': system_dir,
                     '**': system_hier,
                     'build': {
-                        'arch': {
-                            'x86': {
-                                'tools': {
-                                    'relocs': system_file_exec
-                                }
-                            }
-                        },
                         'include': {
                             'dt-bindings': {
                                 'clock': {
@@ -64,36 +57,7 @@ def known_permissions_tree():
                             }
                         },
                         'scripts': {
-                            '**/*.sh': system_file_exec,
-                            '**/*.pl': system_file_exec,
-                            'asn1_compiler': system_file_exec,
-                            'atomic': {
-                                'atomics.tbl': system_file_exec,
-                                'fallbacks': {
-                                    '*': system_file_exec
-                                }
-                            },
-                            'basic': {
-                                'fixdep': system_file_exec
-                            },
-                            'bloat-o-meter': system_file_exec,
-                            'bpf_doc.py': system_file_exec,
-                            'check-sysctl-docs': system_file_exec,
-                            'checkkconfigsymbols.py': system_file_exec,
-                            'clang-tools': {
-                                'gen_compile_commands.py': system_file_exec,
-                                'run-clang-tools.py': system_file_exec
-                            },
-                            'cleanfile': system_file_exec,
-                            'cleanpatch': system_file_exec,
-                            'coccicheck': system_file_exec,
-                            'config': system_file_exec,
-                            'decodecode': system_file_exec,
-                            'diffconfig': system_file_exec,
-                            'documentation-file-ref-check': system_file_exec,
                             'dtc': {
-                                'dt_to_config': system_file_exec,
-                                'dtx_diff': system_file_exec,
                                 'include-prefixes': {
                                     'arc': system_link('../../../arch/arc/boot/dts'),
                                     'arm': system_link('../../../arch/arm/boot/dts'),
@@ -110,77 +74,10 @@ def known_permissions_tree():
                                 }
                             },
                             'dummy-tools': {
-                                'gcc': system_file_exec,
-                                'ld': system_file_exec,
                                 'nm': system_link('ld'),
                                 'objcopy': system_link('ld')
                             },
-                            'extract-cert': system_file_exec,
-                            'extract-ikconfig': system_file_exec,
-                            'extract-vmlinux': system_file_exec,
-                            'faddr2line': system_file_exec,
-                            'gcc-ld': system_file_exec,
-                            'genksyms': {
-                                'genksyms': system_file_exec
-                            },
-                            'get_dvb_firmware': system_file_exec,
-                            'gfp-translate': system_file_exec,
-                            'jobserver-exec': system_file_exec,
-                            'kallsyms': system_file_exec,
-                            'kconfig': {
-                                'conf': system_file_exec
-                            },
-                            'kernel-doc': system_file_exec,
-                            'Lindent': system_file_exec,
-                            'makelst': system_file_exec,
-                            'mkcompile_h': system_file_exec,
-                            'mksysmap': system_file_exec,
-                            'mod': {
-                                'mk_elfconfig': system_file_exec,
-                                'modpost': system_file_exec
-                            },
-                            'objdiff': system_file_exec,
-                            'package': {
-                                '*': system_file_exec,
-                                'snapcraft.template': system_file
-                            },
-                            'patch-kernel': system_file_exec,
-                            'prune-kernel': system_file_exec,
-                            'remove-stale-files': system_file_exec,
-                            'selinux': {
-                                'genheaders': {
-                                    'genheaders': system_file_exec,
-                                },
-                                'mdp': {
-                                    'mdp': system_file_exec
-                                }
-                            },
-                            'setlocalversion': system_file_exec,
-                            'show_delta': system_file_exec,
-                            'sorttable': system_file_exec,
-                            'spdxcheck.py': system_file_exec,
-                            'spdxcheck-test.sh': system_file,
-                            'sphinx-pre-install': system_file_exec,
-                            'stackdelta': system_file_exec,
-                            'stackusage': system_file_exec,
-                            'tracing': {
-                                'draw_functrace.py': system_file_exec
-                            },
-                            'ver_linux': system_file_exec,
                         },
-                        'tools': {
-                            'lib': {
-                                'lockdep': {
-                                    'lockdep': system_file_exec,
-                                    'run_tests.sh': system_file_exec
-                                }
-                            },
-                            'objtool': {
-                                'fixdep': system_file_exec,
-                                'objtool': system_file_exec,
-                                'sync-check.sh': system_file_exec
-                            }
-                        }
                     },
                     'source': system_link('build')
                 }
@@ -270,7 +167,7 @@ def system_dir(path, stats, logger, md5sum):
 # Either system file or system dir
 def system_hier(path, stats, logger, md5sum):
     is_dir = stat.S_ISDIR(stats.st_mode)
-    mode = 0o0755 if is_dir else 0o0644
+    mode = 0o0755 if is_dir or stat.S_IMODE(stats.st_mode) == 0o0755 else 0o0644
     file_type = FT_DIR if is_dir else FT_REG
     return permissions(mode, 'admin', 'administrators', file_type)(path, stats, logger, md5sum)
 
--- a/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
+++ b/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
@@ -227,31 +227,32 @@ def log_mismatch(kind, exp, act, path, logger, md5sum):
 def permissions(mode, user, group, file_type):
     def ret(path, stats, logger, md5sum):
         nonlocal mode, user, group, file_type
+        success = True
         if file_type == FT_DIR and not stat.S_ISDIR(stats.st_mode):
             log_mismatch(
                 'file type', 'dir', 'reg' if file_type == FT_REG else 'lnk', path, logger, md5sum
             )
-            return False
+            success = False
         if file_type == FT_REG and not stat.S_ISREG(stats.st_mode):
             log_mismatch(
                 'file type', 'reg', 'dir' if file_type == FT_DIR else 'lnk', path, logger, md5sum
             )
-            return False
+            success = False
         if file_type == FT_LNK and not stat.S_ISLNK(stats.st_mode):
             log_mismatch(
                 'file type', 'lnk', 'dir' if file_type == FT_DIR else 'reg', path, logger, md5sum
             )
-            return False
+            success = False
         if mode != stat.S_IMODE(stats.st_mode):
             log_mismatch('mode', oct(mode), oct(stat.S_IMODE(stats.st_mode)), path, logger, md5sum)
-            return False
+            success = False
         if user != pwd.getpwuid(stats.st_uid).pw_name:
             log_mismatch('user', user, pwd.getpwuid(stats.st_uid).pw_name, path, logger, md5sum)
-            return False
+            success = False
         if group != grp.getgrgid(stats.st_gid).gr_name:
             log_mismatch('group', group, grp.getgrgid(stats.st_gid).gr_name, path, logger, md5sum)
-            return False
-        return True
+            success = False
+        return success
     return ret
 
 # Typical system file
--- a/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
+++ b/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
@@ -43,43 +43,7 @@ def known_permissions_tree():
                 '.': system_dir,
                 '$(uname -r)': {
                     '.': system_dir,
-                    '**': system_hier,
-                    'build': {
-                        'include': {
-                            'dt-bindings': {
-                                'clock': {
-                                    'qcom,dispcc-sm8150.h': system_link('qcom,dispcc-sm8250.h')
-                                },
-                                'input': {
-                                    'linux-event-codes.h':
-                                        system_link('../../uapi/linux/input-event-codes.h')
-                                }
-                            }
-                        },
-                        'scripts': {
-                            'dtc': {
-                                'include-prefixes': {
-                                    'arc': system_link('../../../arch/arc/boot/dts'),
-                                    'arm': system_link('../../../arch/arm/boot/dts'),
-                                    'arm64': system_link('../../../arch/arm64/boot/dts'),
-                                    'dt-bindings': system_link('../../../include/dt-bindings'),
-                                    'h8300': system_link('../../../arch/h8300/boot/dts'),
-                                    'microblaze': system_link('../../../arch/microblaze/boot/dts'),
-                                    'mips': system_link('../../../arch/mips/boot/dts'),
-                                    'nios2': system_link('../../../arch/nios2/boot/dts'),
-                                    'openrisc': system_link('../../../arch/openrisc/boot/dts'),
-                                    'powerpc': system_link('../../../arch/powerpc/boot/dts'),
-                                    'sh': system_link('../../../arch/sh/boot/dts'),
-                                    'xtensa': system_link('../../../arch/xtensa/boot/dts'),
-                                }
-                            },
-                            'dummy-tools': {
-                                'nm': system_link('ld'),
-                                'objcopy': system_link('ld')
-                            },
-                        },
-                    },
-                    'source': system_link('build')
+                    '**': system_hier
                 }
             }
         }
@@ -164,11 +128,14 @@ def system_file_exec(path, stats, logger, md5sum):
 def system_dir(path, stats, logger, md5sum):
     return permissions(0o0755, 'admin', 'administrators', FT_DIR)(path, stats, logger, md5sum)
 
-# Either system file or system dir
+# Only care about ownership and r/w access not link status, directory status, exec bit, etc
 def system_hier(path, stats, logger, md5sum):
     is_dir = stat.S_ISDIR(stats.st_mode)
-    mode = 0o0755 if is_dir or stat.S_IMODE(stats.st_mode) == 0o0755 else 0o0644
-    file_type = FT_DIR if is_dir else FT_REG
+    is_link = stat.S_ISLNK(stats.st_mode)
+    mode = 0o0777 if is_link else (
+        0o0755 if is_dir or stat.S_IMODE(stats.st_mode) == 0o0755 else 0o0644
+    )
+    file_type = FT_DIR if is_dir else (FT_LNK if is_link else FT_REG)
     return permissions(mode, 'admin', 'administrators', file_type)(path, stats, logger, md5sum)
 
 # What would be a typical system file but is actually a link
--- a/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
+++ b/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
@@ -1,0 +1,162 @@
+import argparse
+import os
+import sys
+import hashlib
+import stat
+import pwd
+import grp
+import fs_permissions_shared
+
+# Function to return tree instead of variable since it contains functions inside its structure
+def known_permissions_tree():
+    return {
+        '.': system_dir,
+        'boot': {
+            '.': system_dir,
+            'bootmode': system_file,
+            'grub': {
+                '.': system_dir,
+                '*': system_file,
+                'grubenv': permissions(0o664, 'admin', 'ni', FT_REG),
+                'grubenv.bak': permissions(0o664, 'admin', 'ni', FT_REG),
+                'recoverytool-ni-version': permissions(0o0444, 'admin', 'administrators', FT_REG)
+            }, 'runmode': {
+                '.': system_dir,
+                '*': system_file,
+                'bzImage': system_link(f'bzImage-{bash("$(uname -r)")[0]}'),
+            }
+        }, 'lib': {
+            '.': system_dir,
+            'modules': {
+                '.': system_dir,
+                '$(uname -r)': {
+                    '.': system_dir,
+                    '**': system_hier,
+                    'source': system_link('build')
+                }
+            }
+        }
+    }
+
+# Use bash to resolve strings like path globs and $() expressions.
+def bash(path):
+    cmd = [ 'bash', '-O', 'extglob', '-c', f'echo {path}' ]
+    return fs_permissions_shared.run_cmd(cmd).strip().split(' ')
+
+# Convert the tree structure into a list of paths and functions to check their permissions
+def known_permissions():
+    def permissions_list(tree, path, perms):
+        for file in tree:
+            if type(tree[file]) is dict:
+                perms = permissions_list(tree[file], f'{path}/{file}', perms)
+            elif file == '.':
+                perms[bash(path if path != '' else '/')[0]] = tree[file]
+            else:
+                subfiles = bash(f'{path}/{file}')
+                for subfile in subfiles:
+                    perms[subfile] = tree[file]
+        return perms
+    return permissions_list(known_permissions_tree(), '', {})
+
+#------- Functions for the tree to check file permissions ------
+
+FT_DIR = 0
+FT_REG = 1
+FT_LNK = 2
+
+# Report errors when a kind of thing is different than expected
+def log_mismatch(kind, exp, act, path, logger, md5sum):
+    md5sum.update(f'{kind}: {exp} != {act}'.encode('utf-8'))
+    logger.log(f'ERROR: Expected {kind} \'{exp}\' but got \'{act}\' for path {path}.')
+
+# Main check for file/directory permissions
+def permissions(mode, user, group, file_type):
+    def ret(path, stats, logger, md5sum):
+        nonlocal mode
+        nonlocal user
+        nonlocal group
+        nonlocal file_type
+        if file_type == FT_DIR and not stat.S_ISDIR(stats.st_mode):
+            log_mismatch(
+                'file type', 'dir', 'reg' if file_type == FT_REG else 'lnk', path, logger, md5sum
+            )
+            return False
+        if file_type == FT_REG and not stat.S_ISREG(stats.st_mode):
+            log_mismatch(
+                'file type', 'reg', 'dir' if file_type == FT_DIR else 'lnk', path, logger, md5sum
+            )
+            return False
+        if file_type == FT_LNK and not stat.S_ISLNK(stats.st_mode):
+            log_mismatch(
+                'file type', 'lnk', 'dir' if file_type == FT_DIR else 'reg', path, logger, md5sum
+            )
+            return False
+        if mode != stat.S_IMODE(stats.st_mode):
+            log_mismatch('mode', oct(mode), oct(stat.S_IMODE(stats.st_mode)), path, logger, md5sum)
+            return False
+        if user != pwd.getpwuid(stats.st_uid).pw_name:
+            log_mismatch('user', user, pwd.getpwuid(stats.st_uid).pw_name, path, logger, md5sum)
+            return False
+        if group != grp.getgrgid(stats.st_gid).gr_name:
+            log_mismatch('group', group, grp.getgrgid(stats.st_gid).gr_name, path, logger, md5sum)
+            return False
+        return True
+    return ret
+
+# Typical system file
+def system_file(path, stats, logger, md5sum):
+    return permissions(0o0644, 'admin', 'administrators', FT_REG)(path, stats, logger, md5sum)
+
+# Typical system dir
+def system_dir(path, stats, logger, md5sum):
+    return permissions(0o0755, 'admin', 'administrators', FT_DIR)(path, stats, logger, md5sum)
+
+# Either system file or system dir
+def system_hier(path, stats, logger, md5sum):
+    is_dir = stat.S_ISDIR(stats.st_mode)
+    mode = 0o0755 if is_dir else 0o0644
+    file_type = FT_DIR if is_dir else FT_REG
+    return permissions(mode, 'admin', 'administrators', file_type)(path, stats, logger, md5sum)
+
+# What would be a typical system file but is actually a link
+def system_link(target):
+    def ret(path, stats, logger, md5sum):
+        nonlocal target
+        real_target = os.readlink(path)
+        correct_link = target == real_target
+        if not correct_link:
+            log_mismatch('link', target, real_target, path, logger, md5sum)
+        return \
+            permissions(0o0777, 'admin', 'administrators', FT_LNK)(path, stats, logger, md5sum) \
+            and correct_link
+    return ret
+
+# ------ End tree functions ------
+
+def check_known_permissions(logger):
+    logger.log('###### Checking expected file permissions against actual for known folders')
+    md5sum = hashlib.md5()
+    success = True
+    perms = known_permissions()
+    for path in perms:
+        stats = os.lstat(path)
+        if not perms[path](path, stats, logger, md5sum):
+            success = False
+    logger.log('###### Finished checking file permissions')
+    return success
+
+def main():
+    argparse.ArgumentParser(description = 'Check fs permissions against expectations').parse_args()
+    logger = fs_permissions_shared.Logger()
+    
+    result = check_known_permissions(logger)
+
+    logger.report()
+    if result:
+        sys.exit(os.EX_OK)
+    else:
+        sys.exit(os.EX_SOFTWARE)
+
+if __name__ == '__main__':
+    main()
+
--- a/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
+++ b/recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
@@ -27,6 +27,16 @@ def known_permissions_tree():
                 'bzImage': system_link(f'bzImage-{bash("$(uname -r)")[0]}'),
             }
         },
+        'etc': {
+            'fstab': permissions(0o0644, 'lvuser', 'ni', FT_REG)
+        },
+        'home': {
+            '.': system_dir,
+            '*': system_dir,
+            'lvuser': permissions(0o2775, 'lvuser', 'ni', FT_DIR),
+            'root': system_link('/home/admin'),
+            'webserv': permissions(0o2755, 'webserv', 'ni', FT_DIR)
+        },
         'lib': {
             '.': system_dir,
             'modules': {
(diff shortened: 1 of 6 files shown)


jpautler:
The location that motivated this request was /lib/modules/$KERNEL_VERSION/*. Can you add that? Since that will exercise the functionality to include the output of shell commands in the path names, it would be good to include it and test it now in the original PR. Thanks!

dylan-matthew-turner:
> The location that motivated this request was /lib/modules/$KERNEL_VERSION/*. Can you add that?

I can, and I was going to use that as the example originally (and I have tested with it), but I noticed that we don't currently check /lib/modules at all. Do you want me to remove /lib/modules from the array used to omit paths from the `find` command?

Omit array currently: `omit_dirs = ['/etc/natinst/niskyline/Data/Assets/Cache', '/lib/modules', '/var/cache', '/var/run', '/var/tmp', '/var/volatile']`

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,153 @@
...
+
+def run_cmd(cmd):
+    output = subprocess.check_output(cmd)
+    return output.decode('utf-8')
+
+# Columns: Path, Mode, User, Group, Link Target
+fs_manifest_format = '%p\t%M\t%u\t%g\t%l'
+def get_fs_manifest():
+    search_dirs = ['/bin', '/boot', '/etc', '/lib', '/lib64', '/sbin', '/usr', '/var']
+    omit_dirs = ['/etc/natinst/niskyline/Data/Assets/Cache', '/lib/modules', '/var/cache', '/var/run', '/var/tmp', '/var/volatile']
```
mike-petersen-ni: Is it necessary to enumerate the paths here (and run `find`) if we are checking only the files in the CSV?

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,153 @@
...
+            'mode': entry_data[1],
+            'user': entry_data[2],
+            'group': entry_data[3],
+            'link': entry_data[4]
+        }
+    return entries
+
+def log_version_info(logger, label, codename, full_version, date, db_id):
+    logger.log(f'INFO: {label} = {codename} {full_version} from {date} with _id {db_id}')
+
```
mike-petersen-ni: dead function `log_version_info`

```suggestion
```

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,153 @@
...
+    failed = False
+    with open('known-permissions.csv') as known_perm_file:
+        permissions = csv.reader(known_perm_file, delimiter = ',')
+        header_row = True
+        for row in permissions:
+            if header_row:
+                header_row = False
+                continue
+
+            # Allow executing shell script to get path name.
```
mike-petersen-ni: Parsing this ourselves is a bit smelly and not very performant. Regexes would have problems with nesting, but could we find a Python library that permits nesting, or assume/enforce that there is no nesting so that we could use something like [`re.finditer`](https://docs.python.org/3/library/re.html?highlight=re#re.finditer)? Perhaps even something so crude as actually invoking the shell like `run_cmd(['bash', '-c', f'echo {path}'])` could work (with proper escaping), and also handle wildcards.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,153 @@
...
+                    i += 1
+                    shell_cmd += path[i]
+                elif path[i] == ')':
+                    shell_out = run_cmd(shell_cmd.split(' ')).strip()
+                    new_path += shell_out
+                    building_shell = False
+                else:
+                    shell_cmd += path[i]
+                i += 1
+            path = new_path
```
mike-petersen-ni: String accumulation has worse than quadratic behavior due to immutability; I'd suggest appending to a list and joining at the end.

```suggestion
            path = "".join(new_path_fragments)
```

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,153 @@
...
+    return entries
+
+def log_version_info(logger, label, codename, full_version, date, db_id):
+    logger.log(f'INFO: {label} = {codename} {full_version} from {date} with _id {db_id}')
+
+def check_known_manifests(manifest, logger):
+    logger.log('###### Checking expected file permissions against actual for known folders')
+    failed = False
+    with open('known-permissions.csv') as known_perm_file:
+        permissions = csv.reader(known_perm_file, delimiter = ',')
```
mike-petersen-ni: I'd suggest using a [`DictReader`](https://docs.python.org/3/library/csv.html#csv.DictReader) to make it more readable later, rather than using magic numbers as indices into the rows.

```suggestion
        permissions = csv.DictReader(known_perm_file, delimiter = ',', fieldnames = ['path', 'mode', 'user', 'group', 'link'])
```

jpautler:
>Do you want me to remove /lib/modules from the array used to omit paths from the `find` command?
> 
Yes. As I recall, it was the fact we were omitting it that originally motivated the known-permissions test. I think we omitted it originally because the names of the folders/files/links would all change each time the kernel bumped, which would make the diff test noisy.

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
```diff
@@ -0,0 +1,2 @@
+Path,Mode,User,Group,Link Target
+/etc/fstab,-rw-r--r--,admin,administrators,
```
mike-petersen-ni: Some other possible ideas (don't feed obligated to do all these):

 - everything in `/boot`, `/bin`, `/lib`, `/lib64`, `/sbin`, and `/usr` should be writable only by superuser
 - versioned kernels in `/boot`
 - home directories (don't bother with traversing their contents) should be mode 0700 (I think) and owned by their user
 - `/tmp` (or its link target) should be mode 1777

Might make more sense to do some of these if we had general predicates rather than explicit listings—it'd be annoying to maintain a CSV like this for some of these.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: Well, we have to check the permissions somehow, idk if there's a better way to do that than with `find`, but we could call `find` only on the files we want instead of batching the find and looping through them to organize it into a dict.

I'll put the logic to check a path's permissions into a method, so if it's not performant, we can adjust it easily:

```
class Permissions:
    def __init__(self, file_name):
        raw_entry = get_permissions(file_name)
        entry = raw_entry.split('\t')

        if len(entry) > 4:
            self.path = entry[0]
            self.mode = entry[1]
            self.user = entry[2]
            self.group = entry[3]
            self.link = entry[4]
        else:
            self.path = None
            self.mode = None
            self.user = None
            self.group = None
            self.link = None

    def get_permissions(file_name):
        find_format = '%p\t%M\t%u\t%g\t%l\n'
        find_cmd = [ 'find', file_name, '-printf', find_format, '-prune' ]
        return run_cmd(find_cmd).strip()

def run_cmd(cmd):
    output = subprocess.check_output(cmd)
    return output.decode('utf-8')
```

mike-petersen-ni: A bit of searching shows a couple of relevant Python functions here: [`os.stat`](https://docs.python.org/3/library/os.html#os.stat) and possibly [`pwd.getpwuid`](https://docs.python.org/3/library/pwd.html#pwd.getpwuid).  
I think that if we are going to expand the known paths ahead of time, and give just a single path to this class, we might as well do that stat call ourselves rather than parsing command output.

dylan-matthew-turner: Okay, here's a python way to do it:

```
class Permissions:
    def __init__(self, file_name):
        if !os.path.exists(file_name):
            self.path = None
            self.mode = None
            self.user = None
            self.group = None
            self.link = None
        stats = os.stat(file_name)
        self.path = file_name
        self.mode = stat.filemode(stats.st_mode)
        self.user = pwd.getpwuid(stats.st_uid)
        self.group = pwd.getpwgid(stats.st_gid)
        if stat.S_ISLINK(self.mode):
            self.link = os.readlink(self.path)
        else:
            self.link = ''
```

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: Here's a solution:

```
def run_cmd(cmd):
    output = subprocess.check_output(cmd)
    return output.decode('utf-8')

def eval_path_expr(path):
    first_pass_cmd = [ 'bash', '-c', f'echo {path}' ]
    first_pass = run_cmd(first_pass_cmd).strip().split('\n')
    return glob.glob(first_pass)
```

The `bash -c` wasn't resolving wildcards, but the built-in glob library can do that for us

dylan-matthew-turner: Hmm. Nevermind. It does seem to be working without the glob:

```
def eval_path_expr(path):
    cmd = [ 'bash', '-c', f'echo {path}' ]
    return run_cmd(cmd).strip().split('\n')
```

dylan-matthew-turner:
I am redoing the commits, so the log stays clean, so these are the changes from the original commits I posted, for reference:
- fs_permissions_known.py:
  + Replace heavy bash `find` search with individual file lookup and single loops
  + Simplify internal shell evaluation using `bash -c 'echo ...'`
  + Support globs
  + Replace magic-number indexing with dictionary and class access
  + Split code into more functions to increase readability
  + Remove dead code
- known-permissions.csv
  + Add more files like /lib/modules/$(uname)/*, /boot, /bin, etc
- fs_permissions_diff.py:
  + update to use new evaluation code
- fs_permissions.py:
  + created to store shared code between the two parts of the test
  + Logger, run_cmd, and eval_path_expr

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
dylan-matthew-turner: I have added some of these. Feel free to look back over the csv again

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: Well, we can avoid it by doing the bash route anyway.

dylan-matthew-turner:
I have retested the changes and verified they work under the same circumstances, including the new globs

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
```diff
@@ -0,0 +1,15 @@
...
+/boot/grub/[^recoverytool-ni-version]*,-rw-rw-r--,admin,ni,
+/boot/grub/recoverytool-ni-version,-r--r--r--,admin,administrators,
+/boot/runmode/[^bzImage]*,-rw-r--r--,admin,administrators,
+/boot/runmode/bzImage,lrwxrwxrwx,admin,administrators,bzImage-$(uname -r)
+/etc/fstab,-rw-r--r--,admin,administrators,
+/home/admin,drwxr-xr-x,admin,administrators,
+/home/lvuser,drwxrwsr-x,lvuser,ni,
+/home/root,lrwxrwxrwx,admin,administrators,/home/admin
+/home/webserv,drwxr-sr-x,webserv,ni,
+"/lib/modules/$(uname -r)/[^build,kernel,source]*",-rw-r--r--,admin,administrators,
```
mike-petersen-ni: I've not seen this `[^word1,word2]` syntax before, and the closest I can find in the manual is [here](https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html), where it says that `[^abc]` operates per-character (and `!(pattern1|pattern2)` is possible with the implied semantics, but requires the extglob option to be set).

Do you have a reference that says these work in the way a human might read this? Otherwise this might be equivalent to

```suggestion
"/lib/modules/$(uname -r)/[^,bcdeiklnorsu]*",-rw-r--r--,admin,administrators,
```

and might mislead readers and accidentally exclude more than intended.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,82 @@
...
+    logger.log(f'ERROR: Expected {col} \'{exp}\' but got \'{act}\' for path {path}.')
+
+class Permissions:
+    def __init__(self, file_name):
+        if not os.path.exists(file_name):
+            self.path = None
+            self.mode = None
+            self.user = None
+            self.group = None
+            self.link = None
```
mike-petersen-ni: There's no `else` here. Does the path need to exist for what follows? If not, why the conditional?

recipes-ni/ni-base-system-image-tests/files/test_fs_permissions.sh
```diff
@@ -2,9 +2,10 @@
 
 source $(dirname "$0")/ptest-format.sh
 
-ptest_change_test $(basename "$0" ".sh") "" "Diff filesystem permissions with previous"
+ptest_change_test $(basename "$0" ".sh") "" "Check known filesystem permissions and diff others with previous"
 
 source /home/admin/.mongodb.creds
+python3 fs_permissions_known.py
```
mike-petersen-ni: Looks like we aren't using the exit code of this test, so I suspect it won't cause ptest failure. Looking at `ptest-format.sh` it looks like this should be in its own script so that it can have its own `ptest_change_test`, `ptest_pass`, `ptest_fail`, and `ptest_report` calls. Either that, or this one script should separate the two tests so each can produce an independent report.

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
dylan-matthew-turner: I don't have a reference. That's just what I got working when | didn't immediately work. I may have missed parenthesis. I'll try the way you're suggesting again

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: I'm missing a return here. Later on I check for path to be None, so at least path needs to be None bc I log an issue later using that, but I suppose we don't need to set the others

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
dylan-matthew-turner: I found this online:

> extglob: Enables extended globbing patterns, such as, for example, !(this|that) (which would match like * but not any name that is this or that).

dylan-matthew-turner: Okay I think I have it working:
- Call with `extglob`: `cmd = [ 'bash', '-O', 'extglob', '-c', f'echo {path}' ]`
- CSV:
   ```
   /lib/modules/$(uname -r)/!(build|kernel|source),-rw-r--r--,admin,administrators,
   /lib/modules/$(uname -r)/@(build|kernel),drwxr-xr-x,admin,administrators,
   ```

recipes-ni/ni-base-system-image-tests/files/test_fs_permissions.sh
dylan-matthew-turner: We can OR the results:

```
python3 test_a ...; A_RES=$?
python3 test_b ...; B_RES=$?

if [ $(($A_RES | $B_RES)) -eq 0 ]; then
    ptest_pass
else
    ptest_fail
fi

ptest_report
```

mike-petersen-ni: Probably more readable/idiomatic to use the [builtin's boolean AND](https://www.gnu.org/software/bash/manual/html_node/Bourne-Shell-Builtins.html#index-test): `[ $A_RES -eq 0 -a $B_RES -eq 0 ]`.

mike-petersen-ni: @jpautler, is there any preference with regard to these tests being combined like this, where they cannot fail independently?

jpautler: I'm inclined to say we should separate them. Since the test logic and reporting format are different, it makes sense to report the results separately. Also, as I've seen with the dmesg_diff test, there is the potential for diff-based tests to be noisier than average. Hopefully we've through this file-permissions-diff-test well-enough that it won't be noisy, but I still think it's a good idea to separate the test reporting, just in case. In summary, I see benefits and no downsides to separating them.

dylan-matthew-turner: Would we need to make a whole new package for that, or could it be something as simple as:
- One script for diff:
   ```
   # Diff Permissions Test
   source $(dirname "$0")/ptest-format.sh
   ptest_change_test $(basename "$0" ".sh") "" "Diff filesystem permissions with previous"
   source /home/admin/.mongodb.creds
   python3 fs_permissions_diff.py \
       --server $MONGO_SERVER --user $MONGO_USER --password $MONGO_PASSWORD
   if [ $? -eq 0 ]; then
      ptest_pass
   else
      ptest_fail
   fi
   ptest_report
   ```
- One for known:
   ```
   # Known Permissions Test:
   source $(dirname "$0")/ptest-format.sh
   ptest_change_test $(basename "$0" ".sh") "" "Check known filesystem permissions"
   python3 fs_permissions_known.py
   if [ $? -eq 0 ]; then
      ptest_pass
   else
      ptest_fail
   fi
   ptest_report
   ```
- Both in run-ptest:
   ```
   ./test_fs_permissions_diff.sh
   ./test_fs_permissions_known.sh
   return 0
   ```
?

mike-petersen-ni: There's no need for a new package. This is why the `run-ptest` script exists: to invoke multiple tests that each report their results. See the [kernel-tests package](https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-kernel/kernel-tests/kernel-tests-files/run-ptest) for an example.

dylan-matthew-turner:
I am redoing the commits, so the log stays clean, so these are the changes from the last set of commits I posted, for reference:
- Add return to Permissions constructor case where file does not exist
- Split known permissions into separate test
- Fix glob support (by adding "extglob" bash option)

recipes-ni/ni-base-system-image-tests/files/fs_permissions_shared.py
```diff
@@ -0,0 +1,31 @@
...
+            print(log)
+
+def run_cmd(cmd):
+    output = subprocess.check_output(cmd)
+    return output.decode('utf-8')
+
+# Paths can be general, using $() to execute shell scripts /* to evaluate to multiple paths
+# Example, to match against all subitems of the current kernel modules use:
+#     /lib/modules/$(uname -r)/*
+# You can pass arguments into the find command via spaces
```
mike-petersen-ni: I don't see why this comment is relevant here.

```suggestion
```

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
```diff
@@ -0,0 +1,15 @@
...
+/boot/runmode/!(bzImage),-rw-r--r--,admin,administrators,
+/boot/runmode/bzImage,lrwxrwxrwx,admin,administrators,bzImage-$(uname -r)
+/etc/fstab,-rw-r--r--,admin,administrators,
+/home/admin,drwxr-xr-x,admin,administrators,
+/home/lvuser,drwxrwsr-x,lvuser,ni,
+/home/root,lrwxrwxrwx,admin,administrators,/home/admin
+/home/webserv,drwxr-sr-x,webserv,ni,
+/lib/modules/$(uname -r)/!(build|kernel|source),-rw-r--r--,admin,administrators,
+/lib/modules/$(uname -r)/@(build|kernel),drwxr-xr-x,admin,administrators,
+/lib/modules/$(uname -r)/source,lrwxrwxrwx,admin,administrators,build
```
mike-petersen-ni: I think that we should recursively check contents of these directories (possibly others too). Could we use the [globstar `**`](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html) here?

recipes-ni/ni-base-system-image-tests/ni-base-system-image-tests.bb
```diff
@@ -13,19 +13,24 @@ ALLOW_EMPTY:${PN} = "1"
...
 do_install_ptest() {
-    install -m 0644 ${WORKDIR}/ptest-format.sh        ${D}${PTEST_PATH}
-    install -m 0755 ${WORKDIR}/run-ptest              ${D}${PTEST_PATH}
-    install -m 0644 ${WORKDIR}/fs_permissions_diff.py ${D}${PTEST_PATH}
-    install -m 0755 ${WORKDIR}/test_*.sh              ${D}${PTEST_PATH}
+    install -m 0644 ${WORKDIR}/ptest-format.sh          ${D}${PTEST_PATH}
+    install -m 0644 ${WORKDIR}/known-permissions.csv    ${D}${PTEST_PATH}
+    install -m 0644 ${WORKDIR}/fs_permissions_*.py   ${D}${PTEST_PATH}
+    install -m 0755 ${WORKDIR}/test_*.sh                ${D}${PTEST_PATH}
+    install -m 0755 ${WORKDIR}/run-ptest                ${D}${PTEST_PATH}
```
mike-petersen-ni: Formatting nitpick

```suggestion
    install -m 0644 ${WORKDIR}/ptest-format.sh       ${D}${PTEST_PATH}
    install -m 0644 ${WORKDIR}/known-permissions.csv ${D}${PTEST_PATH}
    install -m 0644 ${WORKDIR}/fs_permissions_*.py   ${D}${PTEST_PATH}
    install -m 0755 ${WORKDIR}/test_*.sh             ${D}${PTEST_PATH}
    install -m 0755 ${WORKDIR}/run-ptest             ${D}${PTEST_PATH}
```

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,79 @@
...
+        self.mode = stat.filemode(stats.st_mode)
+        self.user = pwd.getpwuid(stats.st_uid).pw_name
+        self.group = grp.getgrgid(stats.st_gid).gr_name
+        if stat.S_ISLNK(stats.st_mode):
+            self.link = os.readlink(self.path)
+        else:
+            self.link = ''
+
+def check_known_permissions(logger):
+    logger.log('###### Checking expected file permissions against actual for known folders')
```
mike-petersen-ni, line 157: Looks like this is going to be the first line reported, and since it's constant the review queue won't be able to distinguish failures. Consider producing a digest and putting it into the first line with `Logger.prefix_log`.

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
mike-petersen-ni: On second thought, globstar might not work well our use case here, since it enumerates both files and directories. Maybe we need another mechanism for recursive checks.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_shared.py
dylan-matthew-turner: I think it's just left over from previous iterations. I'll remove it

recipes-ni/ni-base-system-image-tests/ni-base-system-image-tests.bb
dylan-matthew-turner: Good catch

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
dylan-matthew-turner: I've been thinking about this, and maybe the solution is we remove the first character of the mode string and just ignore checking if it's a file or directory altogether. Makes the csv a lot simpler

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
jpautler: To expand a little on what Mike said, we have tooling that will group together similar failures across test runs. One of the things it looks at to determine whether to group failures together is the first line of the error message. If the error message is constant, it can only put all failures in one group. If the error message is dynamic, it can group more narrowly depending on the specific differences causing the failure.

That being said, it only looks at the first line and only up to the first 100 characters. So you can't cram too much information into the first line, but you can calculate a hash or checksum that uniquely identifies the failure and include that in the first line. You can look at the fs_permissions_diff test as an example.

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
mike-petersen-ni: That's likely insufficient. Directories use the execute bit for determining whether someone can list their contents, so they'll pretty much always have at least 0100 set (e.g. most system dirs are 0755, most config files are 0644).

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: Would something like this work?

```
def add_log_prefix(logger):
    release_text = ''
    with open('/etc/os-release') as release_file:
        release_text = release_file.read()
    build_id_match = re.search(r'BUILD_ID="(\\.|[^\\"])*"', release_text)
    build_id = build_id_match.group(1)[10:-1]
    logger.prefix_log(f'INFO: fs_permissions_known: {build_id}')
    logger.log('\n')
```

mike-petersen-ni: I think it'd be preferable to have a message based on the failure itself, rather than the version. Perhaps interleaving something like this into the check loop would be adequate:

```python
# ...
md5sum = hashlib.md5()
for entry in permissions:
    # ...
    for path in paths:
        # ....
        if not path_matches_expectations(path):
            md5sum.update(path.encode('utf-8'))
            # ...
logger.prefix_log(f'known fs permissions check {md5sum.hexdigest()}')
```

(In hindsight, the fs diff test shouldn't put the version in the first line, either--that's probably a holdover from the dmesg diff test, where the kernel is upgraded at longer intervals.)

recipes-ni/ni-base-system-image-tests/files/known-permissions.csv
dylan-matthew-turner: Yes, but we're not looking at the permissions as numbers directly. Everything here is done after converting it to a human readable string and then comparing, so I think we'd get away with it.

dylan-matthew-turner:
New changes:
- Switched to dictionary object in-code
  + Allows code prefixes for complex and generic permission instead of just bash/glob
  + Matching and represent known files as a tree is more idiomatic
- Added main functions
  + We need to include the new tree functions in the diff test (to exclude the known files), but the lack of main functions interferes when importing
- Recursively fill out file tree, meaning we can leave out more and make it more maintainable
- Support functions for file permissions

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,162 @@
...
+            }, 'runmode': {
+                '.': system_dir,
+                '*': system_file,
+                'bzImage': system_link(f'bzImage-{bash("$(uname -r)")[0]}'),
+            }
+        }, 'lib': {
+            '.': system_dir,
+            'modules': {
+                '.': system_dir,
+                '$(uname -r)': {
```
mike-petersen-ni, line 44: May have intended this:
```suggestion
                bash('$(uname -r)')[0]: {
```

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,162 @@
...
+                '*': system_file,
+                'grubenv': permissions(0o664, 'admin', 'ni', FT_REG),
+                'grubenv.bak': permissions(0o664, 'admin', 'ni', FT_REG),
+                'recoverytool-ni-version': permissions(0o0444, 'admin', 'administrators', FT_REG)
+            }, 'runmode': {
+                '.': system_dir,
+                '*': system_file,
+                'bzImage': system_link(f'bzImage-{bash("$(uname -r)")[0]}'),
+            }
+        }, 'lib': {
```
mike-petersen-ni: formatting nitpick

```suggestion
        },
        'lib: {
```

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,162 @@
...
+import sys
+import hashlib
+import stat
+import pwd
+import grp
+import fs_permissions_shared
+
+# Function to return tree instead of variable since it contains functions inside its structure
+def known_permissions_tree():
+    return {
```
mike-petersen-ni, line 12: Seems we lost some things. [Here](https://github.com/ni/meta-nilrt/commit/ffcbb3c73cbaf95dbd1c5146d84df7ffb2b38835)'s the last state of the CSV:

```
Path,Mode,User,Group,Link Target
/boot/bootmode,-rw-r--r--,admin,administrators,
/boot/grub/!(recoverytool-ni-version),-rw-rw-r--,admin,ni,
/boot/grub/recoverytool-ni-version,-r--r--r--,admin,administrators,
/boot/runmode/!(bzImage),-rw-r--r--,admin,administrators,
/boot/runmode/bzImage,lrwxrwxrwx,admin,administrators,bzImage-$(uname -r)
/etc/fstab,-rw-r--r--,admin,administrators,
/home/admin,drwxr-xr-x,admin,administrators,
/home/lvuser,drwxrwsr-x,lvuser,ni,
/home/root,lrwxrwxrwx,admin,administrators,/home/admin
/home/webserv,drwxr-sr-x,webserv,ni,
/lib/modules/$(uname -r)/!(build|kernel|source),-rw-r--r--,admin,administrators,
/lib/modules/$(uname -r)/@(build|kernel),drwxr-xr-x,admin,administrators,
/lib/modules/$(uname -r)/source,lrwxrwxrwx,admin,administrators,build
/tmp,lrwxrwxrwx,admin,administrators,/var/tmp
```

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,162 @@
...
+        'boot': {
+            '.': system_dir,
+            'bootmode': system_file,
+            'grub': {
+                '.': system_dir,
+                '*': system_file,
+                'grubenv': permissions(0o664, 'admin', 'ni', FT_REG),
+                'grubenv.bak': permissions(0o664, 'admin', 'ni', FT_REG),
+                'recoverytool-ni-version': permissions(0o0444, 'admin', 'administrators', FT_REG)
+            }, 'runmode': {
```
mike-petersen-ni: formatting nitpick
```suggestion
            },
            'runmode': {
```

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,162 @@
...
+    md5sum.update(f'{kind}: {exp} != {act}'.encode('utf-8'))
+    logger.log(f'ERROR: Expected {kind} \'{exp}\' but got \'{act}\' for path {path}.')
+
+# Main check for file/directory permissions
+def permissions(mode, user, group, file_type):
+    def ret(path, stats, logger, md5sum):
+        nonlocal mode
+        nonlocal user
+        nonlocal group
+        nonlocal file_type
```
mike-petersen-ni: formatting nitpick: `nonlocal` [can use a comma-separated list of identifiers](https://docs.python.org/3/reference/simple_stmts.html?highlight=nonlocal#the-nonlocal-statement)

```suggestion
        nonlocal mode, user, group, file_type
```

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,162 @@
...
+                '*': system_file,
+                'bzImage': system_link(f'bzImage-{bash("$(uname -r)")[0]}'),
+            }
+        }, 'lib': {
+            '.': system_dir,
+            'modules': {
+                '.': system_dir,
+                '$(uname -r)': {
+                    '.': system_dir,
+                    '**': system_hier,
```
mike-petersen-ni: Unfortunately, some files under here are executable. The function here will need to accommodate that.

You can use `find /lib/modules/$(uname -r) -type f -perm /111` to discover what those files are.

Perhaps a couple higher-order functions like so could help:

```python
def subset_permissions(included_expectation, excluded_expectation, is_included):
    def ret(path, stats, logger, md5sum):
        nonlocal included_expectation, excluded_expectation, is_included
        if is_included(path):
            return included_expectation(path, stats, logger, md5sum)
        else:
            return excluded_expectation(path, stats, logger, md5sum)
    return ret

def tree_contains(prefix, tree):
    def ret(path):
        # TODO possibly remove prefix, do tree lookup here
    return ret
```

Then, this line could be something like this:

```suggestion
                    '**': subset_permissions(
                        system_executable,
                        system_heir,
                        tree_contains(
                            f'/lib/modules/{bash("$(uname -r)")}',
                            {
                                'build': {
                                    'scripts: {
                                        'checkversion.pl': () # or whatever marker
                                    }
                                }
                            })),
```

(Maybe it'd be simpler to use a list than a tree, but since we're already traversing a tree elsewhere that might be more congruent.)

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: No because paths are [already run through bash](https://github.com/ni/meta-nilrt/pull/557/files#diff-0f9a20fa2e3b0a91cca446d50f2a12047736d56a3439f8bf4df2eb508d30d528R53).

When we traverse the file tree here, it has '.', and that '.' becomes '$(uname -r)' and is then passed through bash.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
mike-petersen-ni: Rather than doing a tree lookup, there's also the option of using heuristics (e.g. file extensions, presence of shebangs or ELF headers) to determine whether a file ought to be executable in a simpler way.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
mike-petersen-ni: In that case, you're missing the run-through-bash step a couple lines prior to where you linked, for the `dict` case.

mike-petersen-ni: Agh, never mind--it'll get expanded per-file. Perhaps a bit wasteful, but not incorrect.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: Do we want to actually keep track of exactly which files are executable?

If not, we can simply add to system_hier to allow files with the executable bit.

If so, I think we could just do something like:

```
{
  ...
  'lib': {
    ...
    'modules': {
      ...
      '$(uname -r)': {
        '.': system_dir,
        '**': system_hier,
        'build': {
          'scripts': {
            '+(*.sh|*.pl)': permissions(0o0755, 'admin', 'administrators', FT_REG),
            'prune-kernel': permissions(0o0755, 'admin', 'administrators', FT_REG)
...
```

Since, if we don't list something, it automatically gets caught by `**`, and vice-versa; if we add a specific permission it will automatically override what gets added with the `**`. The precedence is set by order in the tree.

In other words:

1. `**` gets fleshed out into all the files and they get added to a dictionary mapping their path to `system_hier`.
2. `+(*.sh|*.pl)` gets fleshed out and any matching files replace their entries in the dict with the 0755 permissions check.
3. The specific file listed, `prune-kernel` will also replace its entry in the dict.

I think because of the override and extglob matching, we don't necessarily need to add further logic, although I could wrap `permissions(0o0755, 'admin', 'administrators', FT_REG)` into a `system_script` or `system_executable` function a la `system_dir` and `system_file`

mike-petersen-ni: Sounds like that'll work; give it a try.

dylan-matthew-turner: I just noticed that if you do `bash -O extglob -c "echo /lib/modules/$(uname -r)/**"` it evaluates to the same as just `*`.

Turns out we also need to add the `globstar` extension when running bash, not just `extglob`. Which is why I missed that there were executables not being processed

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: I've readded the missing fstab and home files

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -0,0 +1,317 @@
...
+        if mode != stat.S_IMODE(stats.st_mode):
+            log_mismatch('mode', oct(mode), oct(stat.S_IMODE(stats.st_mode)), path, logger, md5sum)
+            return False
+        if user != pwd.getpwuid(stats.st_uid).pw_name:
+            log_mismatch('user', user, pwd.getpwuid(stats.st_uid).pw_name, path, logger, md5sum)
+            return False
+        if group != grp.getgrgid(stats.st_gid).gr_name:
+            log_mismatch('group', group, grp.getgrgid(stats.st_gid).gr_name, path, logger, md5sum)
+            return False
+        return True
```
mike-petersen-ni: Do we want to do early exit after the first mismatch, or provide all mismatches at once? Later on in `system_link` it appears to perform other checks even when the link target is a known mismatch.

<details><summary>Suggestion</summary>
<p>

```suggestion
        success = True
        if file_type == FT_DIR and not stat.S_ISDIR(stats.st_mode):
            log_mismatch(
                'file type', 'dir', 'reg' if file_type == FT_REG else 'lnk', path, logger, md5sum
            )
            success = False
        if file_type == FT_REG and not stat.S_ISREG(stats.st_mode):
            log_mismatch(
                'file type', 'reg', 'dir' if file_type == FT_DIR else 'lnk', path, logger, md5sum
            )
            success =False
        if file_type == FT_LNK and not stat.S_ISLNK(stats.st_mode):
            log_mismatch(
                'file type', 'lnk', 'dir' if file_type == FT_DIR else 'reg', path, logger, md5sum
            )
            success =False
        if mode != stat.S_IMODE(stats.st_mode):
            log_mismatch('mode', oct(mode), oct(stat.S_IMODE(stats.st_mode)), path, logger, md5sum)
            success =False
        if user != pwd.getpwuid(stats.st_uid).pw_name:
            log_mismatch('user', user, pwd.getpwuid(stats.st_uid).pw_name, path, logger, md5sum)
            success =False
        if group != grp.getgrgid(stats.st_gid).gr_name:
            log_mismatch('group', group, grp.getgrgid(stats.st_gid).gr_name, path, logger, md5sum)
            success =False
        return success
```

</p>
</details>

mike-petersen-ni (approved):
Looks good. Please retest and update PR description.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: Good idea

dylan-matthew-turner:
@mike-petersen-ni I have added the multiple error printing, updated the first comment, and retested everything, so it should be good to go

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -27,13 +27,161 @@ def known_permissions_tree():
...
+            'webserv': permissions(0o2755, 'webserv', 'ni', FT_DIR)
+        },
         'lib': {
             '.': system_dir,
             'modules': {
                 '.': system_dir,
                 '$(uname -r)': {
                     '.': system_dir,
                     '**': system_hier,
+                    'build': {
```
jpautler: I have some concern that listing all the executables and symlinks explicitly could be brittle. If I understand correctly, any time an executable or symlink is added or removed under one of these directories, it is going to cause a test failure and require modifying this code.

For symlinks, it seems better to only validate the permissions (always 0777, correct?) and ownership and just ignore the target of the link. I think our intention in testing known permissions is primarily to ensure secure/appropriate permissions/ownership, which I think can be done without examining the link target. I think we could do this pretty easily by extending system_hier to only check permissions/ownership of links, similar to how it handles both files/directories now. Then we could remove all the symlink entries from this tree.

Regarding executables, I think we can deal with many of them using file extensions (.sh, .pl) as you are now. The trick is the compiled binaries. We could possibly do this dynamically by using some python equivalent of the 'file' shell command or examining ELF headers or similar. I thought we touched on that idea previously. Did we decide it wasn't feasible?

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -27,13 +27,161 @@ def known_permissions_tree():
...
+                                'atomics.tbl': system_file_exec,
+                                'fallbacks': {
+                                    '*': system_file_exec
+                                }
+                            },
+                            'basic': {
+                                'fixdep': system_file_exec
+                            },
+                            'bloat-o-meter': system_file_exec,
+                            'bpf_doc.py': system_file_exec,
```
jpautler: Could we cover this and other .py entries with a **/*.py rule as done above with .sh and .pl?

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
```diff
@@ -27,13 +27,161 @@ def known_permissions_tree():
...
+                            'tracing': {
+                                'draw_functrace.py': system_file_exec
+                            },
+                            'ver_linux': system_file_exec,
+                        },
+                        'tools': {
+                            'lib': {
+                                'lockdep': {
+                                    'lockdep': system_file_exec,
+                                    'run_tests.sh': system_file_exec
```
jpautler: Shouldn't this and other .sh entries by covered already by the **/*.sh rule above?

jpautler (changes requested):

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: No. I tried that. Not all of the .py files are executable. I chose to list the executable ones here because there are less executable .py files than not executable ones.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: I don't remember discussing that idea, but I'll see what I can do. I don't like listing them all out either.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: No. That's for `/lib/modules/$(uname -r)/build/scripts/**/*.sh`
This script is in `/lib/modules/$(uname -r)/build/tools/lib/lockdep/`

I suppose I could move the `**/*.sh` up to the build directory, though it would only add this script and `/lib/modules/$(uname -r)/build/objtool/sync-check.sh`

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: If `readelf -h <file-name>`  succeeds then a file is a program. It will fail on symlinks, directories, and files that are not ELF programs. There might be a way to feed that into find.

dylan-matthew-turner: This should work: `find /lib/modules/$(uname -r)/build/scripts -type f -exec bash -c "if readelf -h {} &> /dev/null; then echo '{}'; fi" \;`

However, that has to check every file under build/scripts, so we would be trading off efficiency for maintainability

mike-petersen-ni: My suggestion previously with `subset_permissions` takes an arbitrary predicate to decide what a particular file should look like. That's one option here: make a predicate that runs `readelf -h` or checks file extensions, and then dispatch appropriately.

jpautler: As Mike suggested, I was thinking of a predicate that would dynamically examine each file already placed in the tree via basic wildcard expansion.

That being said, I don't think **readelf -h** will identify nearly all the files that are currently marked as executable that do not have a file extension. Most of them look to be scripts of some kind. This isn't perfect, but it's closer:
**file <file_path> | grep executable**

It's already clear both extension-based rules and this mechanism are going to find some files that "look" executable but don't actually have execute permissions when installed (and vice-versa in the case of build/scripts/atomic/fallbacks/*). Perhaps we should have a short discussion and determine exactly how to proceed...

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
jpautler: I would say it's worth moving up to the build directory. Pays a small dividend now and maybe more in the future.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_known.py
dylan-matthew-turner: We might be able to use `grep -rIL`

dylan-matthew-turner: Because we mostly care about just that /lib/modules is read/write for admin:administrators, we can add executables to the `system_hier` function and capture them with regular files and directories under the `**`

jpautler: > For symlinks, it seems better to only validate the permissions (always 0777, correct?) and ownership and just ignore the target of the link. I think our intention in testing known permissions is primarily to ensure secure/appropriate permissions/ownership, which I think can be done without examining the link target. I think we could do this pretty easily by extending system_hier to only check permissions/ownership of links, similar to how it handles both files/directories now. Then we could remove all the symlink entries from this tree.

I don't think you resolved this bit of feedback.

dylan-matthew-turner: Ah, yes. I missed that. I'll add them too

jpautler (approved):

amstewart (approved):
Looks good to me.

@dylan-matthew-turner Does this need to be cherry-picked back to hardknott as well?

amstewart:
Cherry picked to hardknott.

PR #561, opened 2023-04-05
Add ptest for measuring and reporting the total boot time

gratian:
Add an NI-specific ptest for logging the total boot time and uploading the data to the NI-internal influxdb data base.

### Testing

- [x] bitbake ni-boot-time
- [x] bitbake packagegroup-ni-ptest-smoke
- [x] installed from local feed and manually run on `cRIO-9042`, and `PXIe-8840`
- [x] verified that the data reported is available via Grafana/InfluxDB
- [x] verified reasonable behavior on error: the test fails if the boot time log is missing or the influxdb connection is not available

### DIFF
--- a/recipes-ni/ni-test-boot-time/files/run-ptest
+++ b/recipes-ni/ni-test-boot-time/files/run-ptest
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+TEST_NAME="test_boot_time"
+TEST_LOG="/var/log/boottime"
+INFLUXDB_INFO="/home/admin/.influxdb.info"
+
+if [[ -f "$INFLUXDB_INFO" ]]; then
+	source "$INFLUXDB_INFO"
+	python3 upload_results.py "-i $TEST_LOG" "-s $INFLUXDB_SERVER" "-p $INFLUXDB_PORT"
+	RESULT=$?
+else
+	echo "INFO: InfluxDB connection information not found at '$INFLUXDB_INFO'"
+	RESULT=77
+fi
+
+case $RESULT in
+    0)
+	echo -n "PASS"
+	;;
+    77)
+	echo -n "SKIP"
+	;;
+    *)
+	echo -n "FAIL"
+	;;
+esac
+echo ": $TEST_NAME"
+exit $RESULT
--- a/recipes-ni/ni-test-boot-time/files/upload_results.py
+++ b/recipes-ni/ni-test-boot-time/files/upload_results.py
@@ -1,0 +1,177 @@
+#!/usr/bin/env python3
+"""Script for uploading the boot time measurement to influxdb"""
+
+import argparse
+import logging
+import mmap
+import re
+import sys
+import subprocess
+from datetime import datetime
+from influxdb import InfluxDBClient
+from influxdb.exceptions import InfluxDBClientError
+from requests.exceptions import ConnectionError
+
+def fw_printenv(key):
+    """Extract firmware information (ex. device, serial#) via fw_printenv"""
+    try:
+        process = subprocess.run(['fw_printenv', key],
+                                 check=True,
+                                 stdout=subprocess.PIPE)
+        value_str = process.stdout.strip().decode("utf-8")
+        value = value_str.split('=')[-1]
+    except (FileNotFoundError, subprocess.CalledProcessError):
+        logging.warning("calling 'fw_printenv %s' failed", key)
+        raise
+    return value
+
+
+def get_device():
+    """Get the device description for the current system (e.g. cRIO-90xx)"""
+    return fw_printenv('DeviceDesc')
+
+
+def get_serial_number():
+    """Get the device serial number; return all-zeros if not present"""
+    try:
+        serialno = fw_printenv('serial#')
+    except (FileNotFoundError, subprocess.CalledProcessError):
+        # Non-fatal error, some early development hardware doesn't have a serial# provisioned
+        logging.warning("serial number not found, defaulting to 00000000")
+        serialno = "00000000"
+    return serialno
+
+
+def get_hostname():
+    """Get the device hostname"""
+    try:
+        process = subprocess.run(['hostname'],
+                                 check=True,
+                                 stdout=subprocess.PIPE)
+        name = process.stdout.strip().decode("utf-8")
+    except (FileNotFoundError, subprocess.CalledProcessError):
+        logging.warning("failed to retrieve the hostname")
+        raise
+    return name
+
+
+def get_kernel_version():
+    """Get the kernel version currently running on the system"""
+    try:
+        process = subprocess.run(['uname', '-r'],
+                                 check=True,
+                                 stdout=subprocess.PIPE)
+        kver_full = process.stdout.strip().decode("utf-8")
+        rgx = re.search(r'([0-9]+\.[0-9]+)\.([0-9]+)', kver_full)
+        kver = rgx.group(1)
+    except (FileNotFoundError, ValueError, AttributeError,
+            subprocess.CalledProcessError):
+        logging.warning("failed to read the current kernel version")
+        raise
+
+    return kver, kver_full
+
+
+def get_os_version():
+    """Extract OS version information from /etc/os-release"""
+    try:
+        with open('/etc/os-release', 'rb', 0) as file, \
+             mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ) as mfile:
+            rgx = re.search(br'VERSION_ID=([0-9.]+)', mfile)
+            version = rgx.group(1).decode()
+
+            rgx = re.search(br'BUILD_ID="([^"]+)"', mfile)
+            build = rgx.group(1).decode()
+
+            rgx = re.search(br'VERSION_CODENAME="([^"]+)"', mfile)
+            codename = rgx.group(1).decode()
+    except (FileNotFoundError, ValueError, AttributeError):
+        logging.warning("failed to parse version information from '/etc/os-release'")
+        raise
+
+    return version, build, codename
+
+
+def get_boot_time(log_file):
+    """Extract boot time from log file"""
+    try:
+        with open(log_file, 'rb', 0) as file, \
+             mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ) as mfile:
+            rgx = re.search(br'([0-9.]+)', mfile)
+            boot_time = float(rgx.group(1).decode())
+    except (FileNotFoundError, ValueError, AttributeError):
+        logging.warning("failed to parse boot time from '%s'", log_file)
+        raise
+
+    return boot_time
+
+
+def read_data(path):
+    """Compile test data and metadata"""
+    controller = get_device()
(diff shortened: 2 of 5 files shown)


recipes-ni/ni-boot-time/ni-boot-time.bb
```diff
@@ -0,0 +1,39 @@
...
+	file://upload_results.py \
+	file://zboottime \
+"
+
+do_install_ptest() {
+	install -m 0755 ${S}/run-ptest		${D}${PTEST_PATH}
+	install -m 0755 ${S}/upload_results.py	${D}${PTEST_PATH}
+
+	install -d ${D}${sysconfdir}/init.d
+	install -m 0755 ${S}/zboottime		${D}${sysconfdir}/init.d
```
amstewart: Can we give this initscript a more informative name, like `zz-ni-record-boottime`?

recipes-ni/ni-boot-time/ni-boot-time.bb
```diff
@@ -0,0 +1,39 @@
...
+	update-rc.d -r ${D} zboottime		start 99 4 5 .
+}
+
+pkg_postinst_ontarget:${PN}-ptest:append() {
+	python3 -m pip install influxdb
```
amstewart: Do we not have an influxdb python3 package in the feeds? Is that intentional, or have we just not gotten around to making one?

recipes-ni/ni-boot-time/ni-boot-time.bb
```diff
@@ -0,0 +1,39 @@
+SUMMARY = "Ptest for measuring and logging the boot time"
```
amstewart: At this point, there are several performance tests in the smoke-ptests group that all upload to our influxdb server; right? Can we merge these all into a single `ni-performance-testing` package?

recipes-ni/ni-boot-time/files/upload_results.py
```diff
@@ -0,0 +1,137 @@
...
+
+import argparse
+import logging
+import mmap
+import re
+import sys
+import subprocess
+from datetime import datetime
+from influxdb import InfluxDBClient, exceptions
+import requests.exceptions
```
amstewart: `requests` isn't python-standard I think. Are you missing an RDEPENDS for this?

amstewart (changes requested):

recipes-ni/ni-boot-time/ni-boot-time.bb
gratian: We do not and I'm not finding an upstream recipe in the OE layer index either. There is a recipe  for the infulxdb i.e. the database itself but not the python package that interfaces with it.

It's not intentional, just not gotten around to it; arguably it should be added to the backlog.

recipes-ni/ni-boot-time/ni-boot-time.bb
gratian: Actually this is the first performance test that we've added to smoke-ptests. I've added it there because it is quick and cheap to run with the other smoke tests.

The other set of tests that upload to influxdb are the kernel-performance-tests which are run separately (due to how long they take) and are not part of smoke-ptests.

recipes-ni/ni-boot-time/ni-boot-time.bb
amstewart: Could you make a TD item for it? Otherwise, I suppose it isn't blocking for this PR.

recipes-ni/ni-boot-time/ni-boot-time.bb
amstewart: Hmm.. OK. Could we rename this recipe to something which makes it more clear that it is a test package then? Maybe `ni-test-boott-time`?

recipes-ni/ni-boot-time/files/upload_results.py
gratian: I'll add an RDEPENDS. Fortunately we do have `python3-requests` in the feed already.
FWIW it works w/o RDEPENDS because it gets pip installed as a dependency of `influxdb`.

gratian:
Changes since v1:

- renamed:    zboottime -> zz-ni-record-boottime
- renamed:    recipes-ni/ni-boot-time/ni-boot-time.bb -> recipes-ni/ni-test-boot-time/ni-test-boot-time.bb and changed all associated references
- added RDEPENDS for `python3-requests`

gratian:
v2 testing

- [x] bitbake packagegroup-ni-ptest-smoke
- [x] opkg install packagegroup-ni-ptest-smoke (from local feed)
- [x] ptest-runner ni-test-boot-time

recipes-ni/ni-boot-time/ni-boot-time.bb
gratian: TD #2358572

recipes-ni/ni-test-boot-time/files/zz-ni-record-boot-time
jpautler, line 1: I think there was originally some discussion about building the write-boot-time-to-logfile behavior directly into our image. That way it would always be present and could be used by anyone for debugging/etc. It could also be the first of potentially other metrics to log (time until network is up, etc). Comments?

recipes-ni/ni-test-boot-time/files/upload_results.py
```diff
@@ -0,0 +1,137 @@
...
+    controller = get_device()
+    kernel_version, kernel_full_version = get_kernel_version()
+    distro_version, distro_build, distro_name = get_os_version()
+    boot_time = get_boot_time(path)
+
+    print("boot time:", boot_time, "(s)")
+
+    data = {
+        "measurement": "boot_performance",
+        "tags": {
```
jpautler, line 122: Would it make sense to log a device hostname/serial number? I'm just imagining what kind of info would be useful if we see an unexpected change. Being able to trace a change to a recently swapped controller-under-test could be helpful.

recipes-ni/ni-test-boot-time/files/run-ptest
```diff
@@ -0,0 +1,21 @@
...
+TEST_LOG="/var/log/boottime"
+INFLUXDB_INFO="/home/admin/.influxdb.info"
+
+if [[ -f "$INFLUXDB_INFO" ]]; then
+	source "$INFLUXDB_INFO"
+	python3 upload_results.py "-i $TEST_LOG" "-s $INFLUXDB_SERVER" "-p $INFLUXDB_PORT"
+	RESULT=$?
+else
+	echo "INFO: InfluxDB connection information not found at '$INFLUXDB_INFO'"
+	RESULT=1
```
jpautler: For local ptest runs or runs of debug dashboard events (events with a "debug" tag), there will be no .influxdb.info file present on the target. I don't know if it's better for this test to "pass" in that case? This test is a little different from some other tests which have a logical pass/fail result independent from whether the data is uploaded or not.

recipes-ni/ni-test-boot-time/files/upload_results.py
```diff
@@ -0,0 +1,137 @@
...
+
+def upload_results(data, server, server_port):
+    """Upload results to influxdb instance"""
+    try:
+        influxdb = InfluxDBClient(host=server, port=server_port)
+        influxdb.switch_database("rtos_boot_performance")
+        if not influxdb.write_points(data):
+            logging.warning("failed to write data points to influxdb")
+        influxdb.close()
+    except (exceptions.InfluxDBClientError,
```
jpautler: I believe an exception here will not trigger the ptest to fail since you are catching and not re-throwing. I don't know if that's what you intend. Since at this point you will have found influxdb info in /home/admin/.influxdb.info, maybe it should cause a failure?

jpautler (changes requested):

amstewart (approved):

recipes-ni/ni-test-boot-time/files/zz-ni-record-boot-time
gratian: The way things are currently architected you can install the base package (i.e. not the -ptest) and get the log:

```
admin@NI-PXIe-8840-00802f23aed1:~# opkg files ni-test-boot-time
Package ni-test-boot-time (1.0-r0.0) is installed on root and has the following files:
/etc/rc4.d/S99zz-ni-record-boot-time
/etc/init.d/zz-ni-record-boot-time
...
/etc/rc5.d/S99zz-ni-record-boot-time
```

We can decide to add this package to the base image however I would like to do that in a separate PR and have a discussion about it. I personally do not think its a good idea for several reasons:

- it is a distro/linux non-standard log.
- it will slow down the boot for everybody and most customers won't know or care about this log.
- it is not a very useful way to measure boot time because it doesn't tell you anything about what is causing the slow down. For that we have more useful tools in the feed, both `bootchart` and `bootchart2` are available.
- if somebody really wants it for their test pipelines they can `opkg install ni-test-boot-time`.

recipes-ni/ni-test-boot-time/files/upload_results.py
gratian: I don't expect large variations for the same controller model unless there's a hardware issue but I can see how having this info would make tracking a hardware issue easier so I'll add it.

recipes-ni/ni-test-boot-time/files/run-ptest
gratian: Since the main action of this test is uploading the boot time to influxdb for later comparisons I have decided it is better to fail the test if the upload fails. That way we would catch if something is wrong and we're not actually capturing results.

If you prefer, I could potentially "skip" the test in the cases where the influxdb info is not present. I am reluctant to always "pass" the test since it is going to hide cases were the reporting stopped working and we won't notice until much later when we see the data is missing.

Long term once we collect enough data from all devices we can consider adding device specific .conf files to this package with thresholds to test against and report a more accurate pass/fail. At this time we don't have enough data to establish these pass/fail thresholds.

gratian:
Changes since v2:

- mark test as SKIP instead of FAIL if the influxdb connection information is not present
- cleaned-up exception handling and re-tested error paths. The test will now fail if the influxdb connection is present but the upload script cannot connect to the server (see jpautler's comment above about re-throwing the exception)
- added metadata for hostname and serial# to the reported data

v3 testing:

- [x] bitbake packagegroup-ni-ptest-smoke
- [x] opkg install ni-test-boot-time-ptest (from local feed)
- [x] ptest-runner
    - [x] verified test is skipped if `.influxdb.info` is missing
    - [x] verified test fails if various error/exceptions paths are activated
    - [x] verified the test now reports hostname and serial# information

recipes-ni/ni-test-boot-time/files/run-ptest
gratian: I went ahead and marked the test SKIP in this situation (see v3).

jpautler (approved):

amstewart (approved):

amstewart:
I assume this should be cherry-picked to kirkstone as well.

amstewart:
Cherry-picked to kirkstone.

PR #562, opened 2023-04-06
[kirkstone] salt: fix restartcheck incompatibility w/ python3.10

amstewart:
Salt 3000.2 has been EOL for a while and lacks broad compatibility with
newer python runtimes, including the python3.10 runtime shipping with
NILRT 10 "kirkstone". As a result, simple workflows like the
restartcheck module to not properly function, causing some software
installation workflows to miss restarting the salt minion when they
should.

Backport upstream salt:master patches which accomplish some parts of
python3.10 compatibility, to at least fix the restartcheck module paths.

The .patches in this set should definitely not be carried forward
if/when we upgrade salt.

* This patchset also adds a missing `python3-singledispatch` dependency. I think this dep was being coincidentally satisfied in prior releases.

NI AZDO: https://dev.azure.com/ni/DevCentral/_workitems/edit/2341699

Work item with a sampling of errors encountered: https://dev.azure.com/ni/DevCentral/_workitems/edit/2329675

# Testing
Without this patch, running even `salt-call --local --help` fails with errors. After this patch, the `salt-call --local restartcheck.restartcheck` method, which is critical to MAX workflows, functions correctly albeit with some warnings about python3.12 incompatibility :(.

While testing this patchset, I found that I couldn't *just upgrade* salt, and have the SW install workflow correctly reboot the target. The `/var/lib/salt/restartcheck_state` data store must first be created by calling any `opkg` module method first. I'm 85% sure that, once this patchset goes in, that data store will be created by some process I don't understand in the default BSI. But I'm running that issue down separately. In either case, this patchset won't make the current broken behavior *worse*.

# Meta
* This patchset is kirkstone-specific and should not be backported to previous mainlines.

### DIFF
--- a/recipes-support/salt/files/0001-Fix-deprecation-warnings-for-imports-from-collection.patch
+++ b/recipes-support/salt/files/0001-Fix-deprecation-warnings-for-imports-from-collection.patch
@@ -1,0 +1,224 @@
+From 49abf954d106048f993b0aceff69627320dac883 Mon Sep 17 00:00:00 2001
+From: Benjamin Drung <benjamin.drung@cloud.ionos.com>
+Date: Fri, 21 Feb 2020 14:01:58 +0100
+Subject: [PATCH] Fix deprecation warnings for imports from collections
+
+DeprecationWarning: Using or importing the ABCs from `collections`
+instead of from `collections.abc` is deprecated since Python 3.3, and in
+3.9 it will stop working.
+
+Therefore try to import the abstract base classes from `collections.abc`
+before falling back to `collections`.
+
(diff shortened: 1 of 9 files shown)


SparkingSpork (approved):
Looks good to me. Any chance of getting these "upstreamed" to the ni salt repo instead of carrying them in our tree?

amstewart:
The plan is to carry this hotpatch for the SW installation workflow without involving the salt maintainers. And then for the SL teams to address the larger salt 3000 incompatibilities in their own way, next cycle. That might mean pulling these patches into our repo, or doing a wholesale salt upgrade, or something else.

PR #563, opened 2023-04-12
fs_permissions_known: correct file permissions

dylan-matthew-turner:
## Reason

In a previous [PR](https://github.com/ni/meta-nilrt/pull/557), two files were added to the known permissions test file tree with incorrect expected permissions:
- `/etc/fstab`:
  + wrong user
  + wrong group
- `/boot/grub/grub-ni-version`:
  + grouped with others in `/boot/grub/*` as `system_file`
  + wrong mode

This happened due to testing permissions on a VM, but was revealed when running the test on real hardware.
In order for the test to succeed in our internal ATS, the permissions need to be corrected

These need to be cherry picked back to hardknott

## Implementation

- `/etc/fstab` to `system_file`
- `/boot/grub/grub-ni-version` to 0o0444

## Testing

The test was modified as shown here and run on the ATS hardware directly


mike-petersen-ni (approved):

amstewart (approved):

amstewart:
Cherry picked to hardknott.

PR #565, opened 2023-04-13
[kirkstone] Add tests for NI-specific locale encoding

SparkingSpork:
This change adds a new ptest package that:
1. Ensures the NI-specific locales defined in the `ni-locale-alias` package are properly installed on a target.
   - This is done through the `locale` and `localedef` tools.
2. Ensures that setting the locale for a program to one of the aliases allows proper conversion between multibyte and unicode/wide character strings.
   - This is done through a C program and text files encoded in the relevant locale encoding.
  
The purpose of these tests are to ensure that common locale usages by internal clients work as expected before those clients receive the image for testing.

[AB#2232374](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2232374)

## Testing
- Built locally and deployed to a target. Confirmed the new tests pass. 
- Mucked with the expected values for the strings, rebuilt, and confirmed the test failed.
- Mucked with the locale settings for each test, rebuilt, and confirmed the test failed.


recipes-core/glibc-locale-tests/glibc-locale-tests/test_locale_aliases.sh
```diff
@@ -0,0 +1,44 @@
...
+
+. /usr/lib/ptest-utils/bash/ptest-format.sh
+
+# Expect these aliases on a Base System Image install.
+EXPECTED_ALIASES=("CP932" "CP936" "L1")
+
+expected_aliases_on_system() {
+    readarray -t locales_and_aliases < <(locale -a | tr "\n" " ")
+    readarray -t installed_locales < <(localedef --list-archive | tr "\n" " ")
+    readarray -t aliases < <(echo ${locales_and_aliases[@]} ${installed_locales[@]} | tr " " "\n" | sort | uniq -u | tr "\n" " ")
```
SparkingSpork, line 13: Happy to add comments explaining this, but basically:
`locale -a` will list all valid locales including aliases.
`localedef --list-archive` will only list the actual installed locales and NOT the aliases.

recipes-core/glibc-locale-tests/glibc-locale-tests/test_locale_aliases.c
```diff
@@ -0,0 +1,106 @@
...
+
+int set_locale(char* locale_alias_name) {
+    setlocale(LC_ALL, locale_alias_name);
+    return strcmp(locale_alias_name, setlocale(LC_ALL, NULL));
+}
+
+int convert_ws_to_mbs(const wchar_t* wide_char_string, char** actual_hex_string) {
+    int len = wcstombs(NULL, wide_char_string, 0);
+    if (len > 0) {
+        *actual_hex_string = malloc((len + 1) * sizeof(char));
```
jpautler, line 29: nit - Assigning pointer here can leak memory if conversion on next line fails, since calling code will assume error means no memory was allocated.

recipes-core/glibc-locale-tests/glibc-locale-tests/test_locale_aliases.c
```diff
@@ -0,0 +1,106 @@
...
+        return wcstombs(*actual_hex_string, wide_char_string, len + 1);
+    }
+    return len;
+}
+
+int convert_mbs_to_ws(const char* multibyte_string, wchar_t** wide_string) {
+    int len = mbstowcs(NULL, multibyte_string, 0);
+    if (len > 0) {
+        *wide_string = malloc((len + 1) * sizeof(wchar_t));
```
jpautler, line 38: nit - Assigning pointer here can leak memory if conversion on next line fails, since calling code will assume error means no memory was allocated.

recipes-core/glibc-locale-tests/glibc-locale-tests/test_locale_aliases.c
```diff
@@ -0,0 +1,106 @@
...
+    }
+    fgets(multibyte_string, 50, fp);
+    fclose(fp);
+
+    if (convert_mbs_to_ws(multibyte_string, &converted_unicode_string) <= 0) {
+        printf("Unable to convert mbs to wcs: %s\n", multibyte_string);
+        return -1;
+    } else if (wcscmp(unicode_string, converted_unicode_string)) {
+        printf("Converted wcs %ls did not match expected string %ls\n", converted_unicode_string, unicode_string);
+        return -1;
```
jpautler: nit - This code path leaks memory assigned to converted_unicode_string.

recipes-core/glibc-locale-tests/glibc-locale-tests/test_locale_aliases.c
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    if (convert_ws_to_mbs(unicode_string, &converted_multibyte_string) <= 0) {
+        printf("Unable to convert wcs to mbs: %ls\n", unicode_string);
+        return -1;
+    } else if (strcmp(multibyte_string, converted_multibyte_string)) {
+        printf("Converted mbs %s did not match expected string %s\n", converted_multibyte_string, multibyte_string);
+        return -1;
```
jpautler: nit - This code path leaks memory assigned to converted_multibyte_string.

jpautler (approved):
A couple nits, but looks good to me.

amstewart (commented):
LGTM. Will there be a commit adding this recipe to one of the ptesting packagegroups, or will that be in its own PR?

recipes-core/glibc-locale-tests/glibc-locale-tests.bb
```diff
@@ -0,0 +1,70 @@
...
+S = "${WORKDIR}"
+
+
+inherit ptest
+
+CC += " ${LDFLAGS}"
+debugsrcdir = "/usr/src/debug/${BPN}"
+
+do_compile() {
+	cd ${S}
```
chaitu236, line 29: I don't think this line is required. Ref [usleep_1.0.bb](https://github.com/ni/meta-openembedded/blob/nilrt/master/kirkstone/meta-oe/recipes-core/usleep/usleep_1.0.bb#L15)

recipes-core/glibc-locale-tests/glibc-locale-tests.bb
```diff
@@ -0,0 +1,70 @@
...
+}
+
+do_install_ptest() {
+	install -m 0755 ${S}/run-ptest                ${D}${PTEST_PATH}
+
+	install -m 0755 ${S}/test_locale_aliases.sh   ${D}${PTEST_PATH}
+	install -m 0755 ${S}/test_locale_aliases      ${D}${PTEST_PATH}
+	install -m 0644 ${S}/CP932.txt                ${D}${PTEST_PATH}
+	install -m 0644 ${S}/CP936.txt                ${D}${PTEST_PATH}
+	install -m 0644 ${S}/L1.txt                   ${D}${PTEST_PATH}
```
chaitu236, line 46: nit: Might be better to specify `WORKDIR` instead of `S` for all these files that are specified in `SRC_URI`. Ref [usleep_1.0.bb](https://github.com/ni/meta-openembedded/blob/nilrt/master/kirkstone/meta-oe/recipes-core/usleep/usleep_1.0.bb#L22)

SparkingSpork:
> LGTM. Will there be a commit adding this recipe to one of the ptesting packagegroups, or will that be in its own PR?

Was going to be separate but realized it's silly to add one line to a packagegroup in a separate PR.

SparkingSpork:
# v2
- Added commit to include new tests in packagegroup-ni-ptest-smoke group.
- Update source file to ensure allocated pointers are always freed.

# Testing:
- Build packagegroup and package locally.

recipes-core/glibc-locale-tests/glibc-locale-tests/test_locale_aliases.c
```diff
@@ -0,0 +1,107 @@
...
+        printf("Unable to open file %s\n", multibyte_file);
+        return -1;
+    }
+    fgets(multibyte_string, 50, fp);
+    fclose(fp);
+
+    if (convert_mbs_to_ws(multibyte_string, &converted_unicode_string) <= 0) {
+        printf("Unable to convert mbs to wcs: %s\n", multibyte_string);
+        return_code = -1;
+    } else if (wcscmp(unicode_string, converted_unicode_string)) {
```
chaitu236, line 67: nitpick: Could you remove the "else" and put this in its own line - makes it easier to understand.

chaitu236 (approved):

recipes-core/glibc-locale-tests/glibc-locale-tests/test_locale_aliases.c
SparkingSpork: In this case, the `else if` is intentional as the comparison shouldn't happen if the first part fails. Since I've changed from early return to return codes, switching to a separate `if` statement would change the behavior and try to compare a potentially unallocated pointer if the first if fails. Going to leave this as is for now.

chaitu236: Ah yes, never mind. For some reason I thought the first if statement's body returned the error code - but seems like we just set the code and return at the end.

SparkingSpork: (It did return the error code in the first draft, you might not have refreshed after my changes from Jeff's feedback)

recipes-core/glibc-locale-tests/glibc-locale-tests.bb
SparkingSpork: I prefer the `cd ${S}` and usage of `${S}` over directly using the `WORKDIR` as it allows us to move the source files into a different location and only have to change the value of `S`. I'd prefer to leave it as is unless you object strongly.

recipes-core/glibc-locale-tests/glibc-locale-tests/test_locale_aliases.c
```diff
@@ -0,0 +1,107 @@
...
+        return mbstowcs(*wide_string, multibyte_string, len + 1);
+    }
+    return len;
+}
+
+int test_locale_conversion(char* locale_alias_name, const char* multibyte_file, const wchar_t* unicode_string) {
+    FILE* fp;
+    char multibyte_string[50];
+    wchar_t* converted_unicode_string;
+    char* converted_multibyte_string;
```
jpautler: nit: There is still a subtle bug here, where convert_ws_to_mbs and convert_mbs_to_ws can return a value <= 0 but not store a new pointer value in converted_unicode_string or converted_multibyte_string. In that case, those variables end up with a random value and I think will crash your program when you try to free them at the bottom of this function. You can avoid this by initializing them to NULL.

For clarity, this happens if the **initial** call to wcstombs/mbstowcs in convert_ws_to_mbs/convert_mbs_to_ws returns a len <=0.

jpautler (changes requested):

recipes-core/glibc-locale-tests/glibc-locale-tests/test_locale_aliases.c
SparkingSpork: Ah, good call. Fixed with the initialization to NULL.

SparkingSpork:
# v3
Added initialization to NULL to protect against pointers being assigned to random locations at initialization.

## Testing
Built locally and confirmed the code built.

jpautler (approved):

amstewart (approved):

amstewart:
@SparkingSpork I don't have access to my ssh keys (temporarily). Could you either cherry-pick this directly into hardknott or post a PR?

PR #567, opened 2023-04-18
Enable create-spdx when building images.

dylan-matthew-turner:
Because [we now skip creating SPDX when a package comes from a remote feed](https://github.com/ni/openembedded-core/pull/90), we can enable creating the spdx zst archives when building images.

To enable it, we need to inherit from create-spdx at the top-level bitbake config, and add it to the list of DISTRO_FEATURES.

I verified this produces the expected archive by building `packagefeed-ni-core` followed by `nilrt-base-system-image`

### DIFF
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -19,7 +19,7 @@ DISTRO_FEATURES += "\
         pam \
         ptest \
         selinux \
-	virtualization \
+        virtualization \
 "
 
 # Because NIRLT uses opkg so intimately, assert that the rootfs should always
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -9,7 +9,10 @@ NILRT_FEED_NAME ?= "next/2023Q3"
 DISTRO_FEATURES:append:x64 = "\
         x11 \
         opengl \
+        create-spdx \
 "
+
+INHERIT += " create-spdx "
 
 VIRTUAL-RUNTIME_xserver_common = "xserver-common"
 


conf/distro/nilrt.inc
```diff
@@ -19,15 +19,16 @@ DISTRO_FEATURES += "\
...
+        create-spdx \
 "
 
 # Because NIRLT uses opkg so intimately, assert that the rootfs should always
 # be made of IPKs.
 PACKAGE_CLASSES =+ "package_ipk"
 
 # Clean *most* package work directories
-INHERIT += " rm_work "
+INHERIT += " rm_work create-spdx "
```
SparkingSpork: Can you add the `create-spdx` tag in a separate `INHERIT +=` line for clarity? This section is labeled as for configuring how the build handles temporary workspaces for each package. We probably want a separate section with comment that adds just the `create-spdx` tag.

SparkingSpork (changes requested):

conf/distro/nilrt.inc
amstewart: MINOR: since the spdx artifacts are a nuance of how _we_ (NI) would like to build the distribution, instead of being a fundamental part of NILRT, I think this code should be moved to `nilrt.conf`.

amstewart (changes requested):

dylan-matthew-turner:
I moved the changes to nilrt.conf from nilrt.inc (and thus also split the INHERIT), but I left a change in nilrt.inc to convert one tab in a string of spaces to all spaces for the sake of consistency.

I rebuilt after moving, and the spdx is still being created

SparkingSpork (approved):

amstewart (approved):

PR #568, opened 2023-04-24
[kirkstone] dkms: Update recipe to version 3.0.10.

SparkingSpork:
An issue upstream was causing warnings to be printed on all unversioned modules. That issue was fixed in version 3.0.10.

[AB#2368804](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2368804)

**Note:** See [here ](https://github.com/dell/dkms/commit/f4c319344d62ff46a15b1e81c4b3feee347956b0)for the fix in question in dkms. 

## Testing performed:
Built a new system image with the new `dkms` version and installed `ni-daqmx` on it. Confirmed the warning no longer appears:
image


gratian (approved):

amstewart (approved):
Is NI doing something weird here by not having versioned modules?

PR #573, opened 2023-05-12
[kirkstone] Build improvements for toolchains

SparkingSpork:
Build improvements including:
- Update `package-index` before image `do_rootfs` for our images, enabling possibility to directly `bitbake nilrt-base-system-image` without prior steps of `package-index` or `packagefeed-ni-core` builds.
- Ensure SDK packages are included in `package-index` builds
- Ensure SDK feed locations are included in `IPK_FEED_URIS` when building SDKs or `meta-toolchain`

Additional Comments:
- I wanted to get nativesdk-* packages added to the toolchain and working for opkg, cmake, and ninja but ran into issues. I don't think these are insurmountable and they should work on Linux toolchain builds. 
   - opkg won't build for mingw32 because of libpcap issues
   - CMake freaks out at Windows file paths in the format the mingw32 toolchain build populates them
   - Didn't try Ninja, but kind of pointless without CMake in there too.

## Testing:
Builds. Lots of builds. 
- From scratch (no build directory)
- In the expected ways (e.g., how our build infrastructure does it now)
- Partial builds
- Package index builds
I highly recommend someone else pulling my branch and doing some testing.

### DIFF
--- a/recipes-core/images/includes/nilrt-image-base.inc
+++ b/recipes-core/images/includes/nilrt-image-base.inc
@@ -6,6 +6,11 @@ do_rootfs[depends] += "shadow-native:do_populate_sysroot"
 # kernel recipe requires depmodwrapper to populate modules.dep
 do_rootfs[depends] += "depmodwrapper-cross:do_populate_sysroot"
 
+# Build package index before trying to build rootfs
+do_rootfs[depends] += "${@oe.utils.conditional('BUILD_IMAGES_FROM_FEEDS', '1', 'package-index:do_package_index', '', d)}"
+do_package_index[nostamp] = "1"
+do_package_index[depends] += "${PACKAGEINDEXDEPS}"
+
 # without package-management update-rc.d gets removed from image
 IMAGE_FEATURES += "package-management"
 
--- a/recipes-core/images/includes/nilrt-image-base.inc
+++ b/recipes-core/images/includes/nilrt-image-base.inc
@@ -11,12 +11,24 @@ do_rootfs[depends] += "${@oe.utils.conditional('BUILD_IMAGES_FROM_FEEDS', '1', '
 do_package_index[nostamp] = "1"
 do_package_index[depends] += "${PACKAGEINDEXDEPS}"
 
+# Normally no package index is created for the following DUMMYARCH but
+# we need one for building SDKs from feeds.
+PACKAGE_ARCHS:append:task-package-index = " sdk-provides-dummy-target"
+SDK_PACKAGE_ARCHS:append:task-package-index = " sdk-provides-dummy-${SDKPKGSUFFIX}"
+
 # without package-management update-rc.d gets removed from image
 IMAGE_FEATURES += "package-management"
 
 IMAGE_INSTALL += "\
 	packagegroup-ni-base \
 	packagegroup-ni-tzdata \
+"
+
+# Add the package sources needed when building the SDK
+IPK_FEED_URIS:append:task-populate-sdk = "\
+    NIOE-dummy-sdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-target \
+    NIOE-nativesdk##${NILRT_LOCAL_FEED_URI}/${SDK_ARCH}-${SDKPKGSUFFIX} \
+    NIOE-dummy-nativesdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-${SDKPKGSUFFIX} \
 "
 
 addtask image_build_test before do_rootfs
--- a/recipes-core/meta/meta-toolchain.bbappend
+++ b/recipes-core/meta/meta-toolchain.bbappend
@@ -2,3 +2,10 @@
 # /usr/local/oe-sdk-hardcoded-buildpath, so avoid causing errors for
 # outgoing symlinks that are there by design.
 CHECK_SDK_SYSROOTS = "0"
+
+# Add the package sources needed when building the SDK with BUILD_IMAGES_FROM_FEEDS
+IPK_FEED_URIS:append:task-populate-sdk = "${@oe.utils.conditional('BUILD_IMAGES_FROM_FEEDS', '1', '\
+    NIOE-dummy-sdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-target \
+    NIOE-nativesdk##${NILRT_LOCAL_FEED_URI}/${SDK_ARCH}-${SDKPKGSUFFIX} \
+    NIOE-dummy-nativesdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-${SDKPKGSUFFIX} \
+', '', d)}"
--- a/recipes-core/meta/package-index.bbappend
+++ b/recipes-core/meta/package-index.bbappend
@@ -1,0 +1,4 @@
+# Normally no package index is created for the following DUMMYARCH but
+# we need one for building SDKs and meta-toolchain from feeds.
+PACKAGE_ARCHS:append = " sdk-provides-dummy-target"
+SDK_PACKAGE_ARCHS:append = " sdk-provides-dummy-${SDKPKGSUFFIX}"


recipes-core/images/includes/nilrt-image-base.inc
```diff
@@ -6,6 +6,16 @@ do_rootfs[depends] += "shadow-native:do_populate_sysroot"
 # kernel recipe requires depmodwrapper to populate modules.dep
 do_rootfs[depends] += "depmodwrapper-cross:do_populate_sysroot"
 
+# Build package index before trying to build rootfs
+do_rootfs[depends] += "${@oe.utils.conditional('BUILD_IMAGES_FROM_FEEDS', '1', 'package-index:do_package_index', '', d)}"
+do_package_index[nostamp] = "1"
+do_package_index[depends] += "${PACKAGEINDEXDEPS}"
```
SparkingSpork, line 12: I heavily debated making this part of `rootfs_<packagetype>.bbclass` instead, but there are a couple of things I think would preclude that:
- `package-index` should probably be a class instead of a task defined in a standalone recipe.
- Building package indexes only impacts local feeds. It was unclear to me if there needed to be differentiation to skip the step if none of the feeds for building were local.

recipes-core/images/includes/nilrt-image-base.inc
```diff
@@ -14,6 +24,13 @@ IMAGE_INSTALL += "\
 	packagegroup-ni-tzdata \
 "
 
+# Add the package sources needed when building the SDK
+IPK_FEED_URIS:append:task-populate-sdk = "\
+    NIOE-dummy-sdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-target \
+    NIOE-nativesdk##${NILRT_LOCAL_FEED_URI}/${SDK_ARCH}-${SDKPKGSUFFIX} \
+    NIOE-dummy-nativesdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-${SDKPKGSUFFIX} \
+"
```
SparkingSpork, line 32: This only needs to apply to SDK tasks as packages in these feeds PROVIDE packages that should be installed normally elsewhere. I didn't see any issues always including them, but this feels "cleaner".

SparkingSpork: It's not currently, but I could also make this conditional like in meta-toolchain.bbappend.

recipes-core/meta/meta-toolchain.bbappend
```diff
@@ -2,3 +2,10 @@
 # /usr/local/oe-sdk-hardcoded-buildpath, so avoid causing errors for
 # outgoing symlinks that are there by design.
 CHECK_SDK_SYSROOTS = "0"
+
+# Add the package sources needed when building the SDK with BUILD_IMAGES_FROM_FEEDS
+IPK_FEED_URIS:append:task-populate-sdk = "${@oe.utils.conditional('BUILD_IMAGES_FROM_FEEDS', '1', '\
+    NIOE-dummy-sdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-target \
+    NIOE-nativesdk##${NILRT_LOCAL_FEED_URI}/${SDK_ARCH}-${SDKPKGSUFFIX} \
+    NIOE-dummy-nativesdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-${SDKPKGSUFFIX} \
+', '', d)}"
```
SparkingSpork, line 11: This won't immediately impact our toolchain builds as we disable `BUILD_IMAGES_FROM_FEEDS` when building currently. That being said, this would allow us to eventually back that out of openembedded-core.

recipes-core/meta/package-index.bbappend
```diff
@@ -0,0 +1,4 @@
+# Normally no package index is created for the following DUMMYARCH but
+# we need one for building SDKs and meta-toolchain from feeds.
+PACKAGE_ARCHS:append = " sdk-provides-dummy-target"
+SDK_PACKAGE_ARCHS:append = " sdk-provides-dummy-${SDKPKGSUFFIX}"
```
SparkingSpork, line 4: I'd like to just fix these being excluded from package-index builds altogether but that currently builds the indexes for values in `ALL_MULTILIB_PACKAGE_ARCHS` and `SDK_PACKAGE_ARCHS` and it's not immediately clear to me what adding these globally might impact elsewhere.

SparkingSpork: Note that `PACKAGE_ARCHS` and `SDK_PACKAGE_ARCHS` are initially set in [`bitbake.conf`](https://github.com/ni/openembedded-core/blob/nilrt/master/hardknott/meta/conf/bitbake.conf).

SparkingSpork:
Also forgot to mention - If you build the SDK for our current `nilrt-runmode-rootfs`, the sysroot ends up being 5+ GB when extracted and the overall toolchain is ~1.5 GB when compressed. Note that it _does_ install *-dev *-staticdev *-doc *-dbg -src -ptest *-bash-completion packages for **everything** in the image.

recipes-core/images/includes/nilrt-image-base.inc
amstewart: I think the only reason we wouldn't want to add this task-dependency to all images when doing `BUILD_FROM_FEEDS` is that it would be unnecessary for all images beyond the first - and it is kind of slow.

>package-index should probably be a class instead of a task defined in a standalone recipe.

I like the existing implementation as an independent target, since it enables us to build indexes for the feeds in the OE_feeds component without invoking an image target.

>Building package indexes only impacts local feeds. It was unclear to me if there needed to be differentiation to skip the step if none of the feeds for building were local.

Nah; doesn't hurt to run the task if there are no local feeds. Not worth the complexity of adding the skip-logic, I don't think.

recipes-core/meta/meta-toolchain.bbappend
```diff
@@ -2,3 +2,10 @@
 # /usr/local/oe-sdk-hardcoded-buildpath, so avoid causing errors for
 # outgoing symlinks that are there by design.
 CHECK_SDK_SYSROOTS = "0"
+
+# Add the package sources needed when building the SDK with BUILD_IMAGES_FROM_FEEDS
+IPK_FEED_URIS:append:task-populate-sdk = "${@oe.utils.conditional('BUILD_IMAGES_FROM_FEEDS', '1', '\
+    NIOE-dummy-sdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-target \
+    NIOE-nativesdk##${NILRT_LOCAL_FEED_URI}/${SDK_ARCH}-${SDKPKGSUFFIX} \
+    NIOE-dummy-nativesdk##${NILRT_LOCAL_FEED_URI}/sdk-provides-dummy-${SDKPKGSUFFIX} \
```
amstewart, line 10: Is there a way that we can make these URIs generic, and add this logic into the OE-core layer with the other BUILD_IMAGES_FROM_FEEDS stuff?

recipes-core/meta/meta-toolchain.bbappend
amstewart: Back what out of OE-core? BIFF or logic to disable BIFF when building the toolchain targets?

recipes-core/images/includes/nilrt-image-base.inc
SparkingSpork: > I think the only reason we wouldn't want to add this task-dependency to all images when doing `BUILD_FROM_FEEDS` is that it would be unnecessary for all images beyond the first - and it is kind of slow.

It actually seems to run pretty quickly for me as part of the build. Any suggestions on how to benchmark the time it adds to a build?

> I like the existing implementation as an independent target, since it enables us to build indexes for the feeds in the OE_feeds component without invoking an image target.

We can actually have our cake and eat it too. Example case: `meta-toolchain` literally just [inherits](https://github.com/ni/openembedded-core/blob/nilrt/master/hardknott/meta/recipes-core/meta/meta-toolchain.bb#LL15C1-L15C21) `populate-sdk` and builds that. My plan was to keep the specific `package-index` recipe but move the logic to a class instead of having it live in the recipe.

> Nah; doesn't hurt to run the task if there are no local feeds. Not worth the complexity of adding the skip-logic, I don't think.

Cool, agreed then.

recipes-core/meta/meta-toolchain.bbappend
SparkingSpork: The logic to disable BIFF when building the toolchain targets. We have two custom changes in OE-core that I don't think were upstreamed to do that ([here ](https://github.com/ni/openembedded-core/commit/ddec9e153978b261f350cbb56b662abbbd5ca378) and [here](https://github.com/ni/openembedded-core/commit/312af55e6cd0b3f57b3e66f57bbd597aea564372)).

recipes-core/meta/meta-toolchain.bbappend
SparkingSpork: The `NILRT_LOCAL_FEED_URI` could be replaced with `DEPLOY_DIR_IPK` or `DEPLOY_DIR_<IMAGE_PKGTYPE>`. The part I'm unable to figure out is if the name portion before the `##` for each feed is required.

recipes-core/images/includes/nilrt-image-base.inc
chaitu236: > I think the only reason we wouldn't want to add this task-dependency to all images when doing BUILD_FROM_FEEDS is that it would be unnecessary for all images beyond the first - and it is kind of slow.

It used to take ~10 mins previously but we [disabled](https://github.com/ni/openembedded-core/pull/70) building package filelist generation and I believe it takes ~10secs now.

I'm not opposed to adding indexing stage into one of the existing recipes but I'd prefer it go in packagegroup recipes instead of image recipes. Better yet, make it part of [packagegroup.bbclass](https://github.com/ni/openembedded-core/blob/nilrt/master/kirkstone/meta/classes/packagegroup.bbclass#L1).

SparkingSpork: > I'm not opposed to adding indexing stage into one of the existing recipes but I'd prefer it go in packagegroup recipes instead of image recipes. Better yet, make it part of [packagegroup.bbclass](https://github.com/ni/openembedded-core/blob/nilrt/master/kirkstone/meta/classes/packagegroup.bbclass#L1).

I prefer having it be a prerequisite step for `do_rootfs` instead for a couple of reasons:
- The package index isn't specifically required when building a packagegroup, but it _is_ required to build an image from feeds. Additionally, I was gating it logically on BIFF being enabled.
- If the packagegroup is where the index build occurs, I'm not sure we get to keep the one command image build that is a big advantage of this to me. E.g., you would have to build the packagegroup and then the image, whereas having it build the index as part of the image build results in only having to build the image.
   - As an aside, I'm actually a bit confused on this. Should BIFF result in all dependencies being built before trying to build the image? If not, then maybe adding it to the packagegroup is right. Or even just having the class and inheriting it in our specific `packagefeed` packagegroups since packagegroups aren't necessarily feeds.

SparkingSpork:
Closing for now. @chaitu236, @amstewart - I'll try to knock out an OpenEmbedded-core version of these changes and post a PR for comparison.

PR #580, opened 2023-06-22
kernel-performance-tests: Read iperf info from external file 

jpautler:
Read iperf connection info from an external file on the target rather than a static file in the ptest package. This matches how we already handle influxdb connection info.


jpautler:
Accidentally created pull request for wrong branch. Abandoning.

PR #590, opened 2023-07-07
Remove salt patches

stefanc18:
The patches are being applied upstream:
- https://github.com/ni/salt/pull/54
- https://github.com/ni/salt/pull/55
- https://github.com/ni/salt/pull/56

There is still one patch left which wasn't removed because we haven't added it upstream it. We will after those PRs from above get in, since it is dependent on them. Will follow up to remove the latest patch as soon as we have ready the PR to push that one upstream as well.


amstewart:
We use commits as the atomic unit of change in the OE layers. Could you squash your commits and make sure the commit message is meaningful and follows the styling of the other commits in the layer? And also sign your commits with `git commit -s`.

chotea (approved):

stefanc18:
@amstewart I updated the commit, let me know if it is okay now

amstewart:
It still needs some commentary about why the patches are being removed, and needs your sign-off.

PR #593, opened 2023-07-17
salt: remove patch

stefanc18:
Remove 0008-Only-import-ABCs-from-collections.abc.patch as it is applied upstream.


chotea (approved):

mike-petersen-ni:
Related: ni/salt#60

In an out-of-band conversation, this change is said to fix build failures (errors are encountered as this patch no longer applies cleanly, presumably due to #590 and the various salt PRs it mentions). I don't know if that has been confirmed.

chaitu236 (approved):

PR #595, opened 2023-07-19
Report distro info to DB during Kernel Performance tests

dylan-matthew-turner:
## Need

We want to be able to sort by distro in our test results

## Implementation

During the test, /etc/os-release is read and parsed for VERSION_ID and DISTRO_CODENAME, and those are stored in the results database as tags called "distro_version" and "distro_name."

In the event that the file cannot be read, the file is improperly formatted, or the keys are missing, the resulting values for those tags will be "unknown".

## Testing

Ran tests locally on a 9037 and verified that distro_name and distro_version show up in Grafana (will need to remove test data)

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-performance-tests-files/upload_cyclictest_results.py
+++ b/recipes-kernel/kernel-tests/kernel-performance-tests-files/upload_cyclictest_results.py
@@ -132,6 +132,22 @@ def get_test_results(path):
 
     return min_latency, avg_latency, max_latency
 
+def get_distro_info():
+    """Extract distro version and codename from os-release and return them as a tuple
+
+    Returns 'unknown' for either in case of file read exception or missing fields.
+
+    """
+    os_release_stmts = []
+    try:
+        with open('/etc/os-release') as os_release_file:
+            os_release_stmts = os_release_file.readlines()
+    except:
+        return ('unknown', 'unknown')
+    pairs = dict(map(lambda e: e.split('='), os_release_stmts))
+    version = pairs.get('VERSION_ID', 'unknown').strip()
+    name = pairs.get('DISTRO_CODENAME', 'unknown').strip()[1:-1]
+    return (version, name)
 
 def read_data(path):
     """Read test data and metadata from a cyclictest log file"""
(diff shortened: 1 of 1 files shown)


recipes-kernel/kernel-tests/kernel-performance-tests-files/upload_cyclictest_results.py
```diff
@@ -132,21 +132,42 @@ def get_test_results(path):
...
+    Returns 'unknown' for either in case of file read exception or missing fields.
+
+    """
+    os_release = ''
+    try:
+        with open('/etc/os-release') as os_release_file:
+            os_release = os_release_file.read()
+    except:
+        return ('unknown', 'unknown')
+    stmts = os_release.split('\n')
```
amstewart: I think, if you open in `"r"` mode, you'll be given a TextWrapper object, and then you can just read it with `.readlines()` or something and let the standard libs split lines for you. Does that work?

recipes-kernel/kernel-tests/kernel-performance-tests-files/upload_cyclictest_results.py
```diff
@@ -132,21 +132,42 @@ def get_test_results(path):
...
+    pairs = list(map(lambda e: e.split('='), stmts))
+    variables = dict(pairs)
```
amstewart: Isn't the `map()` already basically a `dict()`? It seems like you should be able to skip the intermediate list here.

amstewart (changes requested):
LGTM; pending a few minor suggestions.

dylan-matthew-turner:
I simplified based on your suggestions and have now run the tests and seen them shown up in Grafana.

During that I noticed one other change necessary which was to strip the newlines as they got reported to Grafana, so I added that change as well.

dylan-matthew-turner:
Oops, and also I just had to remove surrounding quotes from the reported name. `"kirkstone"\n` to `"kirkstone"` to `kirkstone`

Should be good to go now.

dylan-matthew-turner:
I can rebase and make these all back to one commit if that's preferable

amstewart:
Sure.

blueOkiris:
Done

dylan-matthew-turner:
Done

amstewart (approved):

PR #596, opened 2023-07-19
[kirkstone] packagegroup-ni-internal-deps: Add lz4 package as dep

SparkingSpork:
Adding the lz4 package as a required package so that it is included in the core feed. There is an internal ATS step that builds python on the target and lz4-dev is required to do so.

[AB#2458968](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2458968)

## Testing:
Build packagefeed-ni-core and confirmed it built without errors. Confirmed the lz4-dev package is present.

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
@@ -53,3 +53,11 @@ RDEPENDS:${PN}:append:x64 = "\
 RDEPENDS:${PN} += "\
 	libxml-parser-perl \
 "
+
+# Required for a mobilize step that installs a specific Python version 
+# and requires building Python on the test system
+# Contact: ulf.glaeser@ni.com
+# Team: DAQ.SW.Ops@ni.com
+RDEPENDS:${PN} += "\
+	lz4 \
+"


chaitu236 (approved):

recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
```diff
@@ -53,3 +53,10 @@ RDEPENDS:${PN}:append:x64 = "\
 RDEPENDS:${PN} += "\
 	libxml-parser-perl \
 "
+
+# Required for a mobilize step that installs a specific Python version 
+# and requires building Python on the test system
+# Contact: daq.sw.ops@ni.com
```
amstewart: Is there an actual engineer who can speak to the nature of this dependency?

amstewart (changes requested):

recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
SparkingSpork: Yes, I can put the tech lead for that team instead of their group email. They indicated that they preferred this if possible though. Do I need to switch it to a specific person?

SparkingSpork: Fixed.

SparkingSpork:
## v2
Change contact to specific person instead of email group.

amstewart (approved):

PR #602, opened 2023-07-25
[kirkstone] pkggrp-ni-internal-deps: add PXIPS ATS deps

amstewart:
Per [NI AZDO 2463687](https://dev.azure.com/ni/DevCentral/_boards/board/t/RTOS/Work%20Items/?workitem=2463687), the PXI PS ATS depends on the `memtester` package.

Express this dependency in the internal-deps packagegroup, to ensure that it is built and included in the core feed.

# Testing
* [x] Built `packagegroup-ni-internal-deps` with this change; validated that `memtester` is built as a dependency.

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
@@ -87,3 +87,10 @@ RDEPENDS:${PN} += "\
 	make \
 	pkgconfig \
 "
+
+# Required by PXIPS ATS events
+# Team: PXI PS
+# Contact: Kevin Khai-Wern Lim
+RDEPENDS:${PN} += "\
+	memtester \
+"
--- a/recipes-core/packagegroups/packagefeed-ni-extra.bb
+++ b/recipes-core/packagegroups/packagefeed-ni-extra.bb
@@ -232,7 +232,6 @@ RDEPENDS:${PN} += "\
 	iperf2 \
 	iperf3 \
 	lmbench \
-	memtester \
 	nbench-byte \
 	phoronix-test-suite \
 	tiobench \


chaitu236 (approved):

recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
```diff
@@ -87,3 +87,10 @@ RDEPENDS:${PN} += "\
 	make \
 	pkgconfig \
 "
+
+# Required by PXIPS ATS events
+# Team: PXI PS
+# Contact: Kevin Khai-Wern Lim
+RDEPENDS:${PN} += "\
+	memtester \
```
gratian, line 95: Do we need to remove `memtester` from [packagefeed-ni-extra.bb](https://github.com/ni/meta-nilrt/blob/7ffc454b3fab7c3ee5c4d3aff119a0751f6e63d5/recipes-core/packagegroups/packagefeed-ni-extra.bb#L235) since we're adding it here?

amstewart: No. It should be automatically deconflicted by the build scripting.

gratian: I get that the build tooling can de-conflict it but from a maintenance perspective it seems odd to me to have it both in `packagefeed-ni-extra.bb` and `packagefeed-ni-core.bb` (via packagegroup-ni-internal-deps). In the past IIRC when we moved things into internal deps we removed it from extra.

amstewart: You're right that people do that, but it has never been necessary. I prefer to be conservative wrt. ensuring that needed deps get built into **a** feed, and avoid the possibility that something gets accidentally dropped from all he feeds.

amstewart: Will fix.

gratian (approved):

PR #605, opened 2023-08-04
Add bolt

chaitu236:
These commits add bolt support in nilrt.

WI: [2093177](https://dev.azure.com/ni/DevCentral/_workitems/edit/2093177)

### Testing
Tested on PXIe-8861 connected to PXIe-8301 with a PXI-6683H in same chassis.

With changes from meta-openembedded https://github.com/ni/meta-openembedded/pull/41,

- [x] Built `bolt` package.
- [x] Verified that `boltd` is started by initscript if thunderbolt device is connected before boot up.
- [x] Verified `boltd` is not started if thunderbolt device isn't connected before boot up.
- [x] Verified that `boltd` is started by udev if thunderbolt device is connected after boot up.
- [x] Verified that `enroll`ed device is authorized automatically by bolt whether thunderbolt device is connected before/after host is booted.
- [x] Verified other `boltctl` arguments `authorize`, `forget`, `list` work as expected.


recipes-core/packagegroups/packagegroup-ni-desirable.bb
```diff
@@ -59,3 +59,8 @@ RDEPENDS:${PN} += "\
 RDEPENDS:${PN}:append:x64 = "\
 	kernel-test-nohz \
 "
+
+# Thunderbolt support packages
+RDEPENDS:${PN} += "\
```
chaitu236: Discussions around whether we're going to ship bolt with BSI or put it in feeds are ongoing.
Will update this PR and close this issue when we decide what approach we want to take.

If we decide to ship in feeds, I'd like to move this to core feeds so it is available in offline feeds.
Seeking opinions on this and which packagegroup this should go into.

chaitu236: @amstewart any thoughts on which packagegroup this should go into?

recipes-bsp/bolt/bolt/0001-Update-udev-rule-for-hosts-that-don-t-have-systemd.patch
amstewart, line 1: I would prefer to have this live as a standalone file rather than a .patch file, unless the upstream udev rule is actually *broken* without systemd. ie. Install the new udev rule as something like `/etc/udev.d/rules/91-bolt-ni.rules` or something.

recipes-bsp/bolt/bolt/bolt
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# Strart boltd when thunderbolt devices are present at boot time.
```
amstewart: s/Strart/Start/

recipes-bsp/bolt/bolt/boltctl
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# The purpose of this wrapper is to set LANG to UTF-8 so that non ASCII
+# characters in the output of boltctl can be displayed correctly
+LANG=en_US.UTF-8 /usr/libexec/boltctl "$@"
```
amstewart, line 5: You don't want double quotes around the `$@` in this case; right? I think that would eval to 
```
LANG=en_US.UTF-8 /usr/libexec/boltctl "$0 $1 $2 ..."
```

recipes-bsp/bolt/bolt_%.bbappend
```diff
@@ -0,0 +1,23 @@
...
+FILES:${PN} += "${sysconfdir}/init.d/bolt"
+
+do_install:append() {
+   install -d ${D}${sysconfdir}/init.d
+   install -m 0755 ${WORKDIR}/bolt ${D}${sysconfdir}/init.d/bolt
+
+   # Move boltctl to a directory that's not in PATH and install a wrapper
+   # in its place
+   mv ${D}${bindir}/boltctl ${D}${libexecdir}/
+   install -m 0755 ${WORKDIR}/boltctl ${D}${bindir}/
```
amstewart, line 27: As long as we're messing with the binary paths: Is it sane for a non-privileged user to run this utility? If not, then we should consider putting it in `/sbin` instead.

recipes-bsp/bolt/bolt/bolt
amstewart, line 1: Could you add an LSB initscript header to this file?  We're already half-way-there with the description comment below anyway, and it will help to make the migration to systemd more reliable.

MINOR though, since I think I recall you saying that the upstream recipe already has a systemd unit file.

recipes-core/packagegroups/packagegroup-ni-desirable.bb
amstewart: We don't have a good packagegroup for it, unfortunately. The best place at this time is probably directly into the [core packagefeed](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/recipes-core/packagegroups/packagefeed-ni-core.bb) (like `dkms`).

conf/distro/nilrt.inc
```diff
@@ -17,6 +17,7 @@ DISTRO_FEATURES += "\
         pci \
         sysvinit \
         pam \
+        polkit \
```
amstewart: Could I get some more commentary about what the side-effects are of enabling polkit? Changing `DISTRO_FEATURES` is a relatively big deal, as it impacts the configuration of a lot of recipes.

amstewart (changes requested):

recipes-bsp/bolt/bolt/boltctl
mike-petersen-ni: [`$@` has special quoting rules](https://www.gnu.org/software/bash/manual/html_node/Special-Parameters.html#index-_0040)

amstewart: Oh. Didn't know that. :thinking:

recipes-bsp/bolt/bolt_%.bbappend
chaitu236: The utility can also be used for listing connected thunderbolt devices, their authorization status, etc. I think these operations should be allowed for non-`privileged` users also.

Access to system-configuration-changing-commands of the utility are already controlled by polkit (or dbus rules that I'm now testing).

recipes-bsp/bolt/bolt/0001-Update-udev-rule-for-hosts-that-don-t-have-systemd.patch
chaitu236: Replaced existing rules file as discussed offline.

conf/distro/nilrt.inc
chaitu236: Removed polkit. Now handing authorization from dbus.

recipes-bsp/bolt/bolt/0001-daemon-support-running-without-PolicyKit.patch
```diff
@@ -0,0 +1,180 @@
...
+
+Make the dependency on PolicyKit optional. In that case, don't
+allow all methods and property access and fall back on the
+standard D-Bus policies. NB: These are designed to work with
+PolicyKit so most likely are not suitable at all. This is
+designed for a special use case not intended for general use.
+
+Since PolicyKit is the normal mode of operation, this feature
+requires explicit opt-in, via the new `require-polkit` configure
+option.
```
amstewart, line 14: Upstream-status?

amstewart (changes requested):

recipes-bsp/bolt/bolt/0001-daemon-support-running-without-PolicyKit.patch
chaitu236: The [patch](https://gitlab.freedesktop.org/bolt/bolt/-/commit/86d2596b8da2bef9dd33579b104916aaafbff3ab) is from a `no_polkit` branch in upstream that hasn't been merged in master. Not sure if it's acceptable to add Upstream-Status to a patch not authored by us.

chaitu236:
v2: Remove polkit from REQUIRED_DISTRO_FEATURES as it was added to [meta-oe PR](https://github.com/ni/meta-openembedded/pull/41) per upstream's feedback.

recipes-bsp/bolt/bolt/0001-daemon-support-running-without-PolicyKit.patch
amstewart: The upstream-status is still relevant to include, even if it is a .patch from upstream.

chaitu236:
v3: Added Upstream-Status

recipes-bsp/bolt/bolt/0001-daemon-support-running-without-PolicyKit.patch
chaitu236: Done

amstewart (approved):

PR #608, opened 2023-08-11
Create test to ensure we don't unusually increase size of BSI

dylan-matthew-turner:
The size of our images will likely slowly grow over time, and potentially we could increase faster than we are prepared for. We need to make sure we don't exceed the storage capacity of our smaller devices anytime soon.

Using an estimation of how fast we can grow gives us a value that can be added to our current image size to set the max amount we're allowed to exceed without intervention through 2024. This value is different for ARM (~300MB) and x86_64 (~1400MB).

I've added this test to the existing base system image tests, which run on a newly installed base system image which we can get the install size of to compare against our threshold. The failure state is simply "does the install size exceed the max-size threshold."

This has been run on a cRIO-9037 to verify the test will function.

### DIFF
--- a/recipes-ni/ni-base-system-image-tests/files/disk_size.py
+++ b/recipes-ni/ni-base-system-image-tests/files/disk_size.py
@@ -1,0 +1,82 @@
+import datetime
+import subprocess
+import os
+import argparse
+import sys
+import logging
+
+logging.basicConfig(level = logging.INFO)
+
+TEST_DESC = 'Test to see that the file size of the base system image is not getting too large.'
+
+# Filled in with CLI args
+X86_MAX_SIZE_EOY = -1
+ARM_MAX_SIZE_EOY = -1
+
+def get_max_size():
+    arch = run_cmd([ 'uname', '-m' ]).strip()
+    if arch == 'x86_64':
+        return X86_MAX_SIZE_EOY
+    elif arch == 'arm':
+        return ARM_MAX_SIZE_EOY
+    else:
+        logging.error(f'ERROR: Failed to get size limit! Unknown architecture: "{arch}"')
+        return -1
+
+def run_cmd(cmd):
+    output = subprocess.run(cmd, check = False, capture_output = True)
+    return output.stdout.decode('utf-8')
+
+def du_path(path):
+    du_result = run_cmd([ 'du', '-hs', '-BM', path ]) # Format: "####M      /"
+    size_str = du_result.split('\t')[0].strip()[:-1]
+    return int(size_str)
+
+def get_disk_used():
+    root_size = du_path('/')
+    boot_size = du_path('/boot')
+    return root_size + boot_size
+
+def test_disk_size():
+    logging.info('###### Testing install size against estimated max growth')
+    
+    size_limit = get_max_size()
+    logging.info(f'Size Limit: {size_limit}')
+
+    actual_size = get_disk_used()
+    logging.info(f'Actual Size: {actual_size}')
+
+    success = actual_size < size_limit
+
+    logging.info('###### Finished testing install size')
+    return success
+
+def parse_args():
+    parser = argparse.ArgumentParser(description = TEST_DESC)
+    parser.add_argument(
+        '--x86_max_size', required = True, type = int,
+        help = 'Current size limit for x86 devices'
+    )
+    parser.add_argument(
+        '--arm_max_size', required = True, type = int,
+        help = 'Current size limit for ARM devices'
+    )
+    return parser.parse_args()
+
+def main():
+    global X86_MAX_SIZE_EOY
+    global ARM_MAX_SIZE_EOY
+    args = parse_args()
+    X86_MAX_SIZE_EOY = args.x86_max_size
+    ARM_MAX_SIZE_EOY = args.arm_max_size
+
+    result = test_disk_size()
+
+    if result:
+        sys.exit(os.EX_OK)
+    else:
+        sys.exit(os.EX_SOFTWARE)
+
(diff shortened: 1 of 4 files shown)


recipes-ni/ni-base-system-image-tests/files/disk_size.py
amstewart, line 1: Seems like this commit should be squashed into the previous, since these commits' scripts call each other.

recipes-ni/ni-base-system-image-tests/files/disk_size.py
```diff
@@ -0,0 +1,89 @@
...
+        self.logs.append(log)
+
+    def prefix_log(self, log):
+        self.prefix_logs.append(log)
+
+    def report(self):
+        for log in self.prefix_logs:
+            print(log)
+        for log in self.logs:
+            print(log)
```
amstewart: Is there a reason you're not using the standard python `logging` module for this? That module is pretty neat and I use it all the time.

recipes-ni/ni-base-system-image-tests/files/disk_size.py
```diff
@@ -0,0 +1,89 @@
...
+ARM_EOL_YEAR = 2038         # When the cRIO ___ goes EOL
+ARM_INC_RATE = (ARM_SMALLEST_DRIVE - ARM_BASE_SIZE) / (ARM_EOL_YEAR - BASE_YEAR)
+ARM_MAX_SIZE_EOY = (ARM_INC_RATE * (NEXT_YEAR - BASE_YEAR)) + ARM_BASE_SIZE
+
+# Same thing, but for x86
+X86_BASE_SIZE = 1138        # Image size as of writing. Update when changing x86 device
+X86_SMALLEST_DRIVE = 4096   # The size of cRIO ___'s hardrive
+X86_EOL_YEAR = 2038         # When the cRIO ___ goes EOL
+X86_INC_RATE = (X86_SMALLEST_DRIVE - X86_BASE_SIZE) / (X86_EOL_YEAR - BASE_YEAR)
+X86_MAX_SIZE_EOY = (X86_INC_RATE * (NEXT_YEAR - BASE_YEAR)) + X86_BASE_SIZE
```
amstewart: I know you replace this block later in the patchset, but could you squash down the changes do that there aren't false advertisements of EOL dates in the git history?

recipes-ni/ni-base-system-image-tests/files/run-ptest
amstewart, line 1: The commit message here should be `ni-base-system-image-tests: ...`.

But honestly, I think this could also be silently squashed into the previous commits.

recipes-ni/ni-base-system-image-tests/ni-base-system-image-tests.bb
amstewart, line 1: Squash.

recipes-ni/ni-base-system-image-tests/files/disk_size.py
```diff
@@ -0,0 +1,89 @@
...
+    elif arch == 'arm':
+        return ARM_MAX_SIZE_EOY
+    else:
+        logger.log(f'ERROR: Failed to get size limit! Unknown architecture: "{arch}"')
+        return -1
+
+def get_disk_used():
+    # Format: "####M      /"
+    root_size = int(run_cmd([ 'du', '-hs', '-BM', '/' ]).split('\t')[0].strip()[:-1])
+    boot_size = int(run_cmd([ 'du', '-hs', '-BM', '/boot' ]).split('\t')[0].strip()[:-1])
```
amstewart: I think these operations are complicated enough that they should be moved into their own method. eg.

```python3
def du_path(path):
    proc = subprocess.run(["du", "-hs", ...] + [path], ...)
    result = proc.stdout ...
    return result
```

amstewart (changes requested):

recipes-ni/ni-base-system-image-tests/files/disk_size.py
blueOkiris: This is how our other tests do it, but outside of that, no

recipes-ni/ni-base-system-image-tests/files/disk_size.py
blueOkiris: Good catch

recipes-ni/ni-base-system-image-tests/files/disk_size.py
```diff
@@ -0,0 +1,76 @@
...
+import subprocess
+import os
+import argparse
+import sys
+
+TEST_DESC = 'Test to see that the file size of the base system image is not getting too large.'
+
+# See internal wiki for calculation
+X86_MAX_SIZE_EOY = 1400
+ARM_MAX_SIZE_EOY = 300
```
SparkingSpork: I'd suggest making these parameters for the script that the ptest bash script passes instead of constants in the script. We might have additional limits in the future (e.g., if we add ARM 64-bit support or other images) and having them defined in the test script might be easier to read or update.

recipes-ni/ni-base-system-image-tests/files/disk_size.py
chaitu236: The only reason our other tests used this was because of prefix_log. It looks like you're not using it so you can probably just use logging module

dylan-matthew-turner:
Changes:

- Moved size constants to CLI arguments
- Squashed commits into one since all the scripts call each other
- Pulled out du command into its own function
- Switched to Python logging library instead

amstewart (approved):

SparkingSpork (approved):

PR #609, opened 2023-08-17
[kirkstone] Fix broken packagefeed-ni-extra build and move trace-command to desirable. 

SparkingSpork:
This PR consists of two changes:
- Remove `tbtadm` from `packagefeed-ni-extra` as the recipes for that component in both meta-nilrt and meta-openembedded no longer exist. This was causing `packagefeed-ni-extra` to fail to build as nothing RPROVIDE'd a dependency.
- Move `trace-cmd` to `packagegroup-ni-desirable`. That tool is an extremely common debugging tool for nilrt applications and failures to build `packagefeed-ni-extra` shouldn't make it inaccessible. We also want to be certain it is never dropped.

[AB#2490709](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2490709)

## Testing:
Ran `bitbake packagefeed-ni-extra` and confirmed the build at least started.

### DIFF
--- a/recipes-core/packagegroups/packagefeed-ni-extra.bb
+++ b/recipes-core/packagegroups/packagefeed-ni-extra.bb
@@ -237,11 +237,6 @@ RDEPENDS:${PN} += "\
 	tiobench \
 "
 
-
-# meta-openembedded/meta-oe/recipes-bsp
-RDEPENDS:${PN} += "\
-	tbtadm \
-"
 # meta-openembedded/meta-oe/recipes-connectivity
 RDEPENDS:${PN} += "\
 	gammu \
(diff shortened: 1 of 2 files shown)


recipes-core/packagegroups/packagefeed-ni-extra.bb
```diff
@@ -237,11 +236,6 @@ RDEPENDS:${PN} += "\
 	tiobench \
 "
 
-
-# meta-openembedded/meta-oe/recipes-bsp
-RDEPENDS:${PN} += "\
-	tbtadm \
```
chaitu236, line 243: Oops looks like I missed one. Thanks for fixing.

chaitu236 (approved):

recipes-core/packagegroups/packagefeed-ni-extra.bb
SparkingSpork: Yeah, honestly took me a minute to figure out what was going on. We somehow ended up with it in both packagegroup-ni-runmode and packagefeed-ni-extra? @amstewart - Maybe we need to do an audit of packages in the groups again.

amstewart (approved):

PR #610, opened 2023-08-22
kernel-test-nohz: re-order output so error reason is first

jpautler:
During our own internal testing, the first line of output from a failed ptest is used to indicate the reason for the failure. Making the nohz test consistent to other tests in this regard.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-test-nohz-files/nohz_test.c
+++ b/recipes-kernel/kernel-tests/kernel-test-nohz-files/nohz_test.c
@@ -41,15 +41,19 @@ struct histogram_data {
 	uint64_t *data;
 };
 
-static void error_exit(char* msg)
+static void error_exit(char* error, char *details)
 {
-	printf("error: %s\n", msg);
+	printf("error: %s\n", error);
+	if(details)
+		printf(details);
 	printf("FAIL: nohz_test\n");
 	exit(EXIT_FAILURE);
 }
 
-static void success_exit()
+static void success_exit(char* details)
 {
+	if(details)
+		printf(details);
 	printf("PASS: nohz_test\n");
 	exit(EXIT_SUCCESS);
 }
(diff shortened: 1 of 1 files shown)


chaitu236:
Is this ready for review?

recipes-kernel/kernel-tests/kernel-test-nohz-files/nohz_test.c
```diff
@@ -363,41 +367,43 @@ static void validate_results(struct histogram_data *h)
...
+		error_exit("Results validation failed; no histogram data", NULL);
 
 	p_99_999 = get_percentile(99.999, h);
 	p_99_9999 = get_percentile(99.9999, h);
 
-	printf("Total count: %lu (samples)\n", h->cnt);
-	printf("Maximum latency: %lu (ns)\n", h->max);
-	printf("99.999 percentile: %lu (ns)\n", p_99_999);
-	printf("99.9999 percentile: %lu (ns)\n", p_99_9999);
+	sprintf(buffer, stats_summary, h->cnt, h->max, p_99_999, p_99_9999);
```
mike-petersen-ni: Would it make sense to make `error_exit`/`success_exit` variadic and use [`vprintf`][1] so that `buffer` and `sprintf` are unnecessary here?

If not, at least be sure to avoid overruns:
```suggestion
	snprintf(buffer, sizeof(buffer), stats_summary, h->cnt, h->max, p_99_999, p_99_9999);
```

[1]: <https://linux.die.net/man/3/printf>

chaitu236 (approved):

recipes-kernel/kernel-tests/kernel-test-nohz-files/nohz_test.c
jpautler: Good idea. Updated.

mike-petersen-ni (approved):

chaitu236 (approved):

chaitu236:
Squashed and merged into https://github.com/ni/meta-nilrt/commit/adf0240e44fbd9711895a37b0e093b82d0db9197

PR #611, opened 2023-08-24
[kirkstone] busybox: Patch buffer size for ifplugd.

SparkingSpork:
Systems with a large number of ethernet ports are
becoming common with automotive ethernet applications. For systems like this, ethernet port configuration could fail due to a large number of messages in the notification queue for ifplugd causing a buffer to not have space.

This change applies a patch to busybox-ifplugd to use a higher socket buffer size to avoid the issue.

[AB#2498176](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2498176)

## Testing:
- Built in kirkstone
- Cherry-picked to hardknott, built, and tested on reproducing system

### DIFF
--- a/recipes-core/busybox/busybox/0001-ifplugd.c-Increase-buffer-size-for-netlink-binding.patch
+++ b/recipes-core/busybox/busybox/0001-ifplugd.c-Increase-buffer-size-for-netlink-binding.patch
@@ -1,0 +1,57 @@
+From 0d3331b91bbb7430fa6a13ebe631d0784983ee87 Mon Sep 17 00:00:00 2001
+From: Charlie Johnston <charlie.johnston@ni.com>
+Date: Thu, 24 Aug 2023 09:59:44 -0500
+Subject: [PATCH] ifplugd.c: Increase buffer size for netlink binding.
+
+Currently when binding to netlink to monitor for new
+adapter events, ifplugd uses the default buffer size.
+In cases with a large number of ethernet adapters, the
+buffer can overflow as any network adapter configuration
+change results in message. The buffer overflow can cause
+network adapters to fail initialization and not be put
+into an up state with the proper configuration.
+
+This change modifies the create_and_bind_to_netlink call
+to use a buffer size higher than the typical system default
+for x64-based systems.
+
+Signed-off-by: Charlie Johnston <charlie.johnston@ni.com>
+
+Upstream-Status: Inappropriate [Large numbers of adapters
+are uncommon on embedded systems]
+---
+ networking/ifplugd.c | 11 ++++++++++-
+ 1 file changed, 10 insertions(+), 1 deletion(-)
+
+diff --git a/networking/ifplugd.c b/networking/ifplugd.c
+index a776d4121..ffdc296e5 100644
+--- a/networking/ifplugd.c
++++ b/networking/ifplugd.c
+@@ -134,6 +134,15 @@ enum { // constant fds
+ 	netlink_fd = 4,
+ };
+ 
++enum {
++	/* The default buffersize is not sufficient when there are a large number
++	/ of network adapters (> 30). This can cause network adapters to not receive
++	/ IPv4 configurations from ifplugd. This value uses 128MiB instead of the
++	/ typical system default of 212,992 for 64-bit systems.
++	*/
++	KERN_RCVBUF = 128 * 1024 * 1024,
++};
++
+ struct globals {
+ 	smallint iface_last_status;
+ 	smallint iface_prev_status;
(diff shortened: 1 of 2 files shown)


recipes-core/busybox/busybox/0001-ifplugd.c-Increase-buffer-size-for-netlink-binding.patch
```diff
@@ -0,0 +1,57 @@
...
+ 	netlink_fd = 4,
+ };
+ 
++enum {
++	/* The default buffersize is not sufficient when there are a large number
++	/ of network adapters (> 30). This can cause network adapters to not receive
++	/ IPv4 configurations from ifplugd. This value uses 2x the typical system
++	/ default of 256k for 64-bit systems.
++	*/
++	KERN_RCVBUF = 128 * 1024 * 1024,
```
SparkingSpork, line 40: This value and approach was chosen based on other places in busybox where the same approach is used by a `create_and_bind_to_netlink` call. This is above the maximum allowed for user processes by the system, but `ifplugd` runs as admin so we can ignore the limit.

SparkingSpork: Also worth noting that the buffers are allocated lazily so it won't just grab all this memory in one swoop.

chaitu236: How many network interfaces can this buffer size support?

mike-petersen-ni: I'm not familiar with this code, but the 128M used here is *512x*, not 2x, of 256k. Perhaps the comment needs to be reworded to denote the units?

SparkingSpork: Ah, good catch. I fixed it in the commit message but not the comment.

SparkingSpork: That depends on what you mean. The buffer is created for each instance of `ifplugd` that is running. The buffer in question is receiving any messages related to network config changes that might be happening on the system, but it's not immediately clear to me how large those messages are or how many we might hit at most.

 The previous value was the default at `/proc/sys/net/core/rmem_default` which is 212,992 on x64 Linux. When I tested 3x this value, everything worked with the ~40 ports on the test system.

chaitu236: > The buffer is created for each instance of ifplugd that is running.. 

Oh so we overflow the default buffer size on systems with large no. of interfaces even though the buffers are unique for each instance?

SparkingSpork: Yes, because there's not a way to subscribe to messages only about the adapter in question. It subscribes to all adapters messages.

Honestly, this is kind of a drawback to using the busybox ifplugd that will hopefully go away if we move to systemd and its network management.

chaitu236: I wonder if the issue might bite us back if someone uses even more network interfaces and we find this buffer size still isn't sufficient.

SparkingSpork: Well, that is possible and it's why I set the value so high. Previously it was 212,992 and I tested with 638,976. That was enough for this system, but since the buffer is allocated lazily I set it to 134,217,728 similar to other places in the `busybox` repo. I imagine if that's not sufficient we'd see much worse problems than just this issue.

SparkingSpork: Err, when I say "I tested" I mean when I tested changing the default directly instead of this patch. I tested this patch with the values as-is.

SparkingSpork:
## V2
- Fixed comment in patch.

recipes-core/busybox/busybox/0001-ifplugd.c-Increase-buffer-size-for-netlink-binding.patch
SparkingSpork: Fixed.

chaitu236 (approved):

amstewart (approved):

PR #612, opened 2023-08-30
initscripts-nilrt: add LV RT cgroup version detection to init scripts

gratian:
The cgroup-v1 init scripts, currently used by LabVIEW RT, interfere with other uses of cgroups (ex. containers, systemd) and are run at boot regardless if LV RT is installed or not.

Selectively run the 'nicreatecpuacctgroups' and 'nicreatecpusets' init scripts only if cgroup-v1 is in use by LabVIEW RT. Currently cgroup-v1 is considered "in use" if LabVIEW RT is installed. This will change at a future date when LV RT adds cgroup-v2 support.

Additionally the new '/etc/default/lvrt-cgroup' configuration script can be used to force set the LVRT_CGROUP_VERSION to 0 and leave cgroups under OS control for testing.

Please note that the technically better alternative to make LabVIEW RT depend and directly install these scripts (instead of including them in the base image) was rejected due to the high cost of patching previous LabVIEW releases.

### DIFF
--- a/recipes-ni/initscripts-nilrt/files/lvrt-cgroup
+++ b/recipes-ni/initscripts-nilrt/files/lvrt-cgroup
@@ -1,0 +1,5 @@
+# This variable can be used to explicitly set the cgroup version used by LabVIEW
+# Real-Time cgroup related init scripts. A value of '0' leaves cgroups under OS
+# control.
+#
+# LVRT_CGROUP_VERSION=0
--- a/recipes-ni/initscripts-nilrt/files/nicreatecpuacctgroups
+++ b/recipes-ni/initscripts-nilrt/files/nicreatecpuacctgroups
@@ -1,6 +1,11 @@
 #!/bin/sh
-# Copyright (c) 2012-2013 National Instruments.
+# Copyright (c) 2012-2023 National Instruments.
 # All rights reserved.
+
+# Exit early if LabVIEW RT does not use cgroup-v1
+. /usr/share/initscripts-nilrt/lvrt-cgroup.sh
+identify_lvrt_cgroup_version || exit 0
+[ "$LVRT_CGROUP_VERSION" -eq 1 ] || exit 0
 
 #echo -n "Creating cgroups for CPU accounting for National Instruments LabVIEW Real-Time:"
 [ "${VERBOSE}" != "no" ] && echo -n "Starting nicreatecpuacctgroups:"
--- a/recipes-ni/initscripts-nilrt/initscripts-nilrt.bb
+++ b/recipes-ni/initscripts-nilrt/initscripts-nilrt.bb
@@ -12,6 +12,8 @@ SRC_URI = "\
 	file://cleanvarcache \
 	file://firewall \
 	file://iso3166-translation.txt \
+	file://lvrt-cgroup \
+	file://lvrt-cgroup.sh \
 	file://mountconfig \
 	file://mountdebugfs \
 	file://nicheckbiosconfig \
@@ -80,6 +82,11 @@ do_install () {
 	install -d ${D}${sysconfdir}/natinst
 	install -m 0644 ${WORKDIR}/iso3166-translation.txt ${D}${sysconfdir}/natinst
 
+	install -d ${D}${sysconfdir}/default
+	install -m 0644 lvrt-cgroup ${D}${sysconfdir}/default/lvrt-cgroup
+
+	install -d ${D}${datadir}/${BPN}
+	install -m 0755 lvrt-cgroup.sh ${D}${datadir}/${BPN}/lvrt-cgroup.sh
 }
 
 pkg_postinst_ontarget:${PN} () {
--- a/recipes-ni/initscripts-nilrt/files/lvrt-cgroup.sh
+++ b/recipes-ni/initscripts-nilrt/files/lvrt-cgroup.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+. /etc/default/lvrt-cgroup
+
+identify_lvrt_cgroup_version()
+{
+	# Return early if a version is already set
+	[ -z "$LVRT_CGROUP_VERSION" ] || return 0
+
+	# If lvrt is not installed set version to 0, cgroups not used
+	lvrt_installed=$(opkg list-installed ni-labview-realtime)
+	if [ -z "$lvrt_installed" ]; then
+		LVRT_CGROUP_VERSION=0
+		return 1
+	fi
+
+	# By default assume LabVIEW RT uses cgroups-v1 (current situation). This
+	# will be amended when LV cgroup-v2 upgrade is implemented.
+	LVRT_CGROUP_VERSION=1
+	return 0
+}
--- a/recipes-ni/initscripts-nilrt/files/nicreatecpusets
+++ b/recipes-ni/initscripts-nilrt/files/nicreatecpusets
@@ -1,6 +1,11 @@
 #!/bin/sh
-# Copyright (c) 2012-2013 National Instruments.
+# Copyright (c) 2012-2023 National Instruments.
 # All rights reserved.
+
+# Exit early if LabVIEW RT does not use cgroup-v1
+. /usr/share/initscripts-nilrt/lvrt-cgroup.sh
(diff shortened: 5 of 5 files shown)


gratian:
### Testing
- [x] tested the scripts on a NI Linux RT target (cRIO-9042)
- [x] full OE image build/test
  - [x] `bitbake nilrt-base-system-image`
  - [x] installed on a cRIO-9042 and verified that cgroups are not mounted if LV RT is not installed
  - [x] installed LV RT and verified cgroup v1 scripts run at boot and create the expected mount points/hierarchy

gratian:
@rtollert not sure why I can't add you to the reviewers list but I would like your input on this PR.

amstewart (approved):

chaitu236 (approved):

recipes-ni/initscripts-nilrt/files/lvrt-cgroup
```diff
@@ -0,0 +1,17 @@
+get_lvrt_cgroup_version()
+{
+	local lvrt_installed=$(opkg list-installed ni-labview-realtime)
```
rtollert: This is a bashism; sourcing it from `#!/bin/sh` makes this technically problematic. I would probably just fix `nicreatecpuacctgroups`'s hashbang tho.

rtollert: On a spare 9038 this line takes 1/8 s, which will probably dominate the execution time of all of this. It's right on the threshold where it arguably might matter.

I have a counter-offer, which AFAIK relies on behavior which is already supported, either *de facto* or *de jure*:
- Check for the existence of `/etc/natinst/labview.dir`.
  - If it does not exist, LabVIEW RT is not installed.
  - If it exists and its contents are equal to `/usr/local/natinst/labview`, LabVIEW RT is installed. (It's not possible for LabVIEW desktop to install to that directory, insofar as those are SxS installs, but LabVIEW RT isn't.)
  - If it exists but it has some other value... you probably needed cgroups-v2 anyway.

rtollert: ... hmm, but the rest of all of this is POSIX. So maybe just drop `local`.

recipes-ni/initscripts-nilrt/files/nicreatecpuacctgroups
```diff
@@ -1,7 +1,11 @@
...
-# Copyright (c) 2012-2013 National Instruments.
+# Copyright (c) 2012-2023 National Instruments.
 # All rights reserved.
 
+# Exit early if LabVIEW RT does not use cgroup-v1
+. /etc/default/lvrt-cgroup
+[ "$LVRT_CGROUP_VERSION" -eq 1 ] || exit 0
+
 #echo -n "Creating cgroups for CPU accounting for National Instruments LabVIEW Real-Time:"
 [ "${VERBOSE}" != "no" ] && echo -n "Starting nicreatecpuacctgroups:"
```
rtollert, line 11: Not technically part of this commit, but you should know about it: `/etc/default/rcS` can only be assumed to already be sourced if the initscript ends in `.sh` (which this does not). `VERBOSE` afaik isn't exported.

This distinction is probably going to come in handy when trying to debug all of this, particularly during customer deployments.

recipes-ni/initscripts-nilrt/files/lvrt-cgroup
```diff
@@ -0,0 +1,17 @@
...
+	# If lvrt is not installed return 0, cgroups not used
+	[ -z "$lvrt_installed" ] && echo 0
+
+	# By default assume LabVIEW RT uses cgroups-v1 (current situation).
+	# This will be amended when LV cgroup-v2 upgrade is implemented.
+	echo 1
+}
+
+# Determine cgroup version used by LabVIEW RT.
+LVRT_CGROUP_VERSION=$(get_lvrt_cgroup_version)
```
rtollert: Having nontrivial code execution happen at script source time bugs me. I would have preferred for the function to set `LVRT_CGROUP_VERSION` directly, and return error if lvrt isn't found, adjusting its name appropriately.... something like

```sh
LVRT_CGROUP_VERSION=unknown
identify_lvrt () {
    ... && return 1
    LVRT_CGROUP_VERSION=1
    return 0
}
```

```sh
. .../lvrt-cgroup
identify_lvrt || exit 0
[[ "$LVRT_CGROUP_VERSION" == 1 ] || exit 0
```

recipes-ni/initscripts-nilrt/initscripts-nilrt.bb
```diff
@@ -80,6 +81,8 @@ do_install () {
 	install -d ${D}${sysconfdir}/natinst
 	install -m 0644 ${WORKDIR}/iso3166-translation.txt ${D}${sysconfdir}/natinst
 
+	install -d ${D}${sysconfdir}/default/
+	install -m 0644 lvrt-cgroup ${D}${sysconfdir}/default/lvrt-cgroup
```
rtollert, line 86: Shell script includes don't belong in `/etc/default`. I'm pretty sure this needs to live in `/usr/share/initscripts-nilrt`.

rtollert:
All of my comments above are not showstoppers; ACK. Feel free to respond to the above comments either in this PR or a later one.

rtollert (approved):
Uh, ^.

recipes-ni/initscripts-nilrt/files/lvrt-cgroup
amstewart: I'm not really comfortable embedding that kind of domain knowledge into a script owned by RTOS. If the execution time is an issue, I might prefer to directly interrogate the cgroup subsystem to see if it has been captured by LVRT. But I'm not sure if that is technically feasible.

gratian: I think I agree with @amstewart here. Unfortunately there is no way to "interrogate the cgroup subsystem" because these are the scripts that capture cgroups for LVRT (and we're trying to skip them if LVRT is not installed)

Also, I've tried it on a 9030 and 9034 and what I'm seeing is closer to 100ms:
```
admin@NI-cRIO-9030-XXXXXXXX:~# time lvrt_installed=$(opkg list-installed ni-labview-realtime)

real    0m0.109s
user    0m0.109s
sys     0m0.000s
```

We will get to revisit this script when LVRT switches to cgroup-v2 so my proposal for now is to stick with the current approach.

recipes-ni/initscripts-nilrt/files/nicreatecpuacctgroups
gratian: It works as-is when I enable `VERBOSE`:

```
Starting nicreatecpusets:done
/etc/rc5.d/S01xserver-xfce: Disabled in system configuration.
Starting system message bus: dbus.
Starting nicreatecpuacctgroups:done
```

gratian:
### Changes since V1

- Removed 'local' variable designation from `$lvrt_installed`
- Moved `lvrt-cgroup.sh` script logic to `/usr/share/initscripts-nilrt/`
- Reworked the `identify_lvrt_cgroup_version` function per @rtollert suggestions

### V2 Testing

- [x] `bitbake initscripts-nilrt`
- [x] Installed resulting .ipk on a `cRIO-9043` with LVRT installed. Validated that the cgroup-v1 init scripts run correctly.
- [x] Installed resulting .ipk on a `cRIO-9043` without LVRT installed. Validated that cgroup-v1 init scripts are skipped.
- [x] Installed resulting .ipk on a `PXIe-8821` with LVRT 20.0.3 version installed (oldest selectable for this release). Validated that cgroup-v1 mount points are created correctly.
- [x] Validated that explicitly setting `LVRT_CGROUP_VERSION=0` in `/etc/default/lvrt-cgroup` results in skipping the cgroup-v1 init scripts on boot.

rtollert:
👍🏻

gratian:
@amstewart I've reset your vote as a reminder to review v2 and pull if it looks good to you.

amstewart (approved):
LGTM

amstewart:
Cherry-picked to `nilrt/23.8/kirkstone`.

PR #614, opened 2023-09-05
Create symlink to legacy status_led location

dylan-matthew-turner:
LabVIEW RT uses the older /usr/local/natinst/bin location for status_led while it now resides in /usr/bin.

Even if that gets corrected, we'll want to support earlier versions of LVRT, so the best course of action is to create a symlink to the new path at the old path.

We already [do this for niresetip](https://github.com/ni/meta-nilrt/blob/adf0240e44fbd9711895a37b0e093b82d0db9197/recipes-ni/ni-netcfgutil/ni-netcfgutil.bb#L31C16-L31C16) and other executables as well.

The change to status LED occurs in 22.5 and up, so this should be back ported to hardknott as well.

Internally tracked [here](https://dev.azure.com/ni/DevCentral/_workitems/edit/2501297/)

Tested by running `bitbake ni-utils`

### DIFF
--- a/recipes-ni/ni-utils/ni-utils.bb
+++ b/recipes-ni/ni-utils/ni-utils.bb
@@ -20,6 +20,7 @@ FILES:${PN} += "\
 	${sysconfdir}/init.d/nisetled \
 	${sysconfdir}/init.d/nisetprimarymac \
 	${sysconfdir}/natinst/networking/functions.common \
+	/usr/local/natinst/bin/status_led \
 "
 
 DEPENDS += "shadow-native pseudo-native niacctbase update-rc.d-native"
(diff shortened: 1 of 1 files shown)


amstewart (approved):

amstewart:
>The change to status LED occurs in 22.5 and up, so this should be back ported to hardknott as well.

The only way this fix would get into a customer's hardknott system would be if we went back and purposefully patched the published NILRT feeds for those releases. Is anyone asking that we do that?

chaitu236:
Can you add a link to the WI in description and mention what testing was done

chaitu236 (changes requested):
Can you add a link to the WI in description and mention what testing was done

dylan-matthew-turner:
Done

PR #627, opened 2023-10-10
Extend kernel-performance-tests to run in containers

dylan-matthew-turner:
## Reason

Customers want to use docker from nilrt, and we need to verify there is not a significant performance cost from doing so, so extend the existing kernel-performance-tests to run from containers to verify docker support is functional.

## Implementation

- Create two containers: one for running cyclictest and one to run in parallel with a load from hackbench, fio, or iperf.
- Duplicate the existing tests and modify each so the load command is called in the parallel container and the cyclictest is called in the main container

## Testing

Ran all the way through on a cRIO 9037 and uploaded results to Grafana.

NOTE: One small hiccup which is `fio` segfaults from a container. This will affect results as no load runs during the fio_containerized_test, but it doesn't stop the test or cause it to fail, so we can likely proceed for now and file a bug to fix the segfault.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/common.cfg
+++ b/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/common.cfg
@@ -1,0 +1,9 @@
+# Common configuration for all kernel performance tests
+TEST_DURATION=8h
+
+# Maximum expected latency in useconds.
+# The test will fail if cyclictest latency goes beyond this value.
+MAX_LATENCY=300
+
+# Location for benchmark logs and test results
+LOG_DIR="/var/local/ptest-results/kernel-containerized-performance-tests"
--- a/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/cyclictest-container/Dockerfile
+++ b/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/cyclictest-container/Dockerfile
@@ -1,0 +1,16 @@
+FROM ubuntu:latest
+
+# Install rt-tests, sudo, etc
+RUN DEBIAN_FRONTEND=noninteractive apt update && apt install -y \
+    rt-tests \
+    python3 \
+    python3-pip \
+    dmidecode \
+    grub-common
+
+# Need to send to InfluxDB
+RUN pip install influxdb
+
+# Script to call "run_cyclictest" from outside Docker container
+ADD --chmod=0775 call_run_ct.sh ./call_run_ct.sh
+
--- a/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/cyclictest-container/call_run_ct.sh
+++ b/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/cyclictest-container/call_run_ct.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+cd /ptests
+source ./run-cyclictest
+TEST_LOG=${LOG_DIR}/run_cyclictest-`date +'%Y_%m_%d-%H_%M_%S'`-${1}.log
+run_cyclictest "${1}" > ${TEST_LOG} 2>&1
+echo "${CYCLICTEST_RESULT}"
+
--- a/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/fio.cfg
+++ b/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/fio.cfg
@@ -1,0 +1,15 @@
+[global]
+directory=/var/cache/fio
+numjobs=1
+size=100M
+nrfiles=20
+openfiles=10
+direct=0
+verify=sha256
+do_verify=1
+time_based
+clocksource=clock_gettime
+bs=5k
+
+[random_rw]
+rw=randrw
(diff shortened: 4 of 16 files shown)


recipes-kernel/kernel-tests/kernel-performance-tests-files/cyclictest-container/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Script to call "run_cyclictest" from outside Docker container
+RUN echo '#!/bin/sh' > call_run_ct.sh \
+    && echo 'cd /ptests' >> call_run_ct.sh \
+    && echo 'source ./run-cyclictest' >> call_run_ct.sh \
+    && echo "run_cyclictest \"\${1}\" \\" >> call_run_ct.sh \
+    && echo " > /var/local/ptest-results/kernel-performance-tests/\${1}-run_cyclictest.log 2>&1" \
+        >> call_run_ct.sh \
+    && echo "echo \"\${CYCLICTEST_RESULT}\"" >> call_run_ct.sh \
+    && chmod +x call_run_ct.sh
```
amstewart: You can just put this logic into a `call_run_ct.sh` script in the docker context directory, and then install it like
```dockerfile
ADD --chmod=0775 call_run_ct.sh /call_run_ct.sh
```

recipes-kernel/kernel-tests/kernel-performance-tests-files/parallel-container/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+# Install rt-tests, sudo, etc
+RUN apt update
+RUN DEBIAN_FRONTEND=noninteractive apt install -y \
+    rt-tests \
+    fio \
+    iperf3 \
+    psmisc
+
+# Scripts to start loads from outside container
```
amstewart: Same with all these scripts.

recipes-kernel/kernel-tests/kernel-performance-tests-files/parallel-container/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM ubuntu:22.04
+
+# Install rt-tests, sudo, etc
+RUN apt update
+RUN DEBIAN_FRONTEND=noninteractive apt install -y \
```
amstewart: Docker caches the container at each step. So index updates and install operations are normally done as a single command. eg.
```dockerfile
RUN apt update && apt install -y ...
```

amstewart (changes requested):

dylan-matthew-turner:
Summary of changes:

- Move apt update into same command as apt install
- Extract docker script files to separate files instead of constructing via echo
- Adjust calling scripts to use new files

recipes-kernel/kernel-tests/kernel-performance-tests.bb
```diff
@@ -11,7 +11,7 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}-files:"
 S = "${WORKDIR}"
 
 DEPENDS = "virtual/kernel"
-RDEPENDS:${PN}-ptest += "bash rt-tests fio iperf3 python3 python3-pip"
+RDEPENDS:${PN}-ptest += "bash rt-tests fio iperf3 python3 python3-pip docker"
```
amstewart: The package name is `docker-ce`.

amstewart (changes requested):

amstewart (approved):

recipes-kernel/kernel-tests/kernel-performance-tests-files/cyclictest-container/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:22.04
```
gratian: Can we use a "latest" tag here instead of "22.04"? (re: https://hub.docker.com/_/ubuntu)
Trying to avoid having to manually rev this version in the future.

recipes-kernel/kernel-tests/kernel-performance-tests-files/parallel-container/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu:22.04
```
gratian: Same comment as above, is "latest" an option instead of a specific version?

recipes-kernel/kernel-tests/kernel-performance-tests.bb
```diff
@@ -27,6 +27,16 @@ SRC_URI += "\
     file://test_kernel_cyclictest_hackbench.sh \
     file://test_kernel_cyclictest_fio.sh \
     file://test_kernel_cyclictest_iperf.sh \
+    file://test_kernel_cyclictest_idle_containerized.sh \
```
gratian: Given these tests take a while to run (8 hours per test x 4 tests = 32 hours) can you split the containerized versions into a separate recipe? Maybe "kernel-performance-tests-containerized"  or something like that.
It will require separate ATS event/run but it will give us the option to run it in parallel (contingent on ATS HW availability etc.).

dylan-matthew-turner: Can do

dylan-matthew-turner:
Changes: moved to new test (forgot to update to ubuntu-latest). More changes coming

dylan-matthew-turner:
Changes:
- Moved to new test
- ubuntu:22.04 => ubuntu:latest

gratian (approved):

PR #630, opened 2023-10-25
Add restarting the docker daemon to kernel-containerized-performance-tests

dylan-matthew-turner:
## Reason

Currently, we cannot run the kernel-containerized-performance-tests on our ATS due to it not building the containers after reboot.

This likely a bug with the docker daemon, but until it's fixed, this can be resolved by doing a docker daemon restart.

Also, if the containers fail to build, the tests should fail.

## Implementation

- Add `/etc/init.d/docker.init restart` to the first test
- Add if clause to exit if the docker container doesn't exist after building

## Testing

- Installed on ATS target
- Restart and immediately run `ptest-runner`
- Verify it builds the container and proceeds to the test


amstewart (commented):
Could you split these changes into two commits, in case we want to revert one change or the other?

gratian (approved):

amstewart (approved):

PR #631, opened 2023-10-26
Add test to validate that docker functionality is supported

dylan-matthew-turner:
We want to support docker, so we need to ensure that things like the daemon, `--network host`, parallel containers, etc all work.

WI #2500898

This PR provides two tests:

- test_daemon which simply verifies that `docker run hello-world` runs without error
- test_parallel which sets up a background docker container with a file server and a second docker container which reads the file from that server

Tested by building the package and running ptest-runner on a 9053

### DIFF
--- a/recipes-ni/docker-functional-tests/docker-functional-tests.bb
+++ b/recipes-ni/docker-functional-tests/docker-functional-tests.bb
@@ -1,0 +1,35 @@
+SUMMARY = "Functional test of docker container capabilities on NILRT"
+HOMEPAGE = "https://github.com/ni/nilrt"
+SECTION = "tests"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+DEPENDS = ""
+
+SRC_URI = "\
+    file://run-ptest \
+    file://test_daemon.sh \
+    file://test_parallel.sh \
+    file://test_parallel.Dockerfile \
+    file://test_parallel_test_file.txt \
+"
+S = "${WORKDIR}"
+
+inherit ptest
+
+ALLOW_EMPTY:${PN} = "1"
+
+do_install_ptest:append() {
+    install -m 0755 ${S}/run-ptest ${D}${PTEST_PATH}
+    install -m 0755 ${S}/test_daemon.sh ${D}${PTEST_PATH}
+    install -m 0755 ${S}/test_parallel.sh ${D}${PTEST_PATH}
+    mkdir -p ${D}${PTEST_PATH}/test_parallel_container
+    install -m 0755 ${S}/test_parallel.Dockerfile \
+        ${D}${PTEST_PATH}/test_parallel_container/Dockerfile
+    install -m 0664 ${S}/test_parallel_test_file.txt \
+        ${D}${PTEST_PATH}/test_parallel_container/test_file.txt
+}
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+RDEPENDS:${PN}-ptest += "bash docker-ce"
+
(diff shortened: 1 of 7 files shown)


recipes-ni/docker-functional-tests/docker-functional-tests.bb
```diff
@@ -0,0 +1,37 @@
+SUMMARY = "Functional test of docker container capabilities on NILRT"
+HOMEPAGE = "https://docker.com"
+SECTION = "tests"
+LICENSE = "GPL-2.0-only & GPL-2.0-or-later"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/GPL-2.0-or-later;md5=fed54355545ffd980b814dab4a3b312c"
+
+inherit ptest
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
```
amstewart: The `FILESEXTRAPATHS` trick is only needed in `.bbappend` files. The recipe-local `files/` directory should already be in the PATHS for a `.bb` recipe.

recipes-ni/docker-functional-tests/docker-functional-tests.bb
```diff
@@ -0,0 +1,37 @@
+SUMMARY = "Functional test of docker container capabilities on NILRT"
+HOMEPAGE = "https://docker.com"
+SECTION = "tests"
+LICENSE = "GPL-2.0-only & GPL-2.0-or-later"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/GPL-2.0-or-later;md5=fed54355545ffd980b814dab4a3b312c"
+
+inherit ptest
```
amstewart: Try to follow the recipe layout [from the styleguide](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/docs/styleguide.md#layout). Class inherits should be pushed down toward the middle of the file.

recipes-ni/docker-functional-tests/docker-functional-tests.bb
```diff
@@ -0,0 +1,37 @@
...
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+S = "${WORKDIR}"
+
+DEPENDS = ""
+RDEPENDS:${PN}-ptest += "bash docker-ce"
+
+ALLOW_EMPTY:${PN} = "1"
+
+SRC_URI += "\
```
amstewart: MINOR: since this is a bb-recipe, we should prefer to use the regular `=` assignment, to keep anything screwy from happening with the `SRC_URI` variable.

recipes-ni/docker-functional-tests/docker-functional-tests.bb
```diff
@@ -0,0 +1,37 @@
+SUMMARY = "Functional test of docker container capabilities on NILRT"
+HOMEPAGE = "https://docker.com"
```
amstewart: For NI-authored recipes, we generally set the `HOMEPAGE` to `https://github.com/ni/nilrt`.

recipes-ni/docker-functional-tests/docker-functional-tests.bb
```diff
@@ -0,0 +1,37 @@
...
+SECTION = "tests"
+LICENSE = "GPL-2.0-only & GPL-2.0-or-later"
```
amstewart: Why not just MIT?

amstewart (changes requested):

recipes-ni/docker-functional-tests/docker-functional-tests.bb
dylan-matthew-turner: I copied over from another test that I used as reference. Is MIT our preference for new recipes?

chaitu236 (changes requested):
Can you add work item number to description

chaitu236 (approved):

dylan-matthew-turner:
Changes from previous commits:

- restructured to fit layout guide
- switched from GPL2 to MIT license
- remove unneeded changes to work with multiple recipes
- correct homepage

recipes-ni/docker-functional-tests/docker-functional-tests.bb
amstewart: I'd say so; yeah.

amstewart (approved):

PR #632, opened 2023-10-30
Fix packagegroup-ni-ptest-smoke typo

dylan-matthew-turner:
## Reason

In a [previous PR](https://github.com/ni/meta-nilrt/pull/631), we added a new test. That test was incorrectly added to packagegroup-ni-ptest-smoke and is causing errors in the PR build now.

## Implementation

- Change `docker-functional-ptest` to `docker-funcitonal-tests-ptest`

## Testing

Built `packagegroup-ni-ptest-smoke`

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-ptest-smoke.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-ptest-smoke.bb
@@ -16,7 +16,7 @@ RDEPENDS:${PN}:append = "\
     busybox-ptest \
     bzip2-ptest \
     coreutils-ptest \
-    docker-functional-ptest \
+    docker-functional-tests-ptest \
     e2fsprogs-ptest \
     elfutils-ptest \
     ethtool-ptest \


xhuff:
In https://github.com/ni/meta-nilrt/pull/631, in the testing section, you said you 'Tested by building the package' -- is that the same testing you're doing here? If so, what confidence do you have that this change will work if it wasn't caught last time?

dylan-matthew-turner:
So the problem is that last time, I built the packagegroup package, but decided to rename the docker testing package after to match the names of other recipes.

Then later, while I rebuilt the docker-functional-tests package anytime I made changes, I did not rebuild the packagegroup package, hence why I missed it.

To improve certainty, I can undo my change locally to verify it won't build as in the PR.

dylan-matthew-turner:
Yes, it succeeds with this change, but when I go back to master, it fails to build like in our PR builds:
image

So I am confident that this is the fix

chaitu236 (approved):

PR #636, opened 2023-11-09
kernel-test-nohz: add alternative to cgroups v1 and improve boot flags

gratian:
The current implementation assumes cgroups v1 mount points, and a cgroup cpuset directory layout configured for LabVIEW RT. It also does not fully take advantage of all possible 'isolcpus' flags.

Address these issues by providing a non-cgroup_v1 test configuration and adding the required boot flags.

AB #2540629

### Testing

- [x] `bitbake kernel-test-nohz-ptest`
- [x] `opkg install kernel-test-nohz-ptest` from local package feed on a NI Linux RT target
- [x] validated that the test starts correctly when run in a configuration where LabVIEW RT cgroup v1 layout is mounted
- [x] validated that the test starts correctly if no cgroups are mounted

Further performance validation left for the ATS.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-test-nohz.bb
+++ b/recipes-kernel/kernel-tests/kernel-test-nohz.bb
@@ -40,7 +40,7 @@ pkg_postinst_ontarget:${PN}-ptest:append() {
     ISOLATED_CPU=$((CPUS - 1))
 
     if [ $ISOLATED_CPU -gt 0 ]; then
-        echo 'set otherbootargs="${otherbootargs} isolcpus='$ISOLATED_CPU' nohz_full='$ISOLATED_CPU' mitigations=off intel_pstate=disable intel_idle.max_cstate=0 processor.max_cstate=0 nosoftlockup mce=ignore_ce audit=0 tsc=nowatchdog"' > /boot/runmode/no-hz-full-params.cfg
+        echo 'set otherbootargs="${otherbootargs} isolcpus=nohz,domain,managed_irq,'$ISOLATED_CPU' nohz_full='$ISOLATED_CPU' mitigations=off intel_pstate=disable intel_idle.max_cstate=0 processor.max_cstate=0 nosoftlockup mce=ignore_ce audit=0 tsc=nowatchdog"' > /boot/runmode/no-hz-full-params.cfg
         grep -qsxF 'source /runmode/no-hz-full-params.cfg' /boot/runmode/bootimage.cfg || echo 'source /runmode/no-hz-full-params.cfg' >> /boot/runmode/bootimage.cfg
     else
         echo "[kernel-test-nohz:error] This test requires a system with 2 or more CPUs"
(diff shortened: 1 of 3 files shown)


recipes-kernel/kernel-tests/kernel-test-nohz.bb
```diff
@@ -40,7 +40,7 @@ pkg_postinst_ontarget:${PN}-ptest:append() {
     ISOLATED_CPU=$((CPUS - 1))
 
     if [ $ISOLATED_CPU -gt 0 ]; then
-        echo 'set otherbootargs="${otherbootargs} isolcpus='$ISOLATED_CPU' nohz_full='$ISOLATED_CPU' mitigations=off intel_pstate=disable intel_idle.max_cstate=0 processor.max_cstate=0 nosoftlockup mce=ignore_ce audit=0 tsc=nowatchdog"' > /boot/runmode/no-hz-full-params.cfg
+        echo 'set otherbootargs="${otherbootargs} isolcpus=nohz,domain,managed_irq,'$ISOLATED_CPU' nohz_full='$ISOLATED_CPU' mitigations=off intel_pstate=disable intel_idle.max_cstate=0 processor.max_cstate=0 nosoftlockup mce=ignore_ce audit=0 tsc=nowatchdog"' > /boot/runmode/no-hz-full-params.cfg
```
chaitu236, line 43: Would we want to mention these params in [this readme](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/recipes-kernel/linux/nilrt-nohz/README.nohz#L44C55-L44C55)?
Also, for non-LV based applications [the instructions](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/recipes-kernel/linux/nilrt-nohz/README.nohz#L56) should be updated right?

chaitu236 (approved):

recipes-kernel/kernel-tests/kernel-test-nohz.bb
mike-petersen-ni, line 43: Not entirely relevant to this change, but it appears that `isolcpus` has been deprecated since ni/linux@b0d40d2b22fe.

gratian: Yeah, it has been marked deprecated and at some point it may get removed, however:

- It is still present in the latest Linus master (6.7-rc1 at this point)
- It has users that oppose it's removal: https://lore.kernel.org/linux-rt-users/CANV1gkdPJ=gqT2nUuGzN=ytOxHN92opoS1aGEDqy3FmTWTmusw@mail.gmail.com/
- We will notice when it gets removed (via this test)
- I would rather keep this test simple for now and not have to add support for the intermediate cgroup support state we're currently in until LV RT upgrades to v2

gratian:
Changes since V1:

- updated README.nohz per @chaitu236 suggestion

recipes-kernel/kernel-tests/kernel-test-nohz.bb
gratian: README.nohz updated.

chaitu236 (approved):

PR #640, opened 2023-11-29
dkms: autoinstall for all kernels by default

bstreiff:
If you have installed another kernel that would be the next one you boot into, but you're still on the old kernel, then if you install a dkms-enabled driver then by default the dkms common posting will only install for the "current" kernel and the "newest" (which in 3.0.10 only has logic for dpkg and rpm distros-- e.g., not nilrt).

On nilrt, if you do:
- opkg install packagegroup-ni-next-kernel
- opkg install ni-daqmx

then the dkms-enabled packages will run their postinsts, but only build for the currently-running kernel; this will _not_ be the kernel you enter on next boot, because we've installed the -next kernel.

dkms does have an option for building for all available kernels; turn this on for the nilrt distribution, as this better matches the user expectations for the way that kernels are treated on nilrt.

Natinst-AzDO-ID: [AB#2574521](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2574521)

### DIFF
--- a/recipes-kernel/dkms/dkms.bb
+++ b/recipes-kernel/dkms/dkms.bb
@@ -10,7 +10,9 @@ LIC_FILES_CHKSUM = "file://COPYING;md5=b234ee4d69f5fce4486a80fdaf4a4263"
 PV = "3.0.10"
 
 
-SRC_URI = "git://github.com/dell/dkms.git;protocol=https;branch=master \
+SRC_URI = "\
+	git://github.com/dell/dkms.git;protocol=https;branch=master \
+	file://0001-autoinstall-all-kernels.patch \
 "
 
 SRCREV = "3f72c9b059bccbc1ca1ff8b431c1763b1bdb60cd"
--- a/recipes-kernel/dkms/dkms/0001-autoinstall-all-kernels.patch
+++ b/recipes-kernel/dkms/dkms/0001-autoinstall-all-kernels.patch
@@ -1,0 +1,46 @@
+From f2923412125d8dfecc0be6dc8d5cab97ede6e2e1 Mon Sep 17 00:00:00 2001
+From: Brenda Streiff <brenda.streiff@ni.com>
+Date: Wed, 29 Nov 2023 11:14:27 -0600
+Subject: [PATCH] dkms: autoinstall for all kernels by default
+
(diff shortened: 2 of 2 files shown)


amstewart:
Can you link to where the dpkg/rpm logic you're referencing are? I'm wondering if it would make sense to upstream some opkg/NILRT logic.

bstreiff:
[dkms_common.postinst](https://github.com/dell/dkms/blob/3f72c9b059bccbc1ca1ff8b431c1763b1bdb60cd/dkms_common.postinst#L82) was where I found it; `NEWEST_KERNEL` will be an empty string if neither dpkg nor rpm are on the system.

That said I'm not sure how much "newest kernel" makes sense in nilirt, where you get the kernels out of a versioned feed, and because of the association of images to feeds you're not going to have old ones sitting around (as you might with a Debian or RHEL desktop install as you upgrade it from version to version.

It's also not clear to me how "newest kernel" works in a setup with variants; on Debian I can have `linux-image-$(uname -r)-686`, `-amd64`, `-cloud-amd64`, `-rt-686`, and `-rt-amd64` all installed and I'm not sure what dkms would do about that if I were to install them all in one go (similar to how I could install `-next` and `-nohz` on nilrt).

Instead of building for all kernels, the other thing that could be done is "run dkms autoinstall once you're in the new kernel". The [dkms_autoinstaller](https://github.com/dell/dkms/blob/v3.0.10/dkms_autoinstaller) script was at one time installed as an initscript to do that (it eventually got relegated to the `install-redhat-sysv` target [which is now gone](https://github.com/dell/dkms/commit/f3674f71d61b6537463b06b8c75101ec7c742907)?). There is a [systemd service](https://github.com/dell/dkms/blob/v3.0.10/dkms.service) to do `dkms autoinstall`, but it looks like it only ever gets onto the system in the [`install-redhat` codepath](https://github.com/dell/dkms/blob/3f72c9b059bccbc1ca1ff8b431c1763b1bdb60cd/Makefile#L59)? Maybe there's some distro-specific packaging that deals with this-- I haven't chased down Debian nor Red Hat's DKMS packages to see.

amstewart (approved):

PR #641, opened 2023-11-30
kernel-performance-tests: Add container load test

chaitu236:
Add a cyclictest ptest with container start/stop in background.

WI: [2500900](https://dev.azure.com/ni/DevCentral/_workitems/edit/2500900)

### Testing
- [x] `bitbake kernel-performance-tests`
- [x] Installed `kernel-performance-tests-ptest` on a cRIO-9043 and a VM
- [x] Ran `ptest-runner` with `TEST_DURATION` set to 8s. The new test passes. Docker load is cleaned up at the end even if test is aborted.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-performance-tests.bb
+++ b/recipes-kernel/kernel-tests/kernel-performance-tests.bb
@@ -11,7 +11,7 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}-files:"
 S = "${WORKDIR}"
 
 DEPENDS = "virtual/kernel"
-RDEPENDS:${PN}-ptest += "bash rt-tests fio iperf3 python3 python3-pip"
+RDEPENDS:${PN}-ptest += "bash rt-tests fio iperf3 python3 python3-pip docker"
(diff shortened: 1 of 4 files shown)


recipes-kernel/kernel-tests/kernel-performance-tests.bb
```diff
@@ -11,7 +11,7 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}-files:"
 S = "${WORKDIR}"
 
 DEPENDS = "virtual/kernel"
-RDEPENDS:${PN}-ptest += "bash rt-tests fio iperf3 python3 python3-pip"
+RDEPENDS:${PN}-ptest += "bash rt-tests fio iperf3 python3 python3-pip docker"
```
amstewart, line 14: The package name might be `docker-ce` here. Could you double-check that this is correct?

chaitu236: The package name is `docker-ce` but `docker-ce` provides `docker` so this works.

amstewart (approved):

PR #642, opened 2023-12-05
Kernel Containerized Performance Tests: Print run-cyclictest output

dylan-matthew-turner:
## Reason

The script `run-cyclictest` is run from within a container, and although the output is logged, it is never printed, meaning the tests do not print the PASS/FAIL messages required of ptests.

## Implementation

- After tests run, print the run-cyclictest log to stdout
- Remove date from log name (unneeded and makes it complicated to correctly cat the log)

## Testing

Ran through on ATS hardware and saw PASS messages appear

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/test_kernel_cyclictest_fio_containerized.sh
+++ b/recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/test_kernel_cyclictest_fio_containerized.sh
@@ -37,9 +37,11 @@ RESULT=$(docker run --privileged --network=host \
     bash call_run_ct.sh "fio_containerized" \
         | tr -d '\r' | tr -d '\n')
 
+# Make sure we print the PASS/FAIL message
+cat ${LOG_DIR}/run_cyclictest-fio_containerized.log
+
 # Clean up the background container
 docker exec ${LOAD_CONT} bash -c "killall -INT fio > /dev/null 2>&1"
 
-echo "Test Result: ${RESULT}"
 exit ${RESULT}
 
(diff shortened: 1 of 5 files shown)


recipes-kernel/kernel-tests/kernel-containerized-performance-tests-files/test_kernel_cyclictest_fio_containerized.sh
```diff
@@ -37,9 +37,11 @@ RESULT=$(docker run --privileged --network=host \
     bash call_run_ct.sh "fio_containerized" \
         | tr -d '\r' | tr -d '\n')
 
+# Make sure we print the PASS/FAIL message
+cat ${LOG_DIR}/run_cyclictest-`date +'%Y_%m_%d-%H_%M_%S'`-fio_containerized.log
```
mike-petersen-ni: You're recalculating the timestamp, so if you're unlucky with the clock, these files may not exist.

dylan-matthew-turner: I guess we don't really need the timestamp anyway since the target is wiped every run

dylan-matthew-turner:
Changes:

- Removed timestamp from log name since target is rewritten anyway

dylan-matthew-turner:
Changes:

- Correct signed-off email

chaitu236 (changes requested):
Please describe the testing you did in description. And add work item number.

dylan-matthew-turner:
I hadn't done testing bc it is a small change, but I will set something up and get back

dylan-matthew-turner:
@chaitu236 Added

chaitu236 (approved):

PR #647, opened 2023-12-11
Allow choice of source for fw_printenv/fw_setenv

mike-petersen-ni:
As noted in [AB#2546902](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2546902), fw_printenv will print the "same" variable multiple times if it is found in multiple places (for instance, both SMBIOS and GRUB). Since we don't know which table should take priority in all cases, we can offer a choice to the user with a command-line argument.

## Testing
Used a VM with doubly-defined variables from SMBIOS and GRUB, as described in the bug. I haven't tested examining/manipulating EFI variables.

 - The default behavior of `fw_printenv` has not changed: doubly-defined variables are still printed twice without indicating their origin.
 - The default behavior of `fw_printenv variable` has not changed: doubly-defined variables have one of them chosen "arbitrarily" (really, it's just the first successful lookup, as before).
 - `fw_printenv -t table` prints all variables from the indicated table.
 - `fw_printenv -t table variable` prints the indicated variable from the indicated table.
 - The behavior of `fw_setenv` without the `-t` argument has not changed.
 - `fw_setenv -t table variable value` behaves appropriately.

### DIFF
--- a/recipes-ni/fw-printenv/files/fw_printenv
+++ b/recipes-ni/fw-printenv/files/fw_printenv
@@ -9,10 +9,10 @@
 print_help(){
 cat <<EOF
 fw_printenv/fw_setenv, an implementation of the U-Boot environment variable facility in Grub/EFI
-usage:  fw_printenv [-n] [variable name]
-	fw_setenv [variable name] [variable value]
-	fw_setenv -s [ file ]
-	fw_setenv -s - < [ file ]
+usage:  fw_printenv [-t table name] [-n] [variable name]
+        fw_setenv [-t table name] [variable name] [variable value]
+        fw_setenv [-t table name] -s [ file ]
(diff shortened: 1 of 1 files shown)


amstewart:
I don't think I like the idea of continuing to return multiple (or arbitrary) values unless the user specifies otherwise. I think we're in agreement about that behavior being a bug, and I don't see a reason to preserve it.z

What about emulating [how `git config` works](https://git-scm.com/docs/git-config)? We can document a priority for the tables, and have `fw_printenv` return the latest (most specific) table's result by default. Users can specify a particular scope, if they are interested in one table over the others. That would be pretty similar to what you have here.

amstewart:
Dropped in favor of #657

PR #656, opened 2024-01-12
kernel-tests: Update debugging options

jpautler:
Retrieving current dmesg log from database by date is more convenient. Add additional options to aide debugging.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
+++ b/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
@@ -1,5 +1,4 @@
 import argparse
-import copy
 import datetime
 import difflib
 import hashlib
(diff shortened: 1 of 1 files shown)


recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
```diff
@@ -232,49 +242,38 @@ def parse_args():
...
+    kernel_version = KernelVersion()
+    os_version = OsVersion()
+    device_desc = get_device_desc()
+    previous_dmesg_record = get_previous_dmesg_record(db, kernel_version, os_version.major_minor, device_desc, logger)
+
+previous_dmesg_log = strip_headers(previous_dmesg_record['dmesg_log']) if previous_dmesg_record else ''
 
-if not args.get_current_log_from_db:
-    upload_log(db, dmesg_log, logger)
+if not args.current_log_db_date and not args.skip_upload:
```
chaitu236, line 274: nitpick: it appears that logs aren't uploaded not only when `skip_upload` is set but also when `current_log_db_date` is set. Maybe mention this in `current_log_db_date`'s help?

chaitu236 (approved):

recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
jpautler: Added a note to this effect and made a couple other small tweaks to the command line help.

PR #657, opened 2024-01-12
fw_printenv: Refactor variable printing and avoid duplicates

jeminor:
The existing command line usage of fw_printenv \<variable\> will correctly pick an order and return only a single value from BIOS, EFI or GRUB.  However, when fw_printenv is called without any \<variable\> argument, all values are printed from all tables.  This can lead to duplicates being printed, and doesn't match the behavior of the single-variable case.

This is also an opportunity to clean up all of the conditionals in each of the show functions so that only a common code path is used to print the variables in the "show all" and "show single variable" cases.

[AB#2546902](https://ni.visualstudio.com/DevCentral/_workitems/edit/2546902)

### Testing

- [X] Tested on a VM emulating BIOS and GRUB-sourced variables.
- [X] Tested on a cRIO-9030.

### DIFF
--- a/recipes-ni/fw-printenv/files/fw_printenv
+++ b/recipes-ni/fw-printenv/files/fw_printenv
@@ -23,6 +23,9 @@ File syntax example:
 In this case, var1 will be deleted, var2 and other_value
 will be set with the values passed. The value can contain any
 number of spaces.
+
+Variables defined in multiple places will have the
+following preference order: BIOS -> EFI -> GRUB
 EOF
 }
 
@@ -30,32 +33,23 @@ EOF
 #they will write the value of the "$1" variable name
 
 show_variable_grub(){
-	if [ $# -eq 0 ] ; then
-		#show all variables
-		grub-editenv list
-		if [ $? -ne 0 ]; then
-			return 1
+	set -o pipefail
+	value=$(grub-editenv - list | grep -- "$1=" | while read -r line; do
+		key=$(echo "$line" | cut -d "=" -f 1)
+		#if the $1 matches exactly the key, extract the value
+		if [ "$key" = "$1" ] ; then
+			value_tmp=$(echo "$line" | cut -d "=" -f 2-)
+			echo "$value_tmp"
+			break
 		fi
-		return 0
-	else
-		set -o pipefail
-		value=$(grub-editenv - list | grep -- "$1=" | while read -r line; do
-			key=$(echo "$line" | cut -d "=" -f 1)
-			#if the $1 matches exactly the key, extract the value
-			if [ "$key" = "$1" ] ; then
-				value_tmp=$(echo "$line" | cut -d "=" -f 2-)
-				echo "$value_tmp"
-				break
-			fi
-		done)
-		if [ $? -ne 0 ]; then
-			return 1
-		fi
-		if [[ $value ]]; then
-			return 0
-		fi
+	done)
+	if [ $? -ne 0 ]; then
 		return 1
 	fi
+	if [[ $value ]]; then
+		return 0
+	fi
+	return 1
 }
 
 readonly_grub_variable()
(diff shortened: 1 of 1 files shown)


xhuff:
Please link the AzDO work item in the PR description using the form AB#XXXXXXX

recipes-ni/fw-printenv/files/fw_printenv
```diff
@@ -23,41 +23,46 @@ File syntax example:
 In this case, var1 will be deleted, var2 and other_value
 will be set with the values passed. The value can contain any
 number of spaces.
+
+Variables defined in multiple places will have the
+following preference order: BIOS -> EFI -> GRUB
```
amstewart, line 28: This seems like the inverse ordering from what I would expect. EFI is a more recent and featureful technology than BIOS, and I would expect it to be preferred. And the GRUB install has a more-specific scope than BIOS/EFI, so I would expect it to be preferred over all.

I understand there was a previous behavior on ARM to lock the uboot variables and consider them canonical. But I don't see the reason that we would carry that behavior to x64.

Can we reverse the ordering?

amstewart (commented):
Does this implementation have a way for a user to request a variable value from a particular store?

I imagine we may have users who are unwittingly depending on the current behavior. And while I'm fine changing it on them to something more correct, it would be nice to have an easy pathway for them to keep using their old behaviors. Something like `fw_printenv --efi $var`.

jeminor:
> Does this implementation have a way for a user to request a variable value from a particular store?
> 
> I imagine we may have users who are unwittingly depending on the current behavior. And while I'm fine changing it on them to something more correct, it would be nice to have an easy pathway for them to keep using their old behaviors. Something like `fw_printenv --efi $var`.

To be clear, I'm not changing the behavior for the single variable case in this commit at all.  It already worked that way with that priority preference.  It was only the "print all variables" case that I'm changing to accurately reflect what will happen in the single variable case.

That's what I meant by this at the top of the PR:
> The existing command line usage of fw_printenv will correctly pick an order and return only a single value from the BIOS, EFI or GRUB.

Given that the most used instances of fw_printenv are from things like scripts and code, I don't know if it's worth having an override on the command line as to which store to use?  I'm happy to add such a feature if we feel like it could be used, but we might instead need some kind of "global switch" to avoid changing all the individual places maybe?

amstewart:
> > Does this implementation have a way for a user to request a variable value from a particular store?
> > I imagine we may have users who are unwittingly depending on the current behavior. And while I'm fine changing it on them to something more correct, it would be nice to have an easy pathway for them to keep using their old behaviors. Something like `fw_printenv --efi $var`.
> 
> To be clear, I'm not changing the behavior for the single variable case in this commit at all. It already worked that way with that priority preference. It was only the "print all variables" case that I'm changing to accurately reflect what will happen in the single variable case.

Yeah, I was making that comment on the assumption that we would reverse the ordering for the single-variable and all-variable get-operations - which would be a breaking change.

recipes-ni/fw-printenv/files/fw_printenv
jeminor: During a team discussion we decided this might inadvertently change behavior client teams were depending on.  I therefore think we should keep the order as it exists until we make a larger change to source these variables in a direct way from their data sources.

Marking resolved, feel free to reopen if you want to reconvene the larger conversation offline.

recipes-ni/fw-printenv/files/fw_printenv
```diff
@@ -23,41 +23,44 @@ File syntax example:
...
+		return 0
+	fi
+	return 1
 }
 
+get_available_variables_grub() {
+	all_variables+=$(grub-editenv - list | while read -r line; do
+		key=$(echo "$line" | cut -d "=" -f 1)
+		echo $key
+	done)
```
amstewart: What about...
```bash
all_variables+=$(grub-editenv - list | awk -F= '{print $2}')
```

recipes-ni/fw-printenv/files/fw_printenv
```diff
@@ -23,41 +23,44 @@ File syntax example:
...
+	set -o pipefail
+	value=$(grub-editenv - list | grep -- "$1=" | while read -r line; do
+		key=$(echo "$line" | cut -d "=" -f 1)
+		#if the $1 matches exactly the key, extract the value
+		if [ "$key" = "$1" ] ; then
+			value_tmp=$(echo "$line" | cut -d "=" -f 2-)
+			echo "$value_tmp"
+			break
 		fi
+	done)
```
amstewart: What about?
```bash
value=$(grub-editenv - list | awk -F= -v key="$key" '{ if ($1 == $key) { print $2; exit }}')
```

amstewart (changes requested):

recipes-ni/fw-printenv/files/fw_printenv
jeminor: I think you meant $1 here, but point taken :)

recipes-ni/fw-printenv/files/fw_printenv
jeminor: I had to tweak it a little bit to make it work (no "$" in var names in the awk script), but that seems cleaner.

amstewart (approved):

PR #660, opened 2024-01-22
kernel-tests: add replacement pattern for filesystem UUIDs

jpautler:
I noticed that kernel_dmesg_diff.py was returning different hashes for each run against the 6.6 kernel. Many of those diffs would have identical hashes except that some filesystem UUIDs in the dmesg log are changing on each run. I updated the kernel_dmesg_diff.py test with an additional regex pattern so they can be grouped going forward.

[AB#2600436](https://dev.azure.com/ni/DevCentral/_workitems/edit/2600436)

### Testing
Ran kernel_dmesg_diff.py manually with this change and supplied the date of an affected dmesg log with --current_log_db_date. The output confirmed the changing UUIDs are replaced with a constant string.


chaitu236 (approved):
Looks good. But can you add description, WI# (if it exists) and Testing done in description.

amstewart (approved):

PR #665, opened 2024-01-26
Allow fs manifest test to ignore version changes and make debugging easier

jeminor:
The existing version of this test notifies us of every version change.  Given how much we're changing these days, suppress the version changes since those are reviewed separately in their respective PRs.

While at it, I also added some command-line option similar to the changes that were just made to the dmesg test.  These could be used in future to reset the manifest basis, if we deem that to be useful.

[AB#2507029](https://ni.visualstudio.com/DevCentral/_workitems/edit/2507029)

### Testing

- [X] Tested with a set of failures we have already seen.  Diff shows only non-version differences.

### DIFF
--- a/recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
+++ b/recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
@@ -104,7 +104,32 @@ def upload_manifest(db, fs_manifest, logger):
 def strip_headers(fs_manifest):
     return ''.join(line + '\n' for line in fs_manifest.splitlines() if not line.startswith('#'))
 
-def get_old_fs_manifests(db, logger):
+def get_previous_fs_manifest_as_current(db, previous_date, logger):
+    query = {}
+    query['date'] = previous_date
+    count = db.count_documents(query)
+    logger.log('!!!!!OVERRIDING CURRENT MANIFEST!!!!!\n'.format(count))
+    logger.log('Found {} records\n'.format(count))
+    if count == 1:
+        results = db.find(query).limit(1)
+        result = next(results)
+        logger.log('INFO: Found fs_manifest record with these details')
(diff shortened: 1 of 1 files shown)


amstewart (approved):

chaitu236 (approved):

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -319,17 +374,33 @@ def parse_args():
...
+                             'test or can be extracted from the Mongo database using a viewer tool like Compass. This flag also skips the upload of the filesystem manifest record to the '\
+                             'database, so the --skip_upload flag is not needed. This flag is useful for debugging issues with this test.')
+    parser.add_argument('--basis_log_db_date', metavar="<date>",
+                        help='Use this flag to supply a date string that will be used to locate a filesystem manifest in the database. That log will be used as the basis filesystem manifest. '\
+                             'Should be of the format "2023-03-30 15:32:43.203476". Date strings for previous filesystem manifest can be found in the output of previous runs of this '\
+                             'test or can be extracted from the Mongo database using a viewer tool like Compass. This flag is useful for resetting the comparison baseline for the test.')
+    parser.add_argument('--skip_upload', help='Skip upload of dmesg record to database. Useful when debugging.', action="store_true")
```
jpautler: Change "dmesg records" to "filesystem manifest".

jpautler (changes requested):
LGTM. One small thing to fix.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
jeminor: Done thx for noticing the copy/pasta :)

jpautler (approved):

PR #668, opened 2024-01-31
fs_permissions_diff: Run logrotate before running the test

jeminor:
logrotate runs via cron every 5 mins, which means that sometimes logrotate.status will be there and sometimes it won't during runs.  We do want to check that it's permissions are correct, so manually poke it the same way that cron will do it before running the test.

### Testing

None (aside from running the command on a target), but it's difficult to reproduce the condition of the DUT outside the ATS at the moment of testing.  Once this is complete the results will either start passing or continue failing and need a bit of further tweaking.


jeminor (commented):
Offline request to move this logic into the python, stand by...

jpautler (approved):
LGTM

chaitu236 (approved):

PR #675, opened 2024-02-22
initrdscripts: remove efifix script

jpautler:
This unused script was originally added for a new RAUC/NXG bootflow which is no longer being pursued. Removing now as it contains calls to egrep which are being removed due to obsolescence in the context of the recovery tool.

[AB#2655026](https://dev.azure.com/ni/DevCentral/_workitems/edit/2655026)

### Testing
Built nilrt-recovery-media, put iso on a USB key, provisioned a target, exited to recovery shell, rebooted, and confirmed target seemed to provision correctly. Confirmed efifix script was no longer present and that absence did not cause any issues.


amstewart (approved):

amstewart:
~~Oh; wait. This PR should be targeted against the `nilrt/master/kirkstone` ref, in order to hit the 23Q2 release. I'll cherry-pick it to the `/next` mainline when I merge it.~~

Wait; scratch that. This bug is only against the scarthgap release. so this makes sense.

amstewart (approved):

PR #678, opened 2024-03-01
kernel_tests: Add replacement pattern

chaitu236:
Add replacement pattern for clocksource

WI: [AB#2654791](https://dev.azure.com/ni/DevCentral/_workitems/edit/2654791)

### Testing
- [x] Verified that regex pattern matches dmesg logs.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
+++ b/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
@@ -217,7 +217,8 @@ replacement_patterns = [
         [r'software IO TLB: mapped mem .*', 'software IO TLB: mapped mem'],
         [r'eth\d:', 'ethX:'],
         [r'renamed from eth\d', 'renamed from ethX'],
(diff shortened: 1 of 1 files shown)


jpautler:
Do you really need the patterns for MMCONFIG or pcieport?

It looks like the address range reported for MMCONFIG is consistent from run to run and failures are already being grouped together with the same hash.

For the pcieport lines, it also looks like the numeric values are consistent from run to run and failures are already being grouped together with the same hash. Also, these lines seem to have disappeared from the kernel output (rather than being added). So even if the values were changing, they won't be an issue in future runs since those lines of output no longer exist.

The remaining pattern for clocksource seems correct and necessary.

chaitu236:
> Do you really need the patterns for MMCONFIG or pcieport?
> 
> It looks like the address range reported for MMCONFIG is consistent from run to run and failures are already being grouped together with the same hash.
> 
> For the pcieport lines, it also looks like the numeric values are consistent from run to run and failures are already being grouped together with the same hash. Also, these lines seem to have disappeared from the kernel output (rather than being added). So even if the values were changing, they won't be an issue in future runs since those lines of output no longer exist.
> 
> The remaining pattern for clocksource seems correct and necessary.

I hadn't realized patterns were consistent for MMCONFIG. Removed
Likewise removed pcieport patterns.

amstewart (approved):

jpautler (approved):

gratian (approved):

PR #679, opened 2024-03-11
README.md: Remove branch refs

jeminor:
The fido branch references were still in the README, but that was old quite a few revisions ago.

Fix it by just removing them so as to not require updates in the future, as the nilrt repo should dictate such things.

No testing since this is just a trivial documentation change.


gratian (approved):

README.md
```diff
@@ -6,43 +6,25 @@ This layer provides support for the NI Linux Real-Time distribution. The NI Linu
 This layer depends on:
```
amstewart: Could you just remove this whole section. The layer URIs aren't current either.

amstewart (changes requested):

README.md
jeminor: Done.

amstewart (approved):

PR #684, opened 2024-03-27
fs_permissions_diff: Add command-line option to specify recent manifest

jpautler:
When reproducing test failures or debugging the test itself, it can be useful to run the test from a non-NILRT developer machine. This is not currently possible because there is no way to specify the recent manifest. Therefore the logic to determine a recent manifest must run, which queries the system to get NILRT OS version information. This fails when running on a non-NILRT machine. This change includes a new command-line option to specify a recent manifest.

### Testing

Ran fs_permissions_diff.py manually with this change both with and without manifest options on both NILRT and non-NILRT systems. Confirmed that the behavior looked normal in all cases. Confirmed test now runs on non-NILRT systems when all manifests are specified using command-line options.


jpautler:
Closing pull request. I found some related issues I'd like to think through before possibly commiting this.

PR #690, opened 2024-04-19
ni-systemimage: Add support for getting full disk image

chaitu236:
- ni-systemimage: Remove unnecessary dependencies
- Add support for getting full disk image i.e., including nigrub and safemode files

WI: [AB#2371308](https://dev.azure.com/ni/DevCentral/_workitems/edit/2371308)

### Testing
- [x] `bitbake ni-systemimage && bitbake package-index && bitbake nilrt-recovery-media`
- [x] Tested on a VM: `nisystemimage getall -x tgz -f test.tgz` works from both safemode and restoremode; the resulting image contains nigrub, safemode files, runmode files, config files, rootfs.
- [x] Existing `get` functionality continues to work.
- [x] With a rudimentary `setall` implementation, verified that resulting image works

### Note
This is a basic implementation to unblock other WIs; future PRs will address maintainability, etc.

### DIFF
--- a/recipes-ni/ni-systemimage/files/nisystemimage
+++ b/recipes-ni/ni-systemimage/files/nisystemimage
@@ -5,8 +5,6 @@
 set -eEu -o pipefail
 shopt -s dotglob nullglob
 
-. /etc/natinst/networking/functions.common
-
 ##
 ## Globals
 ##
--- a/recipes-ni/ni-systemimage/files/nisystemimage
+++ b/recipes-ni/ni-systemimage/files/nisystemimage
@@ -49,6 +49,13 @@ BLACKLIST_LISTS=()
 BLACKLIST_PATHS=(
 	# syslog-ng socket (sockets can't be tar'd)
 	/var/syslog-ng.ctl
+	/var/volatile
+	# Systemlink Configuration Files
+	/etc/salt/minion_id
+	/etc/salt/pki
+)
+
+BLACKLIST_PATHS_get=(
 	# Safe mode
 	/boot/.safe
 	# Grub variables
@@ -57,13 +64,13 @@ BLACKLIST_PATHS=(
 	/boot/.defbit
 	# User bitfile
 	/boot/user.bit*
-	/var/volatile
 	# Restore-mode fail-safe files
 	/.restore
-	# Systemlink Configuration Files
-	/etc/salt/minion_id
-	/etc/salt/pki
 )
+
+if [[ "${MODE}" == "get" ]]; then
+	BLACKLIST_PATHS+=(${BLACKLIST_PATHS_get[*]})
+fi
 
 # All excludes to pass to `tar`. Includes BLACKLIST_PATHS and contents of all
 # files in BLACKLIST_LISTS. The contents of EXCLUDES are paths relative to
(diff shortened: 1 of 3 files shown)


recipes-ni/ni-systemimage/files/nisystemimage
```diff
@@ -5,8 +5,6 @@
 set -eEu -o pipefail
 shopt -s dotglob nullglob
 
-. /etc/natinst/networking/functions.common
```
amstewart, line 8: Why is this unnecessary? Is it unused?

recipes-ni/ni-systemimage/ni-systemimage.bb
```diff
@@ -31,6 +31,4 @@ RDEPENDS:${PN} += "\
 	bash \
 	findutils \
 	ni-netcfgutil \
-	ni-utils \
-	niacctbase \
```
amstewart, line 35: Is your thinking here that the `LVRT_GROUP` gid is already encoded into the `ni-systemimage:do_install` by the `DEPENDS`? I think we still need this `RDEPENDS` to ensure that the GROUPADD for lvrt is completed by the package installation.

recipes-ni/ni-systemimage/files/nisystemimage
```diff
@@ -784,9 +843,11 @@ trap 'handle_err ${BASH_SOURCE} ${LINENO} ${FUNCNAME:-unknown} $?' ERR
...
 [[ -f $SCRIPT_PRE ]] && . "$SCRIPT_PRE"
 
-case "$(runlevel)" in
-*\ 6|*\ 0) die ILLTIMED "Called during shutdown; aborting." ;;
-esac
+if ! in_restoremode; then
+	case "$(runlevel)" in
+	*\ 6|*\ 0) die ILLTIMED "Called during shutdown; aborting." ;;
+	esac
+fi
```
amstewart, line 850: Why would we not want to do this runlevel check in restoremode?

recipes-ni/ni-systemimage/files/nisystemimage
```diff
@@ -366,6 +384,15 @@ init_fs_get () {
 	fi
 }
 
+init_fs_getall () {
```
amstewart, line 387: Is this function, or `init_fd_getall()`, called anywhere?

amstewart (changes requested):

recipes-ni/ni-systemimage/files/nisystemimage
chaitu236: Yes, I didn't see any of the symbols from `/etc/natinst/networking/functions.common` being used here.

recipes-ni/ni-systemimage/ni-systemimage.bb
chaitu236: I made a mistake here. My intention was to remove dependency on niacctbase altogether so that this script can be used from restoremode without installing niacctbase and its dependencies. But I hadn't noticed that we use `LVRT_GROUP` in this recipe or that there is a DEPENDS on `niacctbase`.

Let me think about how to do this again.

recipes-ni/ni-systemimage/files/nisystemimage
chaitu236: We use our own `/init` in restoremode and not the one from sysvinit so `runlevel` returns `unknown`.

recipes-ni/ni-systemimage/files/nisystemimage
chaitu236: Yes, `init_fd` calls `init_fd_$MODE` which for `getall` would be `init_fd_getall`.

recipes-ni/ni-systemimage/ni-systemimage.bb
chaitu236: Turns out that `niacctbase` already exists in restoremode through some other dependency chain, so keeping it in this recipe isn't going to change anything. Added it back.

amstewart (approved):

PR #694, opened 2024-04-24
USB replication: Add setall functionality and misc

chaitu236:
This patchset introduces following changes

- Fixes for issues in https://github.com/ni/meta-nilrt/pull/690
- Add setall functionality to ni-systmeimage
- Split ni_provisioning.safemode so common code can be reused for applying full system image.
- Add a rudimentary nisystemreplication script (and its recipe) that can apply full system image from restore mode (with some modifications to mount required volume to get the image from, fixing paths).

WI: [AB#2372259](https://dev.azure.com/ni/DevCentral/_workitems/edit/2372259)

### Testing
 - [x] Built restoremode.
 - [x] Regular ISO provisioning workflow works.
 - [x] nisystemimage's `get`, `getall`, `set`, `setall` works from safemode (*setall only works when image used was from the same system as nisystemimage doesn't set rootuuid; this is expected*).
 - [x] nisystemreplication's `get`, `set` works from restoremode. (*setall works even with images from different system*).
 
# Note
This is a barebones implementation for `setall`. Things to be addressed in future PRs:
- Expand list of files to be excluded
- Correctly handle network settings' replication
- Handle FW variables


amstewart (approved):
The intention is to update these scripts with better error reporting and help documentation once this initial implementation is in; correct?

recipes-ni/ni-systemreplication/ni-systemreplication.bb
```diff
@@ -0,0 +1,25 @@
...
+
+SRC_URI = "\
+    file://nisystemreplication \
+"
+
+S = "${WORKDIR}"
+
+do_install () {
+    install -d ${D}${bindir}
+    install -m 0550 ${S}/nisystemreplication ${D}${bindir}
```
amstewart: Oh. Actually: let's put this in `/sbin`. It doesn't make sense for anyone but root to run this tool; right?

amstewart (changes requested):

chaitu236:
> The intention is to update these scripts with better error reporting and help documentation once this initial implementation is in; correct?

Yes, that's correct.

recipes-ni/ni-systemreplication/ni-systemreplication.bb
chaitu236: Done. Yes, only root is supposed to run.

amstewart (approved):

PR #699, opened 2024-05-09
[next] salt: fix basic compatibility with python_3.12 (scarthgap)

amstewart:
Salt 3000.2 is broken in nilrt/master/next due to incompatibilities between salt and python_3.12. This patchset:
1. Adds a new python recipe for `python3-backports-ssl-match-hostname`. A *forward-port* of the deprecated `ssl.match_hostname` module, that salt depends upon. This recipe won't be appropriate to upstream to OE-core.
2. Adds an `RDEPENDS` on the `python3-charset-normalizer` module, which salt requires. It was previously being implicitly included via another package.

[AB#2719441](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2719441)

# Testing
* [x] With these changes `salt-call --local grains.items` completes successfully (with unrelated warnings).
* [x] Rebuilt the full packagefeed and BSI with these changes and validated that salt now basically works from-installation.

### DIFF
--- a/recipes-core/packagegroups/packagefeed-ni-core.bb
+++ b/recipes-core/packagegroups/packagefeed-ni-core.bb
@@ -16,6 +16,7 @@ RDEPENDS:${PN} = "\
 	packagegroup-ni-restoremode \
 	packagegroup-ni-runmode \
 	packagegroup-ni-safemode \
+	packagegroup-ni-skyline \
 	packagegroup-ni-tzdata \
 	packagegroup-ni-wifi \
 	packagegroup-core-x11 \
--- a/recipes-core/packagegroups/packagegroup-ni-skyline.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-skyline.bb
@@ -1,0 +1,37 @@
+# (C) Copyright 2024,
+#  National Instruments Corporation.
+#  All rights reserved.
+
+SUMMARY = "Open Source package dependencies for the NI Skyline software stack."
+LICENSE = "MIT"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+inherit packagegroup
+
+RDEPENDS:${PN} = "\
+	salt-common \
+	salt-minion \
+"
+
+# Miscellaneous python deps for closed-source Skyline components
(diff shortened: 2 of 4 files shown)


chaitu236 (approved):

amstewart:
# Patch V2
I was wondering why my RDEPENDS didn't seem to be correctly applying to `salt-common`. The reason was that I had accidentally added them to the *wrong* `RDEPENDS:salt-common`. :/

This revision:
1. Fixes the salt-common RDEPENDS to actually have the new dependencies.
2. Moves the confusing skyline dependencies to their own packagegroup.

chaitu236:
> I was wondering why my RDEPENDS didn't seem to be correctly applying to `salt-common`. The reason was that I had accidentally added them to the _wrong_ `RDEPENDS:salt-common`. :/

Both `RDEPENDS:salt-common`'s look the same.. how do they behave differently?

PS: I see that the top one is `+=` and bottom one replaces that list with `=` is this the reason? I'd have expected bitbake to throw away the first list when it is replaced later but I guess it still builds them?

chaitu236 (approved):

amstewart:
> > I was wondering why my RDEPENDS didn't seem to be correctly applying to `salt-common`. The reason was that I had accidentally added them to the _wrong_ `RDEPENDS:salt-common`. :/
> 
> Both `RDEPENDS:salt-common`'s look the same.. how do they behave differently?
> 
> PS: I see that the top one is `+=` and bottom one replaces that list with `=` is this the reason? I'd have expected bitbake to throw away the first list when it is replaced later but I guess it still builds them?

That's correct. The first set gets completely overwritten by the hard-assignment later in the recipe. Hence why my previous change wasn't actually showing up in the IPK's dependencies.

That implementation came from Shruthi and Andrei Z's attempt to fixup the recipe back in the dunfell timeframe. There is a non-zero chance that it was unintentional. But I *think* the intention was to provoke bitbake into adding that first RDEPENDS set to the task tree, but then use the second set for the IPK dependencies. The logic being that the first set are deps for skyline IPKs which are not controlled by OE, but which we still need in the core feeds.

PR #700, opened 2024-05-10
Temporarily disable unbuildable packages in extra feed

rajendra-desai-ni:
It seems that many recipes declared as RDEPENDS in the extra packagefeed have dropped from the layers. That doesn't fail the build, but it renders the whole feed metapackage as unbuildable and invalidates even those recipes which would otherwise be buildable.

This change is to unblock building of packagefeed-ni-extra group so that we can debug and fix the disabled packages and then reenable back.

[AB#2703160](https://dev.azure.com/ni/DevCentral/_workitems/edit/2703160/)

**Testing**
- [X] bitbake --continue packagefeed-ni-extra

### DIFF
--- a/files/group
+++ b/files/group
@@ -6,6 +6,7 @@ ni:x:500:
 openvpn:x:499:
 niwscerts:x:498:
 # free space
+arpwatch:x:402:
 ptest:x:401:
 ossec:x:400:
 mysql:x:399:
--- a/files/passwd
+++ b/files/passwd
@@ -6,6 +6,7 @@ webserv:x:501::::
 lvuser:x:500::::
 openvpn:x:499::::
 # free space
+arpwatch:x:402::::
 ptest:x:401::::
 ossec:x:400::::
 mysql:x:399::::
--- a/recipes-core/packagegroups/packagefeed-ni-extra.bb
+++ b/recipes-core/packagegroups/packagefeed-ni-extra.bb
@@ -14,7 +14,6 @@ RDEPENDS:${PN}:append:x64 = "\
 		geany \
 		gimp \
 		gnuplot \
-		gnuradio \
 		gtk+3 \
 		iceauth \
 		fltk \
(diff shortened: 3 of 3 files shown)


files/passwd
```diff
@@ -57,3 +57,4 @@ opensaf:x:342::::
 # free space
 radiusd:x:301::::
 postgres:x:28::::
+arpwatch:x:402::::
```
amstewart: Could you move this line up so that the UID/GIDs are sorted in reverse order?

amstewart (changes requested):
When you're making individual recipe changes, it's helpful to prepend your commit summary line with the name of the recipe or subsystem that you're changing. ie. `pkggrp-ni-extra: disable unbuildable packages`.

Could you rewrite your summaries in that style?

files/group
```diff
@@ -70,3 +70,4 @@ tracing:x:341:
 radiusd:x:301:
 postgres:x:28:
 mail:x:8:
+arpwatch:x:402:
```
chaitu236: Where did you get these values from?

files/passwd
rajendra-desai-ni: I have updated the commit summary line and have also moved arpwatch UID/GIDs up in a proper order.

files/group
rajendra-desai-ni: I just reserved an available static ID to arpwatch without which the extra feeds build would fail.

amstewart (approved):

files/group
chaitu236: Not sure if this is the right thing to do but can merge this for now. We'll have to take a closer look at how best to fix this. Please create a WI for it.

chaitu236 (approved):

files/group
amstewart: It's not unusual that we have to define a uid/gid for a service that is getting pulled into the feeds. Because we chose to use the OE static_uidgid configuration, we have to define all users and groups at build time. The numbers we choose to assign them are arbitrary and only follow the numbering range policy that we defined, and is in comments at the top of the file. FWIW, nothing about his change seems odd to me.

PR #702, opened 2024-05-17
cyclictest-container: install influxdb with apt

rajendra-desai-ni:
Changed the installation of influxdb in the Dockerfile to install with apt install instead of pip install as the kernel-containerized-performance-tests seemed to be broken due to this.

[AB#2737868](https://dev.azure.com/ni/DevCentral/_workitems/edit/2737868/)

Reference to installation of influxdb package on Debian distro: [Link](https://pypi.org/project/influxdb/5.3.2/#:~:text=On%20Debian/Ubuntu,install%20python%2Dinfluxdb)

**Testing**
Tested by building the image manually
image


chaitu236 (approved):
This change looks fine to me. But we also need to make sure issues like these end up failing the event run.
Can you check if this something we can easily fix, and if so put up a PR?

amstewart (approved):

PR #704, opened 2024-05-29
nisystemreplication: Add compatibility check, support for answers file, progress bar

chaitu236:
This patch set adds
1. Compatibility checks so that only compatible images are listed for Set Image
2. Support for answers file so that there would be no user prompts
3. Show progress bar when getting and setting image from nisystemreplication

WI:[ AB#2372267](https://dev.azure.com/ni/DevCentral/_workitems/edit/2372267), [AB#2372293](https://dev.azure.com/ni/DevCentral/_workitems/edit/2372293)

### Testing
- [x] Verified that Get Image creates a systemimage.conf file
- [x] Verified that images taken on incompatible targets and images with missing .conf file or .tgz file are not listed in "Set Image"
- [x] Verified that with following answers file, Get Image and Set Image don't prompt for user interaction and apply configuration specified in the .conf file.
         
image

- [x] Verified that a progress bar is shown for get and set operations from nisystemreplication but not when nisystemimage is invoked without the new argument
Get Image:
image
Set Image:
image
 
- [x] Verify LV workflow to make sure progress bar doesn't break anything
         - LV workflow doesn't seem to be affected but even then, updated the code to not show progress bar by default in nisystemimage. nisystemreplication uses a cmdline argument to show the progress bar. This is to ensure we don't unintentionally break any existing workflows I haven't tested for.
- [x] Test on serial console to see how progress bar shows up
         - Apparently we configured the serial console to show a shell and not show any console output like user prompts; so progress bar is not displayed on serial console.

### DIFF
--- a/recipes-core/initrdscripts/files/ni_provisioning.answers.default
+++ b/recipes-core/initrdscripts/files/ni_provisioning.answers.default
@@ -45,3 +45,7 @@ PROVISION_PART_NILRT_LABEL="nilrt"
 # A hostname to configure after provisioning.
 # An empty string means hostname is generated on first boot.
 PROVISION_HOSTNAME=""
+
+# System Image name used for "Get Image" and "Set Image"
+# No spaces, no special characters except '-'
+PROVISION_SYSTEMIMAGE_NAME=""
--- a/recipes-ni/ni-systemreplication/files/nisystemreplication
+++ b/recipes-ni/ni-systemreplication/files/nisystemreplication
@@ -89,11 +89,14 @@ image_get () {
 	[ $estimated_image_size -eq "0" ] && die "Error estimating image size"
 	[ $nirecovery_available_space -lt $estimated_image_size ] && die "Image might not fit on media. Estimated image size: $estimated_image_size. Available space on media: $nirecovery_available_space"
 
-	# If no image name is provided, retry
-	local image_name=$(get_image_name)
-	while ! is_image_valid $image_name; do
+	local image_name=$PROVISION_SYSTEMIMAGE_NAME
+	if [ -z $image_name ]; then
 		image_name=$(get_image_name)
-	done
+		# If no image name is provided, retry
+		while ! is_image_valid $image_name; do
+			image_name=$(get_image_name)
+		done
+	fi
 
 	mkdir -p $NIRECOVERY_MOUNTPOINT/Images/$image_name
 	nisystemimage getall -x tgz -f $NIRECOVERY_MOUNTPOINT/Images/$image_name/systemimage.tgz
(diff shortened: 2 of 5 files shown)


recipes-ni/ni-systemreplication/files/nisystemreplication
```diff
@@ -100,15 +106,27 @@ image_get () {
...
-# Presents a list of existing images and let's user select one
+# Presents a list of existing compatible images and let's user select one
 select_image_from_list () {
 	[ ! -d $NIRECOVERY_MOUNTPOINT/Images ] && return
 
-	readarray -t images <<< "$(find $NIRECOVERY_MOUNTPOINT/Images -name systemimage.tgz | sort | sed "s,$NIRECOVERY_MOUNTPOINT/Images/,," | sed 's,/systemimage.tgz,,')"
+	images=()
+
+	for image in $(ls $NIRECOVERY_MOUNTPOINT/Images); do
+		if [ ! -f $NIRECOVERY_MOUNTPOINT/Images/$image/systemimage.tgz ] || [ ! -f $NIRECOVERY_MOUNTPOINT/Images/$image/systemimage.conf ]; then
```
amstewart, line 120: Can we warn when this condition happens? I imagine that it will be unexpected for the user.

recipes-ni/ni-systemreplication/files/nisystemreplication
```diff
@@ -100,15 +106,27 @@ image_get () {
...
+			continue
+		fi
+
+		local ImageDeviceDesc=$(sed 's/DeviceDesc=//g' $NIRECOVERY_MOUNTPOINT/Images/$image/systemimage.conf)
```
amstewart: Probably `awk` is better for pulling a key-value pair out of the conf file; right? What happens if the conf file contains other values?

recipes-core/initrdscripts/files/ni_provisioning.answers.default
amstewart, line 1: Could you rev the version number atop this file?

amstewart (changes requested):

recipes-ni/ni-systemreplication/files/nisystemreplication
chaitu236: > Probably awk is better for pulling a key-value pair out of the conf file; right?

Are you thinking something like `awk -F= '{if ($1 == "DeviceDesc") print $2}'`?

> What happens if the conf file contains other values?

We could let that be undefined behavior as that could only happen when user has tampered with the file.

recipes-core/initrdscripts/files/ni_provisioning.answers.default
chaitu236: That would also require us to update https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/recipes-core/initrdscripts/files/ni_provisioning.common#L601 - which would mean that any customer created answers file with V1 would not work with this recovery image. Do we want to do that seeing that we haven't broken backward compatibility?

In the past we didn't update the number when adding new variables e.g., https://github.com/ni/nilrt/commit/9b8f0c223a3410470c0ed9b5a3cc87dac7e2f67a

recipes-ni/ni-systemreplication/files/nisystemreplication
chaitu236: Updated to `awk`

recipes-ni/ni-systemreplication/files/nisystemreplication
chaitu236: Done.

recipes-core/initrdscripts/files/ni_provisioning.answers.default
amstewart: Sure. That makes sense.

amstewart (approved):

PR #709, opened 2024-06-27
nisystemreplication: Replicate ext4 features

chaitu236:
When "Set Image" operation is done from a recovery media containing newer OS version, mkfs.ext4 may create ext4 partitions with newer features enabled which may not be compatible with the grub version from the System Image being applied.

So save ext4 features used for creating nibootfs during "Get Image" operation and only enable them during "Set Image" operation.

Also, detect if filesystem contains unsupported features and error out in "Get Image"

WI: [AB#2765764](https://dev.azure.com/ni/DevCentral/_workitems/edit/2765764)

### Testing
- [x] "Get Image" operation now saves `Ext4Features` in `systemimage.conf`
- [x] `Ext4Features` specified in `systemimage.conf` are applied during "Set Image" operation and no other default features are enabled.
- [x] Doing a "Set Image" operation using "kirkstone" image with dist-next recovery media works on a VM. VM is bootable afterwards.
- [x] Doing "Get Image" operation on a dist-next provisioned VM from kirkstone recovery media gives out following error
image

### Notes
- Will cherry-pick this into kirkstone.
- Detecting filesystem features in "Set Image" to print a more descriptive error is not easy so not doing that. We can do it in future if desired. Applying dist-next image from kirkstone recovery media shows following error for now

image

### DIFF
--- a/recipes-ni/ni-systemreplication/files/nisystemreplication
+++ b/recipes-ni/ni-systemreplication/files/nisystemreplication
@@ -36,6 +36,10 @@ get_ext4_features () {
 	echo $(debugfs -R features $bootfs_device 2>/dev/null | awk -F: '{if ($1 == "Filesystem features") print $2}' | xargs | sed 's/ /,/g')
 }
 
+ext4_has_unsupported_features () {
+	e2fsck -n LABEL=nibootfs 2>&1 | grep "unsupported feature"
+}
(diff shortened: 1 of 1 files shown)


gratian (approved):

chaitu236:

PR #711, opened 2024-08-07
kernel-test-nohz: tweak parameters to improve performance

gratian:
Tweak parameters for the nohz kernel test to:

- Run test at FIFO/1 instead of FIFO/98. This improves test results because it allows the kernel threads pinned to the isolated CPU to run and get out of the way instead of keeping the scheduler tick on (i.e. because they are runnable but never get a chance to run).

- Use `tsc=reliable` kernel parameter instead of `tsc=nowatchdog`. This disables a repeating timer used to check the TSC validity on all CPUs, including isolated ones. This recurring timer adds to the latency observed on the isolated CPU(s).

WI: [AB#2796625](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2796625)

### Testing

- [x] `bitbake kernel-test-nohz` and served resulting packages from local feed
- [x] `opkg install kernel-test-nohz-ptest` on a PXIe-8881
- [x] `ptest-runner -t 345600`
- [x] validated the priority of nohz_test  with `chrt -p <pid>`
- [x] validated the kernel parameter change with `cat /proc/cmdline`

Initial test numbers show significant improvement in the max latency number; see associated bug for details. Final results to be validated by the ATS system.

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-test-nohz-files/nohz_test.c
+++ b/recipes-kernel/kernel-tests/kernel-test-nohz-files/nohz_test.c
@@ -420,7 +420,7 @@ static void validate_results(struct histogram_data *h)
 	p_99_999 = get_percentile(99.999, h);
 	p_99_9999 = get_percentile(99.9999, h);
 
-	if (h->max > max_latency)
+	if (h->max >= max_latency)
 		error_exit("Maximum latency exceeded", stats_summary, h->cnt, h->max, p_99_999, p_99_9999);
 	if (p_99_999 > percentile_99_999)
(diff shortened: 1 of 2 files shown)


gratian:
@chaitu236 noticed an inconsistency between the way we test for the max latency threshold in the test loop vs. when validating results to decide on a pass/fail result. Added fix commit.

chaitu236 (approved):

chaitu236:
Cherry-picked to `nilrt/master/next`.

PR #714, opened 2024-08-13
[next] Add a packagegroup for SNAC

amstewart:
Add a packagegroup to track package dependencies of the NILRT Secured,
Network-Attached Controller (SNAC) configuration.

Since the SNAC configuration is officially supported, add it to the core
package feed.

This packagegroup SHOULD NOT be installed to the NILRT runmode or
safemode images.

# Testing
* [x] Built `packagegroup-ni-snac` with this change. Meta package seems well formed.
* [x] Ran `bitbake -g packagefeed-ni-core` with this change and confirmed that the new packagegroup is in the pn-buildlist.

# Meta
* This is a NILRT 11 change only.

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-snac.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-snac.bb
@@ -5,9 +5,7 @@ LICENSE = "MIT"
 inherit packagegroup
 
 
-RDEPENDS:${PN} = ""
-
-RDEPENDS:${PN}:append = "\
+RDEPENDS:${PN} = "\
 	cryptsetup \
 	ntp \
 	tmux \
--- a/recipes-core/packagegroups/packagegroup-ni-snac.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-snac.bb
@@ -1,0 +1,11 @@
+SUMMARY = "Open source package dependencies for the NILRT SNAC configuration."
+LICENSE = "MIT"
+
+
+inherit packagegroup
+
+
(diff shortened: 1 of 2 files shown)


recipes-core/packagegroups/packagegroup-ni-snac.bb
```diff
@@ -0,0 +1,10 @@
+SUMMARY = "Open source package dependencies for the NILRT SNAC configuration."
+LICENSE = "MIT"
+
```
amstewart, line 3: Forgot the packagegroup inherit. :) Will fix.

amstewart:
# Patch V2
* Fixed missing inherit.

recipes-core/packagegroups/packagegroup-ni-snac.bb
```diff
@@ -0,0 +1,14 @@
...
+
+inherit packagegroup
+
+
+RDEPENDS:${PN} = ""
+
+RDEPENDS:${PN}:append = "\
```
chaitu236: Why list these in `RDEPENDS:${PN}:append` and not in `RDEPENDS:${PN}`?

amstewart: I guess force of habit - since these packagegroups normally have many subsections. But I guess it doesn't make sense in this case. Will fix.

amstewart: Fixed.

chaitu236 (approved):

PR #715, opened 2024-08-15
packagegroup-ni-internal-deps.bb: Add patool dependencies

usercw88:
packagegroup-ni-internal-deps.bb: Add PAtools dependencies

[AB#2526086](https://dev.azure.com/ni/DevCentral/_workitems/edit/2824160)

Testing:
- [x] Built locally and verified desired packages are pulled in

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-internal-deps.bb
@@ -60,7 +60,7 @@ RDEPENDS:${PN} += "\
 	libxml-parser-perl \
 "
 
-# Required for a mobilize step that installs a specific Python version 
+# Required for a mobilize step that installs a specific Python version
 # and requires building Python on the test system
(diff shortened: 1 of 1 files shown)


chaitu236 (changes requested):
Change looks fine but please remove the "to main feed" from commit message as we're not adding this to ni-main feed.

There is also a 50/72 recommendation for commit message length; i.e., first line around 50 chars, rest of the lines around 72. We don't always follow this strictly but whenever it's possible.

So commit message could be something like `pkggrp-ni-internal-deps: Add PAtools dependencies` perhaps? You'll find similar commit messages in [history](https://github.com/ni/meta-nilrt/commits/nilrt/master/kirkstone/recipes-core/packagegroups/packagegroup-ni-internal-deps.bb) of this file.

chaitu236:
Cherry-picked to nilrt/master/next - 6da2d19c6a90648db377297eb0aaa69477fb8e53

PR #718, opened 2024-08-19
When installing tmux, install a default config file

texasaggie97:
When configuring SNAC mode, tmux will be set as the default shell so that inactivity timeouts can be enforced

When installing tmux also install a default config file that sets vlock as the lock command. The actual timeout will be enabled with SNAC mode

### DIFF
--- a/recipes-extended/tmux/files/tmux.conf
+++ b/recipes-extended/tmux/files/tmux.conf
@@ -1,2 +1,3 @@
 set -g lock-command vlock
 bind X lock-session
+source-file -q /usr/share/tmux/conf.d/*
--- a/recipes-extended/tmux/files/tmux.conf
+++ b/recipes-extended/tmux/files/tmux.conf
@@ -1,0 +1,2 @@
+set -g lock-command vlock
+bind X lock-session
--- a/recipes-extended/tmux/files/tmux.conf
+++ b/recipes-extended/tmux/files/tmux.conf
@@ -1,0 +1,3 @@
+set -g lock-command vlock
+bind X lock-session
+source-file -q /usr/share/tmux/conf.d/*
--- a/recipes-extended/tmux/tmux_3.%.bbappend
+++ b/recipes-extended/tmux/tmux_3.%.bbappend
@@ -1,0 +1,15 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI += "\
+    file://tmux.conf \
+"
+
+do_install:append() {
+    install -d ${D}${sysconfdir}
+    install -m 644 ${WORKDIR}/tmux.conf ${D}${sysconfdir}/tmux.conf
+    install -d ${D}/usr/share/tmux/conf.d
+}
+
+FILES:${PN}     += "${sysconfdir}/tmux.conf"
+CONFFILES:${PN} += "${sysconfdir}/tmux.conf"
+RDEPENDS:${PN}:append = " vlock"
--- a/recipes-extended/tmux/tmux_3.%.bbappend
+++ b/recipes-extended/tmux/tmux_3.%.bbappend
@@ -1,6 +1,6 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
 
-CONFFILES += "\
+SRC_URI += "\
     file://tmux.conf \
 "
 
--- a/recipes-extended/tmux/tmux_3.%.bbappend
+++ b/recipes-extended/tmux/tmux_3.%.bbappend
@@ -1,0 +1,15 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI += "\
+    file://tmux.conf \
+"
+
+FILES:${PN}     += "${sysconfdir}/tmux.conf"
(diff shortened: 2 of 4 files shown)


recipes-extended/tmux/files/tmux.conf
```diff
@@ -0,0 +1,2 @@
+set -g lock-command vlock
+bind X lock-session
```
amstewart, line 2: I had a thought while we were talking about sourcing conf files on Teams. Does the tmux `source` directive work with wildcards? If it does, maybe we can make this even better by having the tmux package install an empty `/usr/share/tmux/conf.d` directory. Then this OE conf file can say `source /usr/share/tmux/conf.d/*`, and the `nilrt-snac.ipk` can just install the remaining timeout snippet into a file it owns there without having to write anything into the `/etc/tmux.conf`. What do you think?

recipes-extended/tmux/tmux_3.%.bbappend
```diff
@@ -0,0 +1,15 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+CONFFILES += "\
+    file://tmux.conf \
+"
+
+FILES:${PN}     += "${sysconfdir}/tmux.conf"
+CONFFILES:${PN} += "${sysconfdir}/tmux.conf"
```
amstewart: It doesn't functionally matter in this case, but move packaging variables like this below the `do_install()` (and other task-overrides), per [the styleguide](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/docs/styleguide.md#layout). Keeping to that standard ordering helps make recipes more readable and to avoid obscure errors with ordering that can sometimes occur when you have more complicated recipes that inherit behavior from bbclasses.

recipes-extended/tmux/tmux_3.%.bbappend
```diff
@@ -1,6 +1,6 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
 
-CONFFILES += "\
+SRC_URI += "\
```
amstewart: Make this commit a fixup commit for aa47fc39db1c7964cbe494a3c2e762116efa32cb, and then autosquash it before submitting the PR. eg.
```bash
git reset 1608eb60bc3348f2e9f345df481ece32c94ecbad~
git add recipes-extended/tmux/*
git commit --fixup aa47fc39db1c7964cbe494a3c2e762116efa32cb
git rebase -i --autosquash
```

amstewart (changes requested):
(1) meta-nilrt, like basically all layers and repos in the OpenEmbedded ecosystem, uses commits as the fundamental unit of change. So make sure your commit messages contain the motivation, solution, and justification for your changes. The reason being that OE layer commits are regularly cherry-picked and rebased, outside of Github.

(2) You can squash this entire patchset into a single commit. Definitely at least squash the final commit (see below) and the merge commit out of the git history. Once you do, you can just force-push the updated history to your branch ref, and github will update the PR. Then I would come back to the PR and make a comment like
```
# Patch V2
* Changed blah blah blah.
```
... so that reviewers know that you made changes and that you are ready for another review.

(3) meta-nilrt doesn't have a PR template (though I should add one), but you should at least give a 'Testing' section, like #699.

amstewart:
Oh, and you can add 'AB#${AZDO ID}' to the commit message to auto-link the PR against your work item, and at-mention 'ni/rtos' to ping the RTOS team for a review.

amstewart (changes requested):
One more thing that is probably not obvious: I think `tmux` is already in the core package feed. But since we're explicitly using it in the SNAC design, lets add it to the new [`packagegroup-ni-snac`](https://github.com/ni/meta-nilrt/blob/nilrt/master/next/recipes-core/packagegroups/packagegroup-ni-snac.bb) meta-package. That will ensure that it always gets built, even if the other dependency falls away for some reason. I would make the packagegroup file change its own commit - for cleanliness. And remember to alpha-sort the RDEPENDS for my sanity. :)

recipes-extended/tmux/tmux_3.%.bbappend
texasaggie97: I didn't look at enough of them. The ones I looked at (the top part of the find all CONFFILES) had them before `do_install()`. If I look further down the list, I do see where these types of commands are after `do_install()`. I will move them since that is actually what I prefer and only moved them to be consistent with what I had found.

recipes-extended/tmux/tmux_3.%.bbappend
texasaggie97: Done

recipes-extended/tmux/tmux_3.%.bbappend
texasaggie97: Moved

recipes-extended/tmux/files/tmux.conf
texasaggie97: Updated to create and source conf files from /usr/share/tmux/conf.d/

texasaggie97:
Abandoning this since I screwed up the branch trying to squash all the commits

PR #721, opened 2024-08-21
nisystemreplication: Split image files to fit FAT32

chaitu236:
FAT32 max file size is limited to 2^32 - 1. So split the image files at this boundary.

### Justification

[AB#2834501](https://dev.azure.com/ni/DevCentral/_workitems/edit/2834501)

### Testing

* [x] Built core feed and recovery image with these changes
* [x] Used the recovery iso to "Get Image" and "Set Image" on a VM where `nirootfs` was 12GB. The image files were split at max file size boundary of FAT32 i.e., 4294967295 bytes.
* [x] Verified that setting the split size to a value larger than 4294967295 bytes causes `File too large` error during "Get Image".
* [x] Verified that largest files on `nirootfs` had the same md5 hash before and after imaging.
* [x] Verified that "Set Image" using the split 12G image works on VM with just 512MB memory. i.e., this this approach doesn't require large memory.

### Procedure

### Notes
- Needs cherry-pick into `nilrt/master/next`

### DIFF
--- a/recipes-ni/ni-systemreplication/files/nisystemreplication
+++ b/recipes-ni/ni-systemreplication/files/nisystemreplication
@@ -80,7 +80,7 @@ is_image_valid () {
 		echo "No image name specified. Please specify an image name" && return 1
 	fi
 	# If image already exists, confirm if it should be overwritten
-	if [ -f $NIRECOVERY_MOUNTPOINT/Images/$image_name/systemimage.tgz ]; then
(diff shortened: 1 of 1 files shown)


gratian (approved):

chaitu236:
Cherry-picked to nilrt/master/next - a3fbdd40adbaccea4975fd032e0b82225d4111aa

PR #724, opened 2024-08-27
enable pam-plugin-faillock when it's installed with custom settings

AlexHearnNI:
### Summary of Changes

- update the pam-plugin-faillock package so that the plugin gets enabled when it's installed
- modify some faillock configuration settings
- prevent pam-plugin-faillock from being installed when `ni-auth` is installed

### Justification

This change simplifies Secured, Network-Attached Controller (SNAC) configuration, [AB#2816939](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2816939). faillock is required to be enabled on a SNAC. The faillock settings were chosen to comply with SNAC requirements. The conflict with `ni-auth` was added because from testing it appears that the faillock plugin is incompatible with the ni-auth plugin.

### Testing

I tested that opkg returns a clear error message when installing `pam-plugin-faillock` describing the conflict if `ni-auth` is installed. I confirmed that `libpam-runtime` contains the customized `/etc/security/faillock.conf`. After removing `ni-auth`, I installed `pam-plugin-faillock` and confirmed that the configuration files changed as I expected.

```text
>~# cat /etc/pam.d/common-auth
#
# /etc/pam.d/common-auth - authentication settings common to all services
#
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.).  The default is to use the
# traditional Unix authentication mechanisms.

# here are the per-package modules (the "Primary" block)
auth    requisite pam_faillock.so preauth
auth    [success=2 default=ignore]      pam_unix.so nullok_secure
auth    [default=die] pam_faillock.so authfail
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
auth    sufficient pam_faillock.so authsucc
```

I confirmed that `common-auth` was reverted after removing `pam-plugin-faillock`.

### Procedure

@ni/rtos @amstewart

### DIFF
--- a/recipes-extended/pam/libpam_1.%.bbappend
+++ b/recipes-extended/pam/libpam_1.%.bbappend
@@ -1,1 +1,23 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
+
+SRC_URI += "\
+	file://security/faillock.conf \
+"
+
+do_install:append() {
+	install -m 644 ${WORKDIR}/security/faillock.conf ${D}${sysconfdir}/security/faillock.conf
+}
+
+pkg_postinst:pam-plugin-faillock:append() {
+	# enable faillock
+	sed -E -i 's/^(.+)success=1(.+)$/auth    requisite pam_faillock.so preauth\n\1success=2\2\nauth    [default=die] pam_faillock.so authfail/' "${sysconfdir}/pam.d/common-auth"
+	echo "auth    sufficient pam_faillock.so authsucc" >> "${sysconfdir}/pam.d/common-auth"
+}
+
+pkg_prerm:pam-plugin-faillock:append() {
+	# disable faillock
+	sed -E -i '/pam_faillock.so/d' "${sysconfdir}/pam.d/common-auth"
+	sed -E -i 's/^(.+)success=2(.+)$/\1success=1\2/' "${sysconfdir}/pam.d/common-auth"
+}
+
+RCONFLICTS:pam-plugin-faillock:append = " ni-auth"
--- a/recipes-extended/pam/libpam/security/faillock.conf
+++ b/recipes-extended/pam/libpam/security/faillock.conf
@@ -1,0 +1,62 @@
+# Configuration for locking the user after multiple failed
+# authentication attempts.
+#
+# The directory where the user files with the failure records are kept.
+# The default is /var/run/faillock.
+# dir = /var/run/faillock
+#
+# Will log the user name into the system log if the user is not found.
+# Enabled if option is present.
+audit
+#
+# Don't print informative messages.
+# Enabled if option is present.
+silent
+#
+# Don't log informative messages via syslog.
+# Enabled if option is present.
+# no_log_info
+#
+# Only track failed user authentications attempts for local users
+# in /etc/passwd and ignore centralized (AD, IdM, LDAP, etc.) users.
+# The `faillock` command will also no longer track user failed
+# authentication attempts. Enabling this option will prevent a
+# double-lockout scenario where a user is locked out locally and
+# in the centralized mechanism.
+# Enabled if option is present.
+# local_users_only
+#
+# Deny access if the number of consecutive authentication failures
+# for this user during the recent interval exceeds n tries.
+# The default is 3.
+deny = 3
+#
+# The length of the interval during which the consecutive
+# authentication failures must happen for the user account
+# lock out is <replaceable>n</replaceable> seconds.
+# The default is 900 (15 minutes).
+fail_interval = 900
+#
+# The access will be re-enabled after n seconds after the lock out.
+# The value 0 has the same meaning as value `never` - the access
+# will not be re-enabled without resetting the faillock
+# entries by the `faillock` command.
+# The default is 600 (10 minutes).
+unlock_time = 0
+#
+# Root account can become locked as well as regular accounts.
+# Enabled if option is present.
+# even_deny_root
+#
+# This option implies the `even_deny_root` option.
+# Allow access after n seconds to root account after the
+# account is locked. In case the option is not specified
+# the value is the same as of the `unlock_time` option.
+# root_unlock_time = 900
+#
+# If a group name is specified with this option, members
+# of the group will be handled by this module the same as
+# the root account (the options `even_deny_root>` and
+# `root_unlock_time` will apply to them.
+# By default, the option is not set.
+# admin_group = <admin_group_name>


recipes-extended/pam/libpam_1.%.bbappend
```diff
@@ -1 +1,16 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
+
+RCONFLICTS:pam-plugin-faillock += "ni-auth"
```
amstewart: The `:append` operator and the `+=` operator in bitbake do nearly the same thing. But for *reasons* the :append operator is preferred for use in bbappend files. Try changing this to
```
RCONFLICTS:pam-plugin-faillock:append = " ni-auth"
```
The additional space (` `) in the assignment value is necessary.

---

Also, move the `RCONFLICTS` assignment below the `pkg_postinst` task override, per the [layer styleguide](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/docs/styleguide.md#layout).

recipes-extended/pam/libpam_1.%.bbappend
```diff
@@ -1 +1,16 @@
...
+
+pkg_postinst:pam-plugin-faillock:append() {
+	# configure faillock settings
+	sed -i 's/^# audit$/audit/' "${sysconfdir}/security/faillock.conf"
+	sed -i 's/^# silent$/silent/' "${sysconfdir}/security/faillock.conf"
+	sed -i 's/^# deny = 3$/deny = 3/' "${sysconfdir}/security/faillock.conf"
+	sed -i 's/^# fail_interval = 900$/fail_interval = 900/' "${sysconfdir}/security/faillock.conf"
+	sed -i 's/^# unlock_time = 600$/unlock_time = 0/' "${sysconfdir}/security/faillock.conf"
```
amstewart: Instead of `sed`ing content into a file that is owned by this package, let's just override the default configuration file we get from upstream OE-core. Place the customized `faillock.conf` into this recipe's `libpam/` FILES directory and append to the recipe `do_install` to install that into the `sysconfdir`.

recipes-extended/pam/libpam_1.%.bbappend
```diff
@@ -1 +1,16 @@
...
+
+	# enable faillock
+	sed -E -i 's/^(.+)success=1(.+)$/auth    requisite pam_faillock.so preauth\n\1success=2\2\nauth    [default=die] pam_faillock.so authfail/' "${sysconfdir}/pam.d/common-auth"
+	echo "auth    sufficient pam_faillock.so authsucc" >> "${sysconfdir}/pam.d/common-auth"
```
amstewart, line 14: (I'm basically always going to push back on using `sed`.) The [PAM rules syntax] supports an `include` operation. Does that work on directories, or with a wildcard. Since we're _plugging_ PAM modules from packages (instead of relying on the sysadmin to write the rules), the better design here may be to:
1. modify the base PAM package to create an `${D}${sysconfdir}pam.d/common-auth.d` directory.
2. modify the `common-auth` default file in the base PAM package to `include .../common-auth.d/` in the appropriate place.
3. have the `pam-plugin-faillock` package install its own rules file (that lives as a static file in this recipe) into `.../common-auth.d/pam-plugin-faillock.conf`.

What do you think?

amstewart (changes requested):
NILRT, the NILRT meta layers, and OE repos more generally use commits as the unit of change. Remember to prefix your commit summary with the recipe name you're changing, eg. `libpam: enable pam-plugin-faillock on install`. And put your justifications and change descriptions in the commit summary.

recipes-extended/pam/libpam_1.%.bbappend
AlexHearnNI: `faillock.conf` is owned by `libpam-runtime`, which is on the base system image. I structured the change this way so that only users who install `pam-plugin-faillock` will be affected. Can you confirm that it's acceptable for me to modify `faillock.conf` on the base system image?

amstewart: Which package installs the faillock pam `.so`?

AlexHearnNI: `/lib/security/pam_faillock.so` is installed by the `pam-plugin-faillock` package.

amstewart: So yes; it's fine that we modify the `faillock.conf` file. We own the distro and can provide whatever defaults we want, including enabling some features by default.

I'm a little confused as to why libpam-runtime owns the `faillock.conf` though. It seems like that should be in `pam-plugin-faillock`.

recipes-extended/pam/libpam_1.%.bbappend
AlexHearnNI: I could not find a good use of `include`. If you `include` a missing file, then it fails the authentication. I tried several different syntaxes, and I was not able to `include` a directory or wildcard.

amstewart: :( Well, in that case let's just make sure that you can also remove this package and have the rule get correctly removed from the `common-auth` file.

You may want to match on `.*pam_faillock.so.*` for the `sed` regex. None of the other rules should refer to that library.

chaitu236: Can we not replace the whole `common-auth` with our version like meta-cloud-service's [bbappends](https://github.com/ni/meta-cloud-services/blob/nilrt/master/next/meta-openstack/recipes-extended/libpam/libpam_openstack.inc) does instead of using `sed`? Might be cleaner that way.

> Since we're plugging PAM modules from packages

Do we expect to add more rules to this file from packages other than `libpam-*`?

AlexHearnNI: The `libpam-runtime` package owns `common-auth`. The `meta-cloud-services` example you mentioned is updating the `libpam-runtime` package. This PR is changing the `pam-plugin-faillock` package, which does not own `common-auth`.

recipes-extended/pam/libpam_1.%.bbappend
```diff
@@ -1 +1,23 @@
...
+	install -m 644 ${WORKDIR}/security/faillock.conf ${D}${sysconfdir}/security/faillock.conf
+}
+
+pkg_postinst:pam-plugin-faillock:append() {
+	# enable faillock
+	sed -E -i 's/^(.+)success=1(.+)$/auth    requisite pam_faillock.so preauth\n\1success=2\2\nauth    [default=die] pam_faillock.so authfail/' "${sysconfdir}/pam.d/common-auth"
+	echo "auth    sufficient pam_faillock.so authsucc" >> "${sysconfdir}/pam.d/common-auth"
+}
+
+pkg_postrm:pam-plugin-faillock:append() {
```
amstewart: `postinst` steps are normally undone in the package `prerm` script. Otherwise, you might encounter an issue removing the package files and end up in a state where the common-auth script rules pass into non-existant PAM libraries.

amstewart (changes requested):
Looks good pending this one change.

recipes-extended/pam/libpam_1.%.bbappend
AlexHearnNI: Fixed. FWIW, I used pkg_postrm because that's what I saw in [screen](https://github.com/ni/openembedded-core/blob/2566f8e51d56848d8b28f37462160e90253b79fc/meta/recipes-extended/screen/screen_4.9.1.bb#L47).

amstewart (approved):

chaitu236 (approved):

PR #725, opened 2024-08-29
nftables, firewalld: add to runmode

rtollert:
### Summary of Changes

Add nftables (modern Linux packet filtering support) and firewalld (modern Linux high-level firewall administration tool) packages to runmode. At present, neither will be enabled, but they can be enabled by the user. The existing iptables initscript is unchanged. This is tracked by Azure DevOps workitem [AB#2823118](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2823118).

### Testing

Built nilrt-base-system-image and confirmed it runs on a VM. nft runs; iptables still runs. firewall-cmd is not yet tested.

This change will grow the uncompressed BSI by approximately **9.5MB** (!). The largest new files break down along the following lines:

| Directory | Size (KiB) |
|-|-|
| /usr/lib/python-3.10/ | 2458 |
| /usr/lib/girepository-1.0/ | 724 |
| /usr/lib/gobject-introspection/ | 669 |
| /usr/lib/\*.so.\* | 1264 |
| /lib/modules/ | 870 |
| /usr/*bin/ | 897 |
| /usr/share/licenses/ | 341 | 
| /usr/share/firewalld/ | 712 |

Several things about this seem problematic.
- The entirety of the increased disk usage in /usr/lib/girepository-1.0/, /usr/lib/gobject-introspection/, and /usr/share/licenses/, and a substantial fraction of /usr/lib/python-3.10/ — perhaps 3 MiB in all — is solely due to firewalld's dependency on python3-pygobject, which in turn pulls in gobject-introspection, libcairo2 (!), libcairo-gobject2 (!!), python3-pycairo, etc.
- The entirety of the disk usage in /usr/share/firewalld/ is due to a new file firewall-config.glade, which appears to solely exist for the sake of GTK3-based firewalld configuration UI; this appears to be installed as `firewall-config`; we don't need it, but if we need GTK3 anyway, I guess it can't hurt.
- /usr/bin/firewall-applet is a python script which requires PyQt5 which is not installed.

I fear the gobject-related bits will be hard to remove; they've been in firewalld since 2012. That said, the only thing I can find so far that clearly requires it is DBus (?). I think I'm missing something.

* [x] I have built the core package feed with this PR in place. (`bitbake packagefeed-ni-core`)

### Procedure

[...]

PR #726, opened 2024-08-30
[next] nftables, firewalld: add to runmode

rtollert:
### Summary of Changes

Add nftables (modern Linux packet filtering support) to runmode; disabled by default.

Add firewalld (modern Linux high-level firewall administration tool) packages to SNAC. (nftables is in runmode, not SNAC, for symmetry with the kernel configuration, which will include the NFT modules in runmode too.) 

The existing iptables initscript is unchanged. This is tracked by Azure DevOps workitem [AB#2823118](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2823118).

Note: this is a continuation of #725, rebased to `nilrt/master/next`.

### Testing

Built nilrt-base-system-image and confirmed it runs on a VM. nft runs; iptables still runs. firewall-cmd is not yet tested.

This change will grow the uncompressed BSI by approximately 9.5MB; see #725 for details.

### Procedure

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-runmode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-runmode.bb
@@ -23,6 +23,7 @@ RDEPENDS:${PN} = "\
 	librtpi \
 	linux-firmware-radeon \
 	lldpd \
+	nftables \
 	ni-configpersistentlogs \
 	ni-locale-alias \
 	ni-modules-autoload \
--- a/recipes-core/packagegroups/packagegroup-ni-snac.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-snac.bb
@@ -7,6 +7,7 @@ inherit packagegroup
 
 RDEPENDS:${PN} = "\
 	cryptsetup \
+	firewalld \
 	ntp \
 	tmux \
 	libpwquality \


recipes-core/packagegroups/packagegroup-ni-runmode.bb
```diff
@@ -40,10 +40,12 @@ RDEPENDS:${PN} = "\
 RDEPENDS:${PN} += "\
 	cpp \
 	cpp-symlinks \
+	firewalld \
```
chaitu236: Any reason these changes need to be here and not in [packagegroup-ni-snac.bb](https://github.com/ni/meta-nilrt/blob/nilrt/master/next/recipes-core/packagegroups/packagegroup-ni-snac.bb#L1) (especially since we're not even going to enable them by default)?

rtollert: My expectation was that packagegroup-ni-snac was about things that were *specifically* incompatible — or something we know we don't want to do — with the nominal runmode configuration; by default, work should happen in "upstream" packagegroups whenever possible. I do not believe that firewalld is in this category, particularly because it is presently disabled.

That said, I'm not sure I care strongly about this, at least yet, and what's in packagegroup-ni-snac clearly falls in the same general category. So I'll change.

amstewart: No; I think you're right. We should try to put this into the runmode package group.

@chaitu236 Our intention in the long term is to switch to nftables and firewalld. I think the NILRT 11.0 release is a reasonable time to introduce those packages, even if we do not enable them until we've had some experience with them in the SNAC configuration.

rtollert:
Bump for v2.

amstewart (approved):

recipes-core/packagegroups/packagegroup-ni-runmode.bb
chaitu236: Ok, my concern was that we were increasing BSI size with SNAC-only stuff but if the intention is to have nftables and firewalld in BSI anyway, this is fine.

chaitu236 (approved):

PR #727, opened 2024-09-05
meta-nilrt: Upgrade recipes to the latest supported versions

rajendra-desai-ni:
### Summary of Changes

- Updated package version of the recipes to the last released version.
i. dkms from v3.0.10 to v3.0.13
ii. lsbinitscripts from v10.04 to v10.26
iii. apr-iconv from v1.2.1 to v1.2.2
iv. libfmi from v2.3 to v3.0a4

### Justification

The meta-nilrt layer contains some original bitbake recipes, which are unique to the NILRT distro. These recipes will only receive upgrades if the RTOS team commits them. We should make a regular pass during major distro revisions, to consider upgrading these recipes to their latest releases.

### Testing

* [X] bitbake packagegroup-ni-desirable
* [X] bitbake package-index && bitbake nilrt-base-system-image
* [X] unpacked resulting nilrt-base-system-image-x64.tar on a VM and verified the target boots into runmode w/o problems.

### Procedure


rajendra-desai-ni:
For the 2nd change that I have mentioned in the description,
- Removed redundant PV variable to maintain consistency across .bb files:
   Some recipes had a "PV" variable defined, (though did not depend on any upstream code) but the others didn't have this 
   variable in .bb files. I went ahead and removed this variable (assuming it would anyways default to 1.0) from such recipes. 
   Let me know if this change is needed. Otherwise, I can just revert those changes and update the details accordingly.

recipes-devtools/ptest-utils/ptest-utils.bb
```diff
@@ -6,7 +6,6 @@ BUGTRACKER = "https://github.com/ni/meta-nilrt/issues"
 LICENSE = "BSD-3-Clause"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/BSD-3-Clause;md5=550794465ba0ec5312d6919e203a55f9"
 
-PV = "1.4"
```
amstewart: When you don't specify a `PV` in a recipe, and the filename does not have a `_$version.bb` suffix, I think it defaults to `1.0`. We especially don't want that in packages like this, that install content from the recipe's directory, because this is our only way to encode that we have actually upgraded the package contents.

recipes-kernel/dkms/dkms.bb
```diff
@@ -7,15 +7,14 @@ LICENSE = "GPL-2.0-only"
 LIC_FILES_CHKSUM = "file://COPYING;md5=b234ee4d69f5fce4486a80fdaf4a4263"
 
 
-PV = "3.0.10"
+PV = "3.0.13"
```
amstewart, line 10: Have you done a review of the changes in this package version? DKMS is a security critical utility with sketchy maintenance. We need to be cautious about accepting changes.

amstewart (changes requested):
Could you break up this PR into individual recipe changes. Upgrade PRs like this are the most likely to need a revert, if they break something. And it is annoying to have to do a partial revert if onlyone recipe is broken.

recipes-devtools/ptest-utils/ptest-utils.bb
rajendra-desai-ni: Understood. Also, even I did not like the idea of removing "PV" for default versions as a part of this PR itself. Thought of creating a separate tech debt to address it (if needed). I am working on a new PR where there are few more recipes that needs an update. I am soon going to add reviewers to it after testing it out (or maybe I can force push the changes to this PR itself). That's the reason I had made this PR as draft.

>Could you break up this PR into individual recipe changes. Upgrade PRs like this are the most likely to need a revert, if they break something. And it is annoying to have to do a partial revert if onlyone recipe is broken.

Sure, I can do this.

recipes-kernel/dkms/dkms.bb
rajendra-desai-ni: I just skimmed through the changes that went in from 3.0.10 to 3.0.13 but did not find such critical commits that needs to be verified (or at least that's what I observed) other than this commit ([CI: install xz-utils on Debian](https://github.com/dell/dkms/pull/393)). But it looks like it does not affect LinuxRT and we are still using older package and not the affected package (5.6.0/5.6.1). Hope I understood it correctly?
image

Also, is there other way than checking the changes manually that would give proper confirmation that we are not depending on critical changes?

amstewart: >Also, is there other way than checking the changes manually that would give proper confirmation that we are not depending on critical changes?

No. It just comes down to reviewing the upstream changelog and/or PRs.

recipes-devtools/ptest-utils/ptest-utils.bb
rajendra-desai-ni: Added each individual recipe changes into its own commits.

recipes-kernel/dkms/dkms.bb
rajendra-desai-ni: I will do a re-review of changes went into dkms. Will publish the PR for review and resolve this comment after that.

rajendra-desai-ni: I went through all the changes in dkms package from 3.0.10 to 3.0.13 (and in-between versions) and did not encounter any unexpected or unacceptable commits. The changes that went in were mostly test related, readme updates, CI work and other improvements. There were some feature additions which seemed to be non-breaking and not-so-major changes. Marking this comment thread as resolved.

chaitu236: Did you try installing some NI drivers on a real target and ensuring that their modules build/install/load without issues?

chaitu236: @rajendra-desai-ni

amstewart: I think the existing PR and client testing we have is sufficient to test a DKMS upgrade. Especially for a patch-rev.

amstewart (approved):

recipes-kernel/dkms/dkms.bb
rajendra-desai-ni: @chaitu236 : I am able to install some NI drivers on cRIO module with new changes. I am currently setting up my host machine to run some example VIs on the target. I will update here once done.

rajendra-desai-ni: @chaitu236 : All the NI drivers that I tried installing on the cRIO module installed without any issues. Was able to run some example VIs as well.

chaitu236 (approved):

PR #732, opened 2024-09-16
Onboard: Add onboard virtual keyboard

pratheekshasn:
### Summary of Changes

The PR adds the virtual keyboard "Onboard" to scarthgap.

### Justification
[#AB2491688](https://dev.azure.com/ni/DevCentral/_workitems/edit/2491688/) requires the keyboard to be present on scarthgap as well.

The virtual keyboard "Onboard" (present on kirkstone) had been removed from scarthgap due to build errors on python 3.12. This PR adds back the same to scarthgap with corrections required for python3.12.

image

### Implementation
Changes that differ from kirkstone:
1. The recipe adds CXXLAGS `-Werror=declaration-after-statement` to remove build errors that get generated when the recipe builds the source of Onboard.
2. The recipe (and a couple of patch files) used to be present in the meta-openembedded layer, but has now been moved to the meta-nilrt layer since upstream no longer maintains this recipe. We have also decided not to push it upstream since we might be the only ones still using it.
3. Removed the MD5 sum since it is no longer being used.

The keyboard can now be launched and used just like in kirkstone. It also auto-starts on boot, and launches as soon as a text box comes into focus.

### Testing

Tested by installing the built IPK on a VM.
* [x] Keyboard launch
* [x] Type on keyboard by clicking on the keys
* [x] Click on other windows (like the file explorer) to test that the keyboard goes to the background
* [x] Click on any window or textbox where something can be typed to test that the keyboard is back in focus and can type

Signed-off by: Pratheeksha S N <pratheeksha.s.n@ni.com>

### DIFF
--- a/recipes-support/onboard/files/0002-onboard-onhover-seg-fault-fix.patch
+++ b/recipes-support/onboard/files/0002-onboard-onhover-seg-fault-fix.patch
@@ -1,0 +1,52 @@
+From: 1be95325d320122efd5dedf7437839cfcca01f7a
+From: https://github.com/void-linux/void-packages/commit/1be95325d320122efd5dedf7437839cfcca01f7a
+Date: Mon, 23 Sep 2024
+Subject: [PATCH] onboard: fix segfault when hovering over the keyboard
+
+Currently, if the mouse pointer is hovered over the Onboard keyboard, the application crashes with a segmentation fault.
+This is because the new_device_event function is not acquiring the GIL before creating a new object. This patch fixes
+the issue by acquiring the GIL before creating a new object. It also acquires the GIL before queueing the event.
+This patch is taken from the fix provided in the following link:
+https://github.com/void-linux/void-packages/commit/1be95325d320122efd5dedf7437839cfcca01f7a
+
+Signed-off by: Pratheeksha S N <pratheeksha.s.n@ni.com>
+
+Upstream-Status: Inactive-Upstream [lastcommit: 2017-10-09]
+
+--- a/Onboard/osk/osk_devices.c
++++ b/Onboard/osk/osk_devices.c
+@@ -97,13 +97,15 @@ osk_device_event_dealloc (OskDeviceEvent
+ static OskDeviceEvent*
+ new_device_event (void)
+ {
+-    OskDeviceEvent *ev = PyObject_New(OskDeviceEvent, &osk_device_event_type);
++    OskDeviceEvent *ev;
++    PyGILState_STATE gstate = PyGILState_Ensure();
++    ev = PyObject_New(OskDeviceEvent, &osk_device_event_type);
+     if (ev)
+     {
+         osk_device_event_type.tp_init((PyObject*) ev, NULL, NULL);
+-        return ev;
+     }
+-    return NULL;
++    PyGILState_Release(gstate);
++    return ev;
+ }
+ 
+ static PyObject *
+@@ -334,6 +336,7 @@ osk_devices_dealloc (OskDevices *dev)
+ static void
+ queue_event (OskDevices* dev, OskDeviceEvent* event, Bool discard_pending)
+ {
++    PyGILState_STATE state = PyGILState_Ensure ();
+     GQueue* queue = dev->event_queue;
+     if (queue)
+     {
+@@ -364,6 +367,7 @@ queue_event (OskDevices* dev, OskDeviceE
+         Py_INCREF(event);
+         g_queue_push_head(queue, event);
+     }
++    PyGILState_Release (state);
+ }
+ 
+ static gboolean idle_process_event_queue (OskDevices* dev)
--- a/recipes-support/onboard/onboard_1.4.1.bb
+++ b/recipes-support/onboard/onboard_1.4.1.bb
@@ -1,0 +1,62 @@
+SUMMARY = "An onscreen keyboard"
+LICENSE = "GPL-3.0-only"
+LIC_FILES_CHKSUM = "file://COPYING.GPL3;md5=8521fa4dd51909b407c5150498d34f4e"
+
+DEPENDS += "gtk+3 hunspell libcanberra libxkbfile dconf python3-distutils-extra-native intltool-native"
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI = "https://launchpad.net/onboard/1.4/${PV}/+download/${BPN}-${PV}.tar.gz \
+           file://0001-pypredict-lm-Define-error-API-if-platform-does-not-h.patch \
+           file://0002-onboard-onhover-seg-fault-fix.patch \
+           file://0001-add-xfce-to-autostart-onlyshowin.patch \
+           file://01-gnome-accessibility \
+           file://NI.colors \
+           file://NI.theme \
+           file://onboard-defaults.conf \
+           "
+SRC_URI[sha256sum] = "01cae1ac5b1ef1ab985bd2d2d79ded6fc99ee04b1535cc1bb191e43a231a3865"
+
+CXXFLAGS += "-Werror=declaration-after-statement"
+
+inherit features_check setuptools3 pkgconfig gtk-icon-cache gsettings mime-xdg
+
+REQUIRED_DISTRO_FEATURES = "x11"
+
+FILES:${PN} += " \
+    ${datadir}/dbus-1 \
+    ${datadir}/icons \
+    ${datadir}/gnome-shell \
+    ${datadir}/help \
+"
+
+RDEPENDS:${PN} += " \
+    ncurses \
+    python3-dbus \
+    python3-pycairo \
+    python3-pygobject \
+"
+
+do_install:append() {
+	install -Dm 0644 ${D}${PYTHON_SITEPACKAGES_DIR}${sysconfdir}/xdg/autostart/onboard-autostart.desktop ${D}${sysconfdir}/xdg/autostart/onboard-autostart.desktop
+
+	install -Dm 0644 ${WORKDIR}/01-gnome-accessibility ${D}${sysconfdir}/dconf/db/local.d/01-gnome-accessibility
+	install -Dm 0644 ${WORKDIR}/onboard-defaults.conf ${D}${sysconfdir}/onboard/onboard-defaults.conf
+
+	install -Dm 0644 ${WORKDIR}/NI.colors ${D}${datadir}/onboard/themes/NI.colors
+	install -Dm 0644 ${WORKDIR}/NI.theme ${D}${datadir}/onboard/themes/NI.theme
+}
+
+pkg_postinst:${PN} () {
+	dconf update
+}
+
+CONFFILES:${PN}:append := " \
+	${sysconfdir}/onboard/onboard-defaults.conf \
+	${sysconfdir}/dconf/db/local.d/01-gnome-accessibility \
+"
+
+RDEPENDS:${PN}:append = " dconf"
+# Onboard uses unicode glyphs in its key_defs.xml file, which means
+# we need a font that has those glyphs present.
+RDEPENDS:${PN}:append = " ttf-dejavu-sans"
--- a/recipes-support/onboard/files/0001-add-xfce-to-autostart-onlyshowin.patch
+++ b/recipes-support/onboard/files/0001-add-xfce-to-autostart-onlyshowin.patch
@@ -1,0 +1,10 @@
+diff --git a/data/onboard-autostart.desktop.in b/data/onboard-autostart.desktop.in
+index 8fb55ac..cf022a0 100644
+--- a/data/onboard-autostart.desktop.in
++++ b/data/onboard-autostart.desktop.in
+@@ -9,4 +9,4 @@ NoDisplay=true
+ X-Ubuntu-Gettext-Domain=onboard
+ AutostartCondition=GSettings org.gnome.desktop.a11y.applications screen-keyboard-enabled
+ X-GNOME-AutoRestart=true
+-OnlyShowIn=GNOME;Unity;MATE;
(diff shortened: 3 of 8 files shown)


recipes-support/onboard/onboard_%.bbappend
```diff
@@ -0,0 +1,41 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+# file://0001-add-xfce-to-autostart-onlyshowin.patch
```
amstewart: Leftover comment?

recipes-support/onboard/files/onboard_hover_seg_fault.patch
```diff
@@ -0,0 +1,37 @@
+--- a/Onboard/osk/osk_devices.c
++++ b/Onboard/osk/osk_devices.c
```
amstewart: This .patch is missing the commit message and .patch upstream status information.

Has this patch been submitted upstream?

amstewart (changes requested):

chaitu236:
@pratheekshasn you mentioned offline that there were issues getting onboard work on cRIO and some segfault issue; have they been addressed? Is this really ready for review?

recipes-support/onboard/files/onboard_hover_seg_fault.patch
pratheekshasn: No, it hasn't been submitted upstream, I found this patch [here](https://github.com/void-linux/void-packages/commit/1be95325d320122efd5dedf7437839cfcca01f7a). The issue was reported in one of the forums, and this was the fix given, although the person hasn't submitted the fix upstream.

pratheekshasn:
> @pratheekshasn you mentioned offline that there were issues getting onboard work on cRIO and some segfault issue; have they been addressed? Is this really ready for review?

The cRIO issue can be solved by tweaking a couple of values in onboard-settings, so yes, this is ready for review.

Here are the similarities and differences between kirkstone and scarthgap for onboard:

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-IN style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr'>


Setting | Kirkstone | Scarthgap | Comments
-- | -- | -- | --
Appear   on system boot | No | No |  
Appears   immediately when process is started through CMD | Yes | No |  
Default   value of Onboard-settings: Auto show->   External keyboards-> Don’t auto-show while external keyboards are   connected | True | False |  
Appears   whenever mouse focus is on a textbox (or a typeable window) | No | Yes | Kirkstone   behaves the same as Scarthgap if the settings are made the same as scarthgap   (Don't show while external keyboards are connected)
Disappears   when mouse focus is on a non-typeable window | Yes | Yes |  
ps ax \|   grep onboard | python3   usr/bin/onboard --not-show-in=GNOME, GNOME-Classic:GNOME --startup-delay=3.0 | python3   usr/bin/onboard | Launching   the process on scarthgap with the same CMD line arguments does not result in   the same behaviour as kirkstone, i.e., it doesn't show up immediately when   the process is started through CMD, it appears only after I click on a   textbox.



</div>


</body>

</html>

recipes-support/onboard/files/onboard_hover_seg_fault.patch
amstewart: OK. Then you can reference [this styleguide](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/docs/styleguide.md#patch-files) on how to add the necessary summary and tracking information to the .patch. It's important that .patch files have good tracking information in them, so that rebasing maintainers know when to drop them from the recipe during upgrades.

recipes-support/onboard/onboard_%.bbappend
pratheekshasn: Sorry about that, I have removed this now.

recipes-support/onboard/files/onboard_hover_seg_fault.patch
pratheekshasn: I have added the commit message from the source I got the patch file, and have added upstream status information as well.

amstewart (changes requested):
Commits look good now; thanks. Just squash them down into a single commit for the pull.

What is your intention with regards to upstreaming the recipe fixes? Are you trying to upstream the meta-OE PR with yocto first?

chaitu236:
> The cRIO issue can be solved by tweaking a couple of values in onboard-settings

Can you describe what these settings are? Are these settings part of this PR?

chaitu236:
> Default value of Onboard-settings: Auto show-> External keyboards-> Don’t auto-show while external keyboards are connected

I'm a little confused here; you mention that for scarthgap the setting is False. But in `onboard-defaults.conf` there is this

```
[auto-show]
# Enable autoshow when there's no keyboard detected.
enabled=True
keyboard-device-detection-enabled=True
keyboard-device-detection-exceptions=['::noserial']
```

which seems to imply that the setting is true. This file is also exactly the same as [kirkstone's](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/recipes-support/onboard/files/onboard-defaults.conf#L1). Does that mean that despite the same settings as kirkstone onboard works differently on scarthgap and ignores these settings? Or am I misunderstanding?

pratheekshasn:
Sorry about the confusion, the default value of `Auto show-> External keyboards-> Don’t auto-show while external keyboards are connected` is true in both kirkstone and scarthgap. I double-checked this today by formatting the cRIOs and reinstalling the image and onboard on it.

When we set the value through the command line (via `gsettings set org.onboard.auto-show keyboard-device-detection-enabled false`), or through the onboard-settings UI, the values get stored in dconf's database. It doesn't affect onboard-defaults.conf.

Furthermore, we suspect that the mousetweaks warning that gets thrown might be preventing onboard from launching immediately (like in kirkstone).

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-IN style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr'>

Setting | Kirkstone | Scarthgap | Comments
-- | -- | -- | --
Appear on system boot | No | No |  
Appears immediately when process is started through CMD | Yes | No |  
Default value of Onboard-settings: Auto show-> External keyboards-> Don’t auto-show while external keyboards are connected | True | ~~False~~ True |  
Appears whenever mouse focus is on a textbox (or a typeable window) | No | Yes | Kirkstone  behaves the same as Scarthgap if the settings are made the same as  scarthgap (Don't show while external keyboards are connected)
Disappears when mouse focus is on a non-typeable window | Yes | Yes |  
ps ax \| grep onboard | python3 usr/bin/onboard --not-show-in=GNOME, GNOME-Classic:GNOME --startup-delay=3.0 | python3 usr/bin/onboard | Launching  the process on scarthgap with the same CMD line arguments does not  result in the same behaviour as kirkstone, i.e., it doesn't show up  immediately when the process is started through CMD, it appears only  after I click on a textbox.
On launch of onboard | Appears immediately, throws no warnings on the terminal | Does not appear unless we disable keyboard detection. Throws mousetweaks warnings and acpid warnings |  
On launch of onboard-settings | Throws mousetweaks warning, then launches. | Throws mousetweaks warning, then launches. |  
On close of onboard-settings | Throws "critical" warning about tab_label not being found | Closes without issues |  


</div>


</body>

</html>


I think we can still go ahead and submit this PR that makes the onboard keyboard at least appear, albeit with a workaround. For the issue of it not showing up on a cRIO target when it can still show up in a VM, it can be fixed as a separate work item. What are your thoughts, @chaitu236?

chaitu236:
> Appears immediately when process is started through CMD

If you mean launching `onboard` via terminal, I don't think we care about this use case; the only requirement is that the virtual keyboard appears when text fields are focused without launching the process manually or putting in any workarounds (i.e., like kirkstone).

> Launching the process on scarthgap with the same CMD line arguments does not result in the same behaviour as kirkstone, i.e., it doesn't show up immediately when the process is started through CMD, it appears only after I click on a textbox.

Launching onboard process from terminal is just a debugging step we used to figure out why the keyboard wasn't appearing; if everything works correctly, we shouldn't have to launch it from terminal so we don't need to care about this difference except for debugging purposes.

> Does not appear unless we disable keyboard detection. Throws mousetweaks warnings and acpid warnings

We do need to find a fix for this before we submit/upstream as we do want the keyboard to appear without user intervention or workarounds (like in kirkstone).


> I think we can still go ahead and submit this PR that makes the onboard keyboard at least appear, albeit with a workaround.

Like we discussed, it'd be better to identify the correct fix as we're not in a rush to add this package back.

recipes-support/onboard/onboard_%.bbappend
```diff
@@ -0,0 +1,43 @@
...
+	install -d ${D}${datadir}/onboard/themes
+	install -d ${D}${sysconfdir}/xdg/autostart
+
+	install -m 644 ${WORKDIR}/01-gnome-accessibility ${D}${sysconfdir}/dconf/db/local.d/
+	install -m 644 ${WORKDIR}/onboard-defaults.conf ${D}${sysconfdir}/onboard/
+
+	install -m 644 ${WORKDIR}/NI.colors ${D}${datadir}/onboard/themes/
+	install -m 644 ${WORKDIR}/NI.theme ${D}${datadir}/onboard/themes/
+
+	mv ${D}${PYTHON_SITEPACKAGES_DIR}${sysconfdir}/xdg/autostart/onboard-autostart.desktop ${D}${sysconfdir}/xdg/autostart
```
chaitu236: We want this to go in the source .bb file not the .bbappend.

chaitu236 (changes requested):

recipes-support/onboard/onboard_%.bbappend
```diff
@@ -0,0 +1,41 @@
...
+	file://onboard-defaults.conf \
+"
+
+CXXFLAGS += "-Werror=declaration-after-statement"
+
+do_install:append () {
+	install -d ${D}${sysconfdir}/dconf/db/local.d
+	install -d ${D}${sysconfdir}/onboard
+	install -d ${D}${datadir}/onboard/themes
+	install -d ${D}${sysconfdir}/xdg/autostart
```
chaitu236: I suppose this line isn't required anymore as the folder is already created in the main recipe in meta-oe

recipes-support/onboard/files/0002-onboard-onhover-seg-fault-fix.patch
```diff
@@ -0,0 +1,52 @@
...
+
+Currently, if the mouse pointer is hovered over the Onboard keyboard, the application crashes with a segmentation fault.
+This is because the new_device_event function is not acquiring the GIL before creating a new object. This patch fixes
+the issue by acquiring the GIL before creating a new object. It also acquires the GIL before queueing the event.
+This patch is taken from the fix provided in the following link:
+https://github.com/void-linux/void-packages/commit/1be95325d320122efd5dedf7437839cfcca01f7a
+
+Signed-off by: Pratheeksha S N <pratheeksha.s.n@ni.com>
+
+Upstream-Status: Not Submitted [This is being added to the meta-nilrt layer for now. If it is accepted upstream, it will be removed from here.]
```
chaitu236: Oh I forgot about this patch.. if you're planning to send this upstream along with https://github.com/ni/meta-openembedded/pull/67, you should ideally put this file in that PR so that other NI folk can review the final to-be-sent-upstream changes.

chaitu236 (changes requested):

recipes-support/onboard/files/0002-onboard-onhover-seg-fault-fix.patch
pratheekshasn: Have moved the patch to meta-openembedded now.

chaitu236: I see that the file is still here.. I suppose you're planning to remove it here once upstream accepts meta-oe changes? In that case let's keep this issue open so we don't accidentally merge this PR. (Also, the file is not usable in the PR anyway as it isn't included `SRC_URI`.

pratheekshasn: I have deleted the file now.

recipes-support/onboard/onboard_%.bbappend
pratheekshasn: Fixed.

recipes-support/onboard/onboard_1.4.1.bb
```diff
@@ -0,0 +1,62 @@
...
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI = "https://launchpad.net/onboard/1.4/${PV}/+download/${BPN}-${PV}.tar.gz \
+           file://0001-pypredict-lm-Define-error-API-if-platform-does-not-h.patch \
+	       file://0002-onboard-onhover-seg-fault-fix.patch \
+           file://0001-add-xfce-to-autostart-onlyshowin.patch \
+	       file://01-gnome-accessibility \
+	       file://NI.colors \
+	       file://NI.theme \
+	       file://onboard-defaults.conf \
```
chaitu236: Indentation not consistent in SRC_URI

pratheekshasn: Fixed this now.

chaitu236 (approved):

PR #739, opened 2024-09-24
opkg-keyrings: change version to distro version

pratheekshasn:
### Summary of Changes

The PR changes the version of the opkg-keyrings IPK so that it is shipped with the distro version.

### Justification

[#AB2346871](https://dev.azure.com/ni/DevCentral/_workitems/edit/2346871)

NILRT deployments which are very old might conceivably outlive the lifetimes of their default opkg signing keys. Eg. a NILRT 20.0 deployment with only the 2019 signing key may be deployed in 2029. NI doesn't provide an appealing user interface to manipulate the opkg keystore. But we should give customers a pathway to upgrade their signing keys which is more easy than hacking newer OE feeds onto their device.

As part of this, the opkg-keyrings IPK will now be versioned with the distro version. This way, the newest release can later be copied into the dist/feed, with which customers can upgrade to the latest signing keys.

### Testing

Tested that the built IPK was versioned with the distro version, i.e., `opkg-keyrings_10.4-r0.0_core2-64.ipk`
Screenshot from 2024-09-24 18-15-15

### Procedure

### DIFF
--- a/recipes-devtools/opkg/opkg-keyrings_%.bbappend
+++ b/recipes-devtools/opkg/opkg-keyrings_%.bbappend
@@ -1,5 +1,7 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
 
+PV = "${DISTRO_VERSION}"
+
 SRC_URI += " \
 	file://nilrt-feed-2019.gpg \
 	file://nilrt-feed-2023.gpg \


amstewart:
Looks good to me. :+1:

chaitu236 (approved):

chaitu236:
Cherry-picked to `nilrt/master/next` - ff625b717bdfc6d02058f90b8faa4624d23ddc11

PR #747, opened 2024-10-03
fw_printenv: Don't escape the - character

erickshepherdNI:
### Summary of Changes

Removed the escape character before `-`.

### Justification

 This fixes this grep warning we were seeing on boot: `grep: warning: stray \ before -`. This issue was likely caused by a grep upgrade that made `-` no longer need to be escaped. The related bug is here: https://dev.azure.com/ni/DevCentral/_workitems/edit/2854653

### Testing

I tested this locally on a system by manually editing fw_printenv on a system and confirming that fixed the grep warnings in the boot log. I still need to build the core package feed and possibly the base image for testing but will be out the rest of the week. I'll test this as soon as I'm back.

### Procedure

### DIFF
--- a/recipes-ni/ni-utils/files/x64/fw_printenv
+++ b/recipes-ni/ni-utils/files/x64/fw_printenv
@@ -235,7 +235,7 @@ parse_opt(){
 
 	#check if there are other options passed incorrectly
 	shift $(($OPTIND - 1))
-	if [ "x$(echo $* | grep ' \-')" != "x" ]; then
+	if [ "x$(echo $* | grep ' -')" != "x" ]; then
 		echo "$0:  invalid syntax" >&2
 		echo "Try '$0 --help' for more information"
 		exit 1


chaitu236 (changes requested):
The change itself looks fine but can you please add some details on why we're making this change in the body of the commit message?

Also, shouldn't this PR target `nirlt/master/next`?
And does this need to be cherry-picked to kirkstone branch?

recipes-ni/ni-utils/files/x64/fw_printenv
```diff
@@ -235,7 +235,7 @@ parse_opt(){
 
 	#check if there are other options passed incorrectly
 	shift $(($OPTIND - 1))
-	if [ "x$(echo $* | grep ' \-')" != "x" ]; then
+	if [ "x$(echo $* | grep ' -')" != "x" ]; then
```
chaitu236, line 238: I assume you already tested that the behavior of new grep without the escape character is same as old grep with escape character?

PR #750, opened 2024-10-08
fw_printenv: Don't escape the - character

erickshepherdNI:
### Summary of Changes

Removed the escape character before `-`.

### Justification

 This fixes this grep warning we were seeing on boot: `grep: warning: stray \ before -`. This issue was likely caused by a grep upgrade that made `-` no longer need to be escaped. The related bug is here: https://dev.azure.com/ni/DevCentral/_workitems/edit/2854653

### Testing

I built the core package feed and base image, formatted a machine with the built image, and confirmed that there were no grep warnings in the boot log.

I also tested the grep command on a NILRT 11.0 system to make sure the `-` character really doesn't need to be escaped to still function correctly. The result was that the output of the grep command was the same regardless of the escape character but threw a warning when it was present.
grep test

### Procedure

### DIFF
--- a/recipes-ni/fw-printenv/files/fw_printenv
+++ b/recipes-ni/fw-printenv/files/fw_printenv
@@ -211,7 +211,7 @@ parse_opt(){
 
 	#check if there are other options passed incorrectly
 	shift $(($OPTIND - 1))
-	if [ "x$(echo $* | grep ' \-')" != "x" ]; then
+	if [ "x$(echo $* | grep ' -')" != "x" ]; then
 		echo "$0:  invalid syntax" >&2
 		echo "Try '$0 --help' for more information"
 		exit 1


erickshepherdNI:
Previous PR and discussion is here: https://github.com/ni/meta-nilrt/pull/747

erickshepherdNI:
@chaitu236 To answer questions from the previous PR, I don't think this needs to be cherrypicked into kirkstone since the bug was introduced with NILRT 11.0, but I can try to reproduce the bug on kirkstone to verify. I still need to test the new grep behavior but will update the PR with my findings.

chaitu236:
> Previous PR and discussion is here: #747

It is preferable to keep using the same PR to preserve comment history, etc. If you want to change base branch or modify commits, there is no need to create a new PR. Existing PR https://github.com/ni/meta-nilrt/pull/747 could've been modified. To change base branch - [ref](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request). And to update commits, just force push to your dev branch.

chaitu236:
> @chaitu236 To answer questions from the previous PR, I don't think this needs to be cherrypicked into kirkstone since the bug was introduced with NILRT 11.0, but I can try to reproduce the bug on kirkstone to verify. I still need to test the new grep behavior but will update the PR with my findings.

FYI changes to PR description don't trigger a notification to reviewers. Best way to let reviewers know is to request a review (or re-review) when all pending items are addressed or add a comment in the PR.

chaitu236 (approved):

PR #752, opened 2024-10-15
nilrt.inc: Change PACKAGE_FEED_* to default assignment

rtollert:
Locally-hosted package repositories are a longstanding NI Linux RT development pattern. While these can be trivially configured in /etc/opkg/ after installation, being able to customize the feed URLs in the base system image itself unlocks some unique automation possibilities.

While nilrt/scripts/tests/test_nilrt-ptest.py presently implements a feed server (and rewriting the post-install opkg configuration accordingly), its implementation is tightly coupled to the needs of ptest execution. Instead, the nominal way to accomplish this should be in site.conf: set PACKAGE_FEED_URIS to point at the local server, and clear out PACKAGE_FEED_BASE_PATHS, because the "extra/" and "main/" subdirectory distinction does not exist at the level of developer bitbake runs, nor arguably should it.

That's not presently possible because PACAKGE_FEED_URIS and PACKAGE_FEED_BASE_PATHS are assigned in the usual way (`=`) in nilrt.inc, overriding anything set in nilrt.inc. If we change nilrt.inc to only set default values, i.e. `?=`, then customizing them in site.conf works. Note that almost all other variables in nilrt.inc are already default-assigned.

PACKAGE_FEED_ARCHS does not need to be customizing for this use case, but is also changed to default assignment for reasons of consistency.

### Procedure

### DIFF
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -138,9 +138,9 @@ NILRT_MAIN_FEED_VERSION            ?= "${NILRT_FEED_NAME}"
 NILRT_MACHINE_FEED_URI_nimain      ?= "${NILRT_FEEDS_URI}/${NILRT_MAIN_FEED_VERSION}/ni-main"
 
 # OE Packages with TUNEARCHes
-PACKAGE_FEED_URIS       = "${NILRT_MACHINE_FEED_URI}"
-PACKAGE_FEED_BASE_PATHS = "extra main"
-PACKAGE_FEED_ARCHS      = "all ${MACHINE} ${TUNE_PKGARCH}"
+PACKAGE_FEED_URIS       ?= "${NILRT_MACHINE_FEED_URI}"
+PACKAGE_FEED_BASE_PATHS ?= "extra main"
+PACKAGE_FEED_ARCHS      ?= "all ${MACHINE} ${TUNE_PKGARCH}"
 
 # NI Packages, which are flat feeds
 # Add additional (non-constructed) opkg sources with the pattern:


amstewart:
Looking at this some more, I wonder if your approach to overriding the feed URI only failed because you were attempting to assign to the `x64` machine-override for `PACKAGE_FEED_BASE_PATHS`.

Does it instead satisfy your usecase to override the [`NILRT_FEEDS_URI_RELEASE`](https://github.com/ni/meta-nilrt/pull/752/commits/ac8f6d8c36d1c3ff3c9f1e60eb9076f39c9f4f8d#diff-772dfa1eb773d0e45e5cd8f24d90b74f2a1263891162dd37eb817320d195d88cR133) in your `local.conf`/`site.conf`? If so, I think we should recommend that users use that approach.

rtollert:
@amstewart 

> Looking at this some more, I wonder if your approach to overriding the feed URI only failed because you were attempting to assign to the x64 machine-override for PACKAGE_FEED_BASE_PATHS.

I'm not sure I understand what you're asking; I think this sounds backwards to me. I never attempted to set `PACKAGE_FEED_BASE_PATHS:x64`, I merely attempted to change it. But in hindsight... I guess that instead of changing nilrt.inc to set `PACAKGE_FEED_BASE_PATHS ?= …` instead of `PACAKGE_FEED_BASE_PATHS = …` ... I could set `PACAKGE_FEED_BASE_PATHS:x64 = …` in site.conf? But that sounds like an inferior solution to me.

> Does it instead satisfy your usecase to override the NILRT_FEEDS_URI_RELEASE in your local.conf/site.conf? If so, I think we should recommend that users use that approach.

No, that can't possibly work, because `NILRT_MACHINE_FEED_URI:x64 = "${NILRT_FEEDS_URI_RELEASE}/x64"`, and the `x64/` is wholly artificial (it's not the `MACHINE`!—which coincidentally happens to also be equal to `x64`).

amstewart (approved):

amstewart:
Cherry-picked to nilrt/master/next.

PR #758, opened 2024-10-28
pkggrp-ni-base: remove opkg-keyrings

rajendra-desai-ni:
### Summary of Changes

Removed opkg-keyrings from packagegroup-ni-base.

### Justification

opkg-keyrings recipe would be building separately as a new subcomponent in the ni-central build and would get published into dist feeds instead of oe-core feeds. This is needed so that the NILRT users can upgrade their keystore and need not depend on the default version deployed with the BSI.
[AB#2346871](https://dev.azure.com/ni/DevCentral/_workitems/edit/2346871/)

### Testing
None

### Procedure

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-base.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-base.bb
@@ -83,7 +83,6 @@ RDEPENDS:${PN} += "\
 	openssh-sshd \
 	openvpn \
 	opkg \
-	opkg-keyrings \
 	os-release \
 	pigz \
 	pstore-save \


recipes-core/packagegroups/packagegroup-ni-base.bb
```diff
@@ -83,7 +83,6 @@ RDEPENDS:${PN} += "\
 	openssh-sshd \
 	openvpn \
 	opkg \
-	opkg-keyrings \
```
chaitu236, line 86: Wouldn't removing it from here cause BSI and safemode to not have this installed anymore? I don't think we'd want that.
I think it's still ok to have the package in both dist feed and core feed and if there is a newer version of package in dist-feed, opkg upgrade should pick up that version - but you should test this.

rajendra-desai-ni: > Wouldn't removing it from here cause BSI and safemode to not have this installed anymore? I don't think we'd want that.

Yes, you are right. It makes sense to not remove this from here. I will abandon this PR.

> I think it's still ok to have the package in both dist feed and core feed and if there is a newer version of package in dist-feed, opkg upgrade should pick up that version - but you should test this.

Got it. I can test this functionality once the other PR gets merged. Hope this is fine?

chaitu236: You can test this without waiting on the other PR. Just host a local feed with newer version of keyrings package (i.e., make changes to keyring package and build it), add it to a target and do `opkg update; opkg upgrade` and verify that keyrings package gets upgraded.

PR #762, opened 2024-11-07
nilrt-grub-runmode: Disable Indirect Branch Tracking (IBT)

jpautler:
### Summary of Changes

The 6.6 NILRT Kernel (currently kernel-next) has picked up IBT support via upstream, which enabled it by default in 6.2. This causes a kernel panic on boot when NI kernel modules are installed because our kernel modules are not yet built with IBT support. This change turns off IBT support via a kernel command-line option.

### Justification

WI [AB#2887566](https://dev.azure.com/ni/DevCentral/_workitems/edit/2887566)

### Testing

Built nilrt-runmode-rootfs with this change and confirmed change in bootimage.cfg in nilrt-runmode-rootfs*.tar.gz
Tested change to bootimage.cfg on a target with NILRT kernel version 6.6 and confirmed kernel panic no longer occurs
[x] I have built the core package feed with this PR in place. (`bitbake packagefeed-ni-core`)

I have confirmed on the same affected test system that safemode does not load any out of tree NI drivers - so there should be no similar issue in safemode.

### Procedure

[x ] I certify that the contents of this pull request complies with the [Developer Certificate of Origin](https://github.com/ni/nilrt/blob/HEAD/docs/CONTRIBUTING.md#developer-certificate-of-origin-dco).

### Note

Cherry-pick to nilrt/master/next.

### DIFF
--- a/recipes-bsp/grub/grub/grub-runmode-bootimage.cfg
+++ b/recipes-bsp/grub/grub/grub-runmode-bootimage.cfg
@@ -3,7 +3,7 @@
 set consoleparam='console=tty0 console=ttyS0,115200n8'
 set kernel_path='/runmode/bzImage'
 
-set otherbootargs="rootwait rw usbcore.usbfs_memory_mb=0 consoleblank=0 rcu_nocbs=all "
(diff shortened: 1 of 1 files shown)


gratian (approved):

amstewart (approved):

amstewart:
Cherry-picked to /next.

PR #763, opened 2024-11-08
[next] Make the embedded UI toggle removable.

amstewart:
### Summary of Changes

1. Fixup the `sysconfig-settings-ui` `prerm` to remove any lingering symlinks.
1. Moved `sysconfig-settings-ui` to `packagegroup-ni-graphical`.
1. Syntax and styling of the `sysconfig-settings` recipe as a whole.
1. Non-functional cleanups in the `-runmode` and `-extra` packagegroups.

### Justification

This patchset makes the `sysconfig-settings-ui` package cleanly removable, which means that we can remove it from the SNAC configuration to prevent users from accidentally enabling the UI. Before this patchset, opkg would complain that removing `sysconfig-settings-ui` would break the `-base` and `-runmode` packagegroups.

[AB#2769018](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2769018)

### Testing

* [x] Built the affected packagegroups and upgraded a NILRT 25.0 runmode with their new versions. `sysconfig-settings-ui` can now be removed without complaining.
* [x] Confirmed that the UI toggle is removed in MAX when the above package is removed.
* [x] Built the runmode image and confirmed that the `sysconfig-settings-ui` package is still installed by default.

### Procedure

### DIFF
--- a/recipes-ni/sysconfig-settings/sysconfig-settings.bb
+++ b/recipes-ni/sysconfig-settings/sysconfig-settings.bb
@@ -4,119 +4,56 @@ LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 SECTION = "base"
 
+
+# RECIPE VARIABLES #
+DEPENDS += "shadow-native pseudo-native niacctbase base-files-nilrt"
+
+
+# SOURCE VARIABLES #
+SRC_URI = "\
+	file://systemsettings/consoleout.ini \
+	file://systemsettings/fpga_target.ini \
+	file://systemsettings/rt_target.ini \
+	file://systemsettings/target_common.ini \
+	file://uixml/nilinuxrt.fpga_disable.binding.xml \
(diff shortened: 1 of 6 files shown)


recipes-ni/sysconfig-settings/sysconfig-settings.bb
```diff
@@ -155,7 +92,88 @@ pkg_postinst_ontarget:${PN} () {
 
 pkg_prerm_ontarget:${PN} () {
 	rm -f ${systemsettingsdir}/target_common.ini \
```
chaitu236, line 94: Indentation doesn't match next lines; did you intend to move `${systemsettingsdir}/target_common.ini \` as well to next line like others?

chaitu236 (approved):

recipes-ni/sysconfig-settings/sysconfig-settings.bb
amstewart: The difference in indentation levels is original to the previous version of the file, and is intentional. The latter lines have an additional level of indentation because they are continuations of the `rm ...` command.

Won't fix.

PR #773, opened 2024-12-09
opkg-keyrings: Add metadata to make package visible on MAX

pratheekshasn:
### Summary of Changes

The package opkg-keyrings needs to become user visible on MAX so that the users can directly upgrade their keyrings from MAX through the dist feed.

### Justification

[#AB2866259](https://ni.visualstudio.com/DevCentral/_workitems/edit/2866259/)

### Testing

* [x] I have verified that the control file generated in the package has the metadata for UserVisible and DisplalyName under the rest of the info for opkg-keyrings.

### Procedure

### DIFF
--- a/recipes-devtools/opkg/opkg-keyrings_%.bbappend
+++ b/recipes-devtools/opkg/opkg-keyrings_%.bbappend
@@ -20,3 +20,5 @@ do_install:append() {
 	install -m 0444 ${WORKDIR}/nilrt-feed-2019.gpg ${D}${datadir}/opkg/keyrings/
 	install -m 0444 ${WORKDIR}/nilrt-feed-2023.gpg ${D}${datadir}/opkg/keyrings/
 }
+
+PACKAGE_ADD_METADATA_IPK:opkg-keyrings = "DisplayName: Opkg-Keyrings\nUserVisible: yes\n"


recipes-devtools/opkg/opkg-keyrings_%.bbappend
```diff
@@ -15,6 +15,8 @@ SRC_URI += " \
 	file://nilrt-feed-2023.gpg \
 "
 
+PACKAGE_ADD_METADATA_IPK:opkg-keyrings = "DisplayName: Opkg-Keyrings\nUserVisible: yes\n"
```
amstewart: Move this declaration below the task definition (`do_install`) below, per [the styleguide](https://github.com/ni/meta-nilrt/blob/nilrt/master/scarthgap/docs/styleguide.md#layout).

amstewart (changes requested):
Change seems fine. Could you change the commit title to be like `opkg-keyrings: ...`, and add a proper description though?

recipes-devtools/opkg/opkg-keyrings_%.bbappend
pratheekshasn: I have fixed this now.

pratheekshasn:
> Change seems fine. Could you change the commit title to be like `opkg-keyrings: ...`, and add a proper description though?

I have fixed this now.

PR #774, opened 2024-12-09
opkg-utils: Add metadata parsing functionality to opkg-utils

pratheekshasn:
### Summary of Changes

The package opkg-keyrings needs to become user visible on MAX so that the users can directly upgrade their keyrings from MAX through the dist feed. This PR adds code to opkg-utils through a patch that lets it parse metadata for `UserVisible` and `DisplayName` added to opkg-keyrings.

### Justification

[#AB2866259](https://ni.visualstudio.com/DevCentral/_workitems/edit/2866259/)

### Testing

* [x] I have verified that the package file generated has the metadata for UserVisible and DisplalyName under the rest of the info for opkg-keyrings.

### Procedure

### DIFF
--- a/recipes-devtools/opkg-utils/files/0001-parse-user_defined_fields-in-control-file.patch
+++ b/recipes-devtools/opkg-utils/files/0001-parse-user_defined_fields-in-control-file.patch
@@ -1,0 +1,27 @@
+From c58f7da958a65cfc7a38836ad0be72a6bdf62fd8 Mon Sep 17 00:00:00 2001
+From: pratheekshasn <pratheeksha.s.n@ni.com>
+Date: Tue, 10 Dec 2024 10:27:14 +0530
+Subject: [PATCH] parse user_defined_fields in control file
+
+Upstream-Status: Not Submitted [This will get submitted to the ni/nilrt repo before upstreaming it.]
+
+Signed-off by: Pratheeksha S N <pratheeksha.s.n@ni.com>
+
+---
+ opkg.py | 3 +--
+ 1 file changed, 1 insertion(+), 2 deletions(-)
+
+diff --git a/opkg.py b/opkg.py
+index e5a65dd..0e469e0 100644
+--- a/opkg.py
++++ b/opkg.py
+@@ -294,8 +294,7 @@ class Package(object):
+                 elif all_fields:
+                     self.user_defined_fields[name] = value
+                 else:
+-                    print("Lost field %s, %s" % (name,value))
+-                    pass
++                    self.user_defined_fields[name] = value
+ 
+                 if line and line[0] == '\n':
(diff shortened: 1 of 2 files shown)


recipes-devtools/opkg-utils/files/0001-add-metadata-parsing-functionality-to-Package-class.patch
```diff
@@ -0,0 +1,66 @@
...
++    
++    def set_user_visible(self, user_visible):
++        print("set_user_visible() UserVisible %s" % user_visible)
++        self.user_visible = user_visible
++    
++    def get_display_name(self):
++        return self.display_name
++    
++    def set_display_name(self, display_name):
++        self.display_name = display_name
```
amstewart: I don't think we need to continue the original file author's mistake of making explicit accessor methods for all the class attributes. You can just remove these and use the class attribute directly.

recipes-devtools/opkg-utils/files/0001-add-metadata-parsing-functionality-to-Package-class.patch
```diff
@@ -0,0 +1,66 @@
...
+         self.user_defined_fields = collections.OrderedDict()
+         if fn:
+@@ -291,6 +293,10 @@ class Package(object):
+                     self.sha256 = value
+                 elif name_lowercase in self.__dict__:
+                     self.__dict__[name_lowercase] = value
++                elif name_lowercase == 'uservisible':
++                    self.user_visible = value
++                elif name_lowercase == 'displayname':
++                    self.display_name = value
```
amstewart: Rather than carry an out-of-tree patch for this change, I imagine you could implement it as a list of custom fields that you can define at runtime. And then we can upstream this patch. (I am the upstream maintainer.)

amstewart (changes requested):
See below. Also, check your commit summary line format.

recipes-devtools/opkg-utils/files/0001-add-metadata-parsing-functionality-to-Package-class.patch
pratheekshasn: I have removed this now.

recipes-devtools/opkg-utils/files/0001-add-metadata-parsing-functionality-to-Package-class.patch
pratheekshasn: I have fixed this now.

pratheekshasn:
> See below. Also, check your commit summary line format.

I have fixed the commit message now.

amstewart:
Merged #776. Closing this version.
@pratheekshasn

PR #783, opened 2024-12-12
libfmi: remove static linkage of expat library

pratheekshasn:
### Summary of Changes

libfmi would link statically link to a snapshot of the expat library present in the libfmi repo. This patch removes that dependency so that libfmi can dynamically link to libexpat.

I have recreated the patch file from this [previous](https://github.com/ni/meta-nilrt/blob/nilrt/master/kirkstone/recipes-devtools/libfmi/libfmi/0001-fmixml-use-system-expat-instead-of-building-own-with.patch) patch. This patch can no longer be applied to the latest source of libfmi, so I have manually made the changes present in the patch and created a new patch.

### Justification

[AB#2950874](https://dev.azure.com/ni/DevCentral/_workitems/edit/2950874?src=WorkItemMention&src-action=artifact_link)

### Testing

TODO: Detail what testing has been done to ensure this submission meets requirements.

I have built the BSI with the patch file changes applied to the libfmi source, and these are the versions present on the target:
image

I have also unzipped the libfmi-src IPK and checked the files, it no longer has expat inside the source.

### Procedure

### DIFF
--- a/recipes-devtools/libfmi/libfmi/0001-Remove-expat-static-linking.patch
+++ b/recipes-devtools/libfmi/libfmi/0001-Remove-expat-static-linking.patch
@@ -1,0 +1,139 @@
+From 6f8e979b422b9eb288e484305a537f552339c0dc Mon Sep 17 00:00:00 2001
+From: pratheekshasn <pratheeksha.s.n@ni.com>
+Date: Fri, 13 Dec 2024 09:44:53 +0530
+Subject: [PATCH] Remove expat static linking
+
+libfmi is currently statically linking to expat. This patch removes that dependency and links to expat dynamically,
+since historically this is what has been followed. The expat source code contained in the libfmi source is older,
+and does not have fixes for certain CVEs, and our CVE reporting was unable to detect these issues as the expat build
+was hidden in the libfmi build. This patch removes the expat build from libfmi and links to the system expat library.
+
+The upstream libfmi has been updated to dynamically link to expat, and this patch will be removed when the libfmi
+version is updated to the latest version.
+
+Upstream-Status: Inappropriate [Temporary patch which will go away when libfmi is updated to the latest version]
+
+Signed-off-by: pratheekshasn <pratheeksha.s.n@ni.com>
+---
+ CMakeLists.txt            |  2 +-
+ Config.cmake/fmixml.cmake | 87 +--------------------------------------
+ 2 files changed, 2 insertions(+), 87 deletions(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 93c8c21..dbaeb4c 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -288,7 +288,7 @@ if(FMILIB_BUILD_TESTS)
+     )
+ endif()
+ 
+-set(FMILIB_SHARED_SUBLIBS fmixml fmizip fmicapi expat minizip zlib c99snprintf)
++set(FMILIB_SHARED_SUBLIBS fmixml fmizip fmicapi minizip zlib c99snprintf)
+ set(FMILIB_SUBLIBS fmiimport jmutils ${FMILIB_SHARED_SUBLIBS})
+ # XXX: Unclear why we create intermediate sublibs instead of just giving everything as source.
+ # Probably also wouldn't get problems with using PRIVATE/PUBLIC on linking sublibs on Windows.
+diff --git a/Config.cmake/fmixml.cmake b/Config.cmake/fmixml.cmake
+index 5cb743f..93e829e 100644
+--- a/Config.cmake/fmixml.cmake
++++ b/Config.cmake/fmixml.cmake
+@@ -101,7 +101,6 @@ endif()
+ 
+ # set(DOXYFILE_EXTRA_SOURCES "${DOXYFILE_EXTRA_SOURCES} \"${FMIXMLDIR}/include\"")
+ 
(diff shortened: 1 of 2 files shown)


recipes-devtools/libfmi/libfmi/0001-remove-expat-static-link-dependency-from-libfmi.patch
```diff
@@ -0,0 +1,133 @@
+From b85dee1d652db0b8a6b2430667ce702186253c41 Mon Sep 17 00:00:00 2001
+From: pratheekshasn <pratheeksha.s.n@ni.com>
+Date: Thu, 12 Dec 2024 16:06:06 +0530
+Subject: [PATCH] remove expat static link dependency from libfmi
+
+Signed-off-by: pratheekshasn <pratheeksha.s.n@ni.com>
```
amstewart: Your .patch file should have a complete commit summary, as though it were a patch to upstream. And it should have an Upstream-Status line to help inform the maintainers on how to handle the patch during rebases.

https://github.com/ni/meta-nilrt/blob/nilrt/master/next/docs/styleguide.md#patch-files

amstewart (changes requested):
The previous .patch was marked as `Pending` upstream. What happened to that change and can we submit this upstream?

recipes-devtools/libfmi/libfmi/0001-remove-expat-static-link-dependency-from-libfmi.patch
jpautler, line 1: I think you can remove even more expat-related content from Config.cmake/fmixml.cmake. Here is the additional content I think you can remove:
```
# XXX: Maybe we could use FetchContent to find targets of expat? Then we hopefully
# wouldn't need below workarounds for guessing expatlib's location and name.
# Requires CMake 3.16 though.

if(MSVC)
    # Expat uses special naming with MSVC, which is mirrored here.
    set(EXPAT_LIB_PREFIX lib)
else()
    set(EXPAT_LIB_PREFIX ${CMAKE_STATIC_LIBRARY_PREFIX})
endif()

if("${CMAKE_CFG_INTDIR}" STREQUAL ".")
    # Ninja complains about 'ExpatEx/./libexpat.a' otherwise. Probably because
    # generator expressions in mergestaticlibs give slighlty different paths.
    set(expatlib "${CMAKE_BINARY_DIR}/ExpatEx/${EXPAT_LIB_PREFIX}expat${CMAKE_STATIC_LIBRARY_SUFFIX}")
else()
    set(expatlib "${CMAKE_BINARY_DIR}/ExpatEx/${CMAKE_CFG_INTDIR}/${EXPAT_LIB_PREFIX}expat${CMAKE_STATIC_LIBRARY_SUFFIX}")
endif()

# Workaround to make it explicit that target 'expatex' produces 'expatlib'. (Ninja complains otherwise.)
add_custom_command(
    OUTPUT "${expatlib}"
    DEPENDS expatex
)
add_custom_target(tmp_expatlib DEPENDS ${expatlib})
```

Doing this would make your patch more similar to this upstream commit (which is not yet included in the version of fmi-library we are building in scarthgap) where they have added support to fmi-library to use an external system copy of expat:
https://github.com/modelon-community/fmi-library/commit/e871e37afc92f8864ab082cfbe0cfa56c4000540#diff-c215b9faa57d7fffad7b40f69fad059471d660e881f369ba237be17bb2ca3f66

You will also notice that commit has a couple other differences from your patch.
1) They don't add both of these lines like you do:
```
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(EXPAT REQUIRED expat)
```
instead they just add this single line:
```
find_package(EXPAT REQUIRED)
```
2) They also do not bother to modify target_link_libraries and target_include_directories.

I'm not certain whether you need to make change for these two differences, but you might investigate/test and see if those seem more correct to you.

recipes-devtools/libfmi/libfmi/0001-remove-expat-static-link-dependency-from-libfmi.patch
jpautler: I agree that having a more complete commit message describing that we prefer to use the newer installed NILRT expat, as we have done in previous releases, is a good idea. Could also mention that the expat source code contained in the fmi-library source is older, with some unpatched CVEs, and that our CVE reporting was not finding them because the build of expat was hidden inside the fmi-library build.

Regarding the upstream status, I suspect we will never want to upstream this patch. It seems that fmi-library upstream already has [a newer commit (which we are not yet including) that adds support for an external expat library](https://github.com/modelon-community/fmi-library/commit/e871e37afc92f8864ab082cfbe0cfa56c4000540#diff-c215b9faa57d7fffad7b40f69fad059471d660e881f369ba237be17bb2ca3f66). We should get that commit in the future, at which point this patch will be unnecessary. Arguably, we could base this patch on that commit now instead of what we have already, although I'll leave that up to Alex to comment. Since we expect to drop this patch next time we update libfmi, we might not care too much.

pratheekshasn: I have made the required changes now.

recipes-devtools/libfmi/libfmi/0001-remove-expat-static-link-dependency-from-libfmi.patch
pratheekshasn: I have made the required changes.

amstewart (approved):

PR #795, opened 2024-12-30
nilrt-base-system-image.postinst: Upgrade opkg-keyrings before updating other opkg sources

pratheekshasn:
### Summary of Changes

The BSI postinst script previously upgraded opkg-keyrings (to get the newly added signing key) after running the command `opkg update`, which in turn verifies the validity of all the feeds accessible by opkg.

This PR changes the order in which these commands are run because:
1. The dist feed (where opkg-keyrings is present) does not require another update, since it is already signed with the previous public key.
2. The validation of the other feeds would fail if they are already signed with the new key, and the target hasn't obtained the new key yet. This can be fixed by obtaining the new key first, i.e., upgrading opkg-keyrings.

### Justification

Although the postinst script does not throw any errors, upgrading the keyring before validating the other opkg sources makes more sense.

### Testing

Tested by building the base system image and installing on a target. Made sure that the opkg-keyrings gets updated before the other opkg sources are fetched and validated.

### Procedure

### DIFF
--- a/recipes-core/images/files/nilrt-base-system-image.postinst
+++ b/recipes-core/images/files/nilrt-base-system-image.postinst
@@ -113,4 +113,6 @@ done
 
 # Upgrade opkg-keyrings so that users can get the latest signing keys
 # OR the command with true since the postinst script shouldn't return an error if the opkg update + upgrade fails.
-opkg -o /mnt/userfs/ update && opkg -o /mnt/userfs/ upgrade opkg-keyrings || true
+# Upgrade the opkg-keyrings first to avoid the "unknown key" error when updating the package list. Since the
+# opkg-keyrings IPK is present in the dist feed, and signed with the previous key, the keyring upgrade will not fail.
+opkg -o /mnt/userfs/ upgrade opkg-keyrings && opkg -o /mnt/userfs/ update || true


prasanna-nib (approved):

chaitu236 (changes requested):
NAK.

> The dist feed (where opkg-keyrings is present) does not require another update, since it is already signed with the previous public key.


> The validation of the other feeds would fail if they are already signed with the new key, and the target hasn't obtained the new key yet. This can be fixed by obtaining the new key first, i.e., upgrading opkg-keyrings.

I'm not sure this does what you're trying to accomplish.

First, I'm not sure if opkg would allow installation/upgrades of packages if the underlying signing key changed even if we don't run `opkg update` first. Can you provide more information on what and how exactly you tested?

Also, `opkg update` updates target's idea of what the feed actually looks like. If we don't run that before doing `opkg upgrade`, opkg may try to fetch old packages which may no longer be present in the feed. Even if the older package is present in the feed, it will be the old opkg-keyrings version that doesn't have the new key; so I don't see how this is useful.

pratheekshasn:
Right, that makes sense. But I thought we might need to do this since without the new signing key, running an update results in a failure for all the other feeds which are signed with the new key.

chaitu236:
@pratheekshasn Has the BTR tests issue that prompted this PR been addressed? Can we discard this PR?

pratheekshasn:
> @pratheekshasn Has the BTR tests issue that prompted this PR been addressed? Can we discard this PR?

Yes, they have closed it, I will be closing this PR.

PR #805, opened 2025-03-05
fs_permissions_diff: Strip commit hash from file name

pratheekshasn:
### Summary of Changes

While comparing files across different builds, the version commit hashes in the file names are not currently stripped off before comparing.

This PR adds the extra regex that matches the commit hashes in the file names.

### Justification

[AB#3014543](https://dev.azure.com/ni/DevCentral/_workitems/edit/3014543)

### Testing

* [x] I have built and installed the new changes and tested on a target, and the logs added because of the difference in file names caused by the commit hash are no longer present.
* [x] The test has been run on the current kernel and the next kernel.

### Procedure

### DIFF
--- a/recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
+++ b/recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
@@ -207,7 +207,7 @@ def strip_versions_from_path(path):
     stripped_path = re.sub(r"salt-\d+\.\d+_\d+_g[0-9a-f]+", "salt-SALTVERSION", path)
     stripped_path = re.sub(r"salt-\d+\.\d+\+\d+\.g[0-9a-f]+", "salt-SALTVERSION", stripped_path)
     # Strip everything that looks like an RT kernel version (so module and kernel version changes are ignored)
-    stripped_path = re.sub(r"\d+\.\d+\.\d+-rt\d+", "VERSION.VERSION.VERSION-rtVERSION", stripped_path)
+    stripped_path = re.sub(r"\d+\.\d+\.\d+-rt\d+(-next)?(-g[a-f0-9]+)?", "VERSION.VERSION.VERSION-rtVERSION", stripped_path)
     # Strip everything that looks like 1.2.3
     stripped_path = re.sub(r"\d+\.\d+\.\d+", "VERSION.VERSION.VERSION", stripped_path)
     # Strip everything that looks like 1.2


recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -207,7 +207,7 @@ def strip_versions_from_path(path):
     stripped_path = re.sub(r"salt-\d+\.\d+_\d+_g[0-9a-f]+", "salt-SALTVERSION", path)
     stripped_path = re.sub(r"salt-\d+\.\d+\+\d+\.g[0-9a-f]+", "salt-SALTVERSION", stripped_path)
     # Strip everything that looks like an RT kernel version (so module and kernel version changes are ignored)
-    stripped_path = re.sub(r"\d+\.\d+\.\d+-rt\d+", "VERSION.VERSION.VERSION-rtVERSION", stripped_path)
+    stripped_path = re.sub(r"\d+\.\d+\.\d+-rt\d+-?(next)?-?g?[0-9a-f]*", "VERSION.VERSION.VERSION-rtVERSION", stripped_path)
```
chaitu236: Isn't following slightly better?

r"\d+\.\d+\.\d+-rt\d+(-next)?(-g[a-f0-9]+)?" [source: copilot]

chaitu236 (commented):
Did you test both current and -next kernels?

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
pratheekshasn: I have fixed this now.

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
```diff
@@ -207,7 +207,7 @@ def strip_versions_from_path(path):
     stripped_path = re.sub(r"salt-\d+\.\d+_\d+_g[0-9a-f]+", "salt-SALTVERSION", path)
     stripped_path = re.sub(r"salt-\d+\.\d+\+\d+\.g[0-9a-f]+", "salt-SALTVERSION", stripped_path)
     # Strip everything that looks like an RT kernel version (so module and kernel version changes are ignored)
-    stripped_path = re.sub(r"\d+\.\d+\.\d+-rt\d+", "VERSION.VERSION.VERSION-rtVERSION", stripped_path)
+    stripped_path = re.sub(r"\d+\.\d+\.\d+-rt\d+(-next)?(-g[a-f0-9]+)?", "VERSION.VERSION.VERSION-rtVERSION", stripped_path)
```
chaitu236, line 210: I think it's better to have two regexes - one for current kernel and one for -next kernel; and not strip the "-next" string in the next kernel one. This way there won't be collisions in stripped paths when the -next kernel is same version as current kernel.

chaitu236 (changes requested):

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
pratheekshasn: I have changed this now, and have tested it on both the current kernel and the next kernel.

chaitu236: Since we have separate lines for -next and current, the next's regex should not have `(-next)` as optional and replaced string should also have -next. i.e., something like (not tested by me)

`re.sub(r"\d+\.\d+\.\d+-rt\d+(-next)(-g[a-f0-9]+)?", "VERSION.VERSION.VERSION-rtVERSION-next", stripped_path)`

chaitu236 (changes requested):

recipes-ni/ni-base-system-image-tests/files/fs_permissions_diff.py
pratheekshasn: When we compare the current kernel files against the next kernel, this would show up as a diff where the only difference would be the addition of the word "-next", it would just cause noise in the test log.

chaitu236: I hadn't realized that we can sometimes compare -next kernels with current kernels. Thanks for catching that

chaitu236 (approved):

PR #813, opened 2025-03-20
kernel_dmesg_diff: Fix OS version passed to fetch the dmesg_record

pratheekshasn:
### Summary of Changes

The os_version would get set to a string type when fetched from the current_dmesg_record, and the string would not have the property 'major_minor', like the object of OsVersion. The major_minor string value is necessary when fetching the previous_dmesg_record, while the whole OsVersion object instance is required when uploading the log to the database.

This PR fixes what was passed in PR 806.

Additionally, [AB#3014551](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3014551) reports the error "fail to initialize ptp_kvm". But rather than an error, this seems to be an entry in the previous_dmesg_log itself, and is getting reported as a difference. Specifying the basis_log_db_date fixes the issue.

I will be creating a follow-up PR that changes the event variables to pass the basis_log_db_date.

### Justification

[AB#3014551](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3014551)

### Testing

Tested by running the new code on a target, and specifying the basis_log_db_date.

I have verified that we get the "fail to initialize ptp_kvm" error with and without specifying the current_log_db_date, and that the error is absent when we specify the basis_log_db_date.

### Procedure

### DIFF
--- a/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
+++ b/recipes-kernel/kernel-tests/kernel-tests-files/kernel_dmesg_diff.py
@@ -302,12 +302,13 @@ if args.current_log_db_date:
 
     current_dmesg_log = strip_headers(current_dmesg_record['dmesg_log'])
     kernel_version = KernelVersion(current_dmesg_record['kernel_version_full'])
-    os_version = current_dmesg_record['os_version_major_minor']
+    os_version_major_minor = current_dmesg_record['os_version_major_minor']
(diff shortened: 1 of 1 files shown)


chaitu236 (approved):

chaitu236:
Cherry-picked to `nilrt/25.3/scarthgap` - f97457d4eaa6fd926b1b68dd306f943def254a03

PR #827, opened 2025-04-24
busybox: Enable ifplugd networking for systemd

usercw88:
### Summary of Changes

Disable `systemd-networkd` on systemd image
Update `busybox-ifplugd` networking to function on systemd

### Justification

[AB#2573001](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2573001)
RTOS Decision: Continue to use busybox-ifplugd due to the potential of NI products, scripts, etc that relate to ifplugd networking

### Testing

* [x] I have built the index package feed with this PR in place. (`bitbake package-index`)
* [x] I have built the safemode images with this PR in place. (`bitbake nilrt-safemode-rootfs`)
* [x] I have applied the systemd nilrt-base-system-image on top of a sysv nilrt-safemode-rootfs

### Procedure

### DIFF
--- a/recipes-core/busybox/busybox_1.%.bbappend
+++ b/recipes-core/busybox/busybox_1.%.bbappend
@@ -4,11 +4,13 @@ SRC_URI =+ " \
             file://busybox-ifplugd \
             file://ifplugd.action \
             file://ifplugd.conf \
+            file://busybox-ifplugd.service \
             file://zcip.script \
             file://busybox-acpid \
             file://acpid.conf \
             file://acpid_poweroff.sh \
             file://acpid-logrotate.conf \
+            file://busybox-acpid.service \
             file://0001-ifplugd.c-Increase-buffer-size-for-netlink-binding.patch \
             file://zcip-allow-action-script-to-reject-chosen-IP.patch \
             file://login-utilities.cfg \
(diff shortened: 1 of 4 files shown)


recipes-core/busybox/busybox_1.%.bbappend
```diff
@@ -30,15 +32,36 @@ PACKAGES =+ " ${PN}-zcip"
...
 
-FILES:${PN}-ifplugd = "${sysconfdir}/init.d/busybox-ifplugd ${sysconfdir}/ifplugd/ifplugd.action ${sysconfdir}/ifplugd/ifplugd.conf"
-FILES:${PN}-acpid = "${sysconfdir}/init.d/busybox-acpid ${sysconfdir}/acpid.conf ${sysconfdir}/acpi ${sysconfdir}/acpi/poweroff.sh"
+script_location = "${@bb.utils.contains('INIT_MANAGER','sysvinit','${sysconfdir}/init.d','${base_sbindir}',d)}"
+FILES:${PN}-ifplugd = " \
+	${script_location}/busybox-ifplugd \
+	${sysconfdir}/ifplugd/ifplugd.action \
+	${sysconfdir}/ifplugd/ifplugd.conf \
+	${systemd_system_unitdir}/busybox-acpid.task \
+	${@bb.utils.contains('INIT_MANAGER','systemd','${systemd_unitdir}/system/busybox-ifplugd.service','',d)} \
```
chaitu236, line 41: (Optional) There seems to be a more compact way to do this - https://github.com/ni/openembedded-core/blob/nilrt/master/scarthgap/meta/recipes-core/busybox/busybox_1.36.1.bb#L36.

You think we can use that here and in `${PN}-acpid`?

recipes-core/busybox/busybox_1.%.bbappend
```diff
@@ -30,15 +32,36 @@ PACKAGES =+ " ${PN}-zcip"
...
+	${script_location}/busybox-acpid \
+	${sysconfdir}/acpid.conf \
+	${sysconfdir}/acpi \
+	${sysconfdir}/acpi/poweroff.sh \
+	${@bb.utils.contains('INIT_MANAGER','systemd','${systemd_unitdir}/system/busybox-acpid.service','',d)} \
+"
 FILES:${PN}-zcip = "${sysconfdir}/natinst/networking/zcip.script"
 FILES:${PN}-udhcpd =+ "${sysconfdir}/udhcpd.wlan0.conf"
+FILES:${PN}-hwclock =+ "${@bb.utils.contains('INIT_MANAGER','systemd','${base_sbindir}/hwclock.sh','',d)}"
+FILES:${PN}-hwclock:remove = "${@bb.utils.contains('INIT_MANAGER','systemd','${sysconfdir}/init.d/hwclock.sh','',d)}"
```
chaitu236, line 53: There is a `do_install` step [here ](https://github.com/ni/openembedded-core/blob/nilrt/master/scarthgap/meta/recipes-core/busybox/busybox.inc#L298)that copies this file. Wouldn't that still need to be removed?
Doesn't removing it only from FILES not cause warnings/errors?

recipes-core/busybox/files/busybox-acpid.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
```
chaitu236, line 1: Do we intend to use ifplugd just as a stopgap until we validate systemd-networking with the rest of our stack or do we intend to ship it with systemd?

If the former, I see that this and the ifplugd service files are already autogenerated by systemd-sysv-generator and run on startup.. so if we just disable systemd networkd, don't things just work?

recipes-core/busybox/files/busybox-ifplugd
```diff
@@ -55,7 +55,7 @@ shift
 
 [ "x$*" != "x" ] && INTERFACES="$*"
 
-[ "x$INTERFACES" = "xauto" ] && INTERFACES="`cat /proc/net/dev | awk '{ print $1 }' | egrep '^(eth|wlan|usb)' | cut -d: -f1`"
+[ "x$INTERFACES" = "xauto" ] && INTERFACES="`cat /proc/net/dev | awk '{ print $1 }' | egrep '^(eth|wlan|usb|ens)' | cut -d: -f1`"
```
chaitu236: Just curious: which target has `ens` network name?

usercw88: Will remove: I added this in order for my VM to function correctly.

recipes-core/busybox/busybox_1.%.bbappend
usercw88: I attempted to use the `d.getVar('')` method, but this seems to only work for items in the SRC_URI section. Use throughout the other sections fail with "ParseErrors"
Note: Side by side, this only saves a couple of characters per line (4 characters)

recipes-core/busybox/busybox_1.%.bbappend
usercw88: The `do_install` step for the /etc/init.d/hwclock.sh is "disabled" for systemd -- From my investigations, the busybox-hwclock package is not produced when systemd services are introduced.

I can update the `do_install` step in this bbappend to remove the /etc/init.d/hwclock.sh (if it exists) as a "just in case" situation

recipes-core/busybox/files/busybox-acpid.service
usercw88: The goal is to to continue with ifplugd until we are confident that there are no NI software that relies on it.

Note: From my experimentation, just disabling systemd networkd does not allow for a functioning network - both the sysv-generator busybox-ifplugd and busybox-acpid do not start automatically and require extra manual intervention to get working

chaitu236 (approved):

PR #828, opened 2025-04-29
[recipes-ni] Update recipes to support systemd

usercw88:
### Summary of Changes

Updated `recipes-ni` recipes to support both systemd and sysvinit

### Justification

recipes-ni services are required for proper functionality

### Testing

* [x] I have built the index package feed with this PR in place. (`bitbake package-index`)
* [x] I have built the safemode images with this PR in place. (`bitbake nilrt-safemode-rootfs`)
* [x] I have applied the systemd nilrt-base-system-image on top of a sysv nilrt-safemode-rootfs

### Procedure

### DIFF
--- a/recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs
+++ b/recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs
@@ -7,9 +7,9 @@ error_and_die() {
     exit 1
 }
 
-FILE_PATH_VOLATILES_CONFIG="/etc/default/volatiles/00_core"
-PERSISTENT_DIR_VAR_LOG_LINE="d root root 0755 /var/log none"
-VOLATILE_SYMLINK_VAR_LOG_LINE="l root root 0755 /var/log /var/volatile/log"
+FILE_PATH_VOLATILES_CONFIG="<PATH_VOLATILES>"
+PERSISTENT_DIR_VAR_LOG_LINE="<PERSISTENT_LINE>"
+VOLATILE_SYMLINK_VAR_LOG_LINE="<VOLATILE_LINE>"
 
 ENABLE_PERSISTENT_LOGS_INI_VAL=$(/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=PersistentLogs.enabled,value="false" | tr "[:upper:]" "[:lower:]")
 
@@ -30,6 +30,11 @@ case $ENABLE_PERSISTENT_LOGS_INI_VAL in
         error_and_die "Unexpected value for PersistentLogs.enabled in /etc/natinst/share/ni-rt.ini. Expected True or False."
         ;;
 esac
+
+if [ ! -f "$FILE_PATH_VOLATILES_CONFIG" ]; then
+   [ "${VERBOSE}" != "no" ] && echo "$FILE_PATH_VOLATILES_CONFIG does not exist."
+   exit 0
+fi
 
 # This grep command looks for the exact new line.
 OUTPUT=$(grep "^$NEW_VAR_LOG_LINE$" "$FILE_PATH_VOLATILES_CONFIG" 2>&1)
--- a/recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs.service
+++ b/recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=NI Persistent Logs Configuration Utility
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/systemd/scripts/ni-configpersistentlogs
+
+[Install]
+WantedBy=multi-user.target rescue.target
--- a/recipes-ni/ni-configpersistentlogs/ni-configpersistentlogs.bb
+++ b/recipes-ni/ni-configpersistentlogs/ni-configpersistentlogs.bb
@@ -8,23 +8,48 @@ PV = "1.0"
 
 SRC_URI = "\
 	file://ni-configpersistentlogs \
+	file://ni-configpersistentlogs.service \
 "
 
 S = "${WORKDIR}"
 
+PATH_VOLATILES_SYSV="/etc/default/volatiles/00_core"
+PERSISTENT_SYSV="d root root 0755 /var/log none"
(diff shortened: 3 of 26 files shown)


recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs
```diff
@@ -32,6 +32,11 @@ case $ENABLE_PERSISTENT_LOGS_INI_VAL in
 esac
 
 # This grep command looks for the exact new line.
```
chaitu236: The comment should go before `OUTPUT=...`

recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs
```diff
@@ -32,6 +32,11 @@ case $ENABLE_PERSISTENT_LOGS_INI_VAL in
...
+if [ ! -f "$FILE_PATH_VOLATILES_CONFIG" ]; then
```
chaitu236, line 34: Why is this section needed? I see that for sysvinit we install this file in initscripts_1.0.bb - don't we have a systemd replacement that does this?

recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=NI Persistent Logs Configuration Utility
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
```
chaitu236: Why `RemainAfterExit=yes`?

recipes-ni/ni-configpersistentlogs/ni-configpersistentlogs.bb
```diff
@@ -8,23 +8,34 @@ PV = "1.0"
...
+SYSTEMD_AUTO_ENABLE:${PN} = "enable"
 
 do_install () {
-	install -d ${D}${sysconfdir}/init.d/
-	install -m 0755 ${S}/ni-configpersistentlogs ${D}${sysconfdir}/init.d/
-}
+	install -d ${D}${script_location}
+	install -m 0755 ${S}/ni-configpersistentlogs ${D}${script_location}
 
+	if ${@bb.utils.contains('INIT_MANAGER','systemd','true','false',d)}; then
```
chaitu236, line 33: Some recipes like [this](https://github.com/ni/meta-openembedded/blob/nilrt/master/scarthgap/meta-networking/recipes-support/chrony/chrony_4.5.bb#L99) are installing both sysvinit, systemd files unconditionally. Why is that? Is the build system smart enough to only place the appropriate files depending on init manager in which case we don't need the conditionals?

recipes-ni/ni-hw-scripts/ni-hw-scripts-common.bb
```diff
@@ -5,55 +5,61 @@ LICENSE = "MIT"
...
-	install -d ${D}${sysconfdir}/init.d/
+	install -d ${D}${script_location}
 
-	install -m 0755 ${S}/init.d/ni-rename-ifaces     ${D}${sysconfdir}/init.d
-	install -m 0755 ${S}/init.d/nisetserialnumber    ${D}${sysconfdir}/init.d
+	install -m 0755 ${S}/init.d/ni-rename-ifaces     ${D}${script_location}
+	install -m 0755 ${S}/init.d/nisetserialnumber    ${D}${script_location}
 }
 
-pkg_postinst:${PN} () {
```
chaitu236, line 31: It's a shame we have two initscripts in one package otherwise we could've used `INITSCRIPT_NAME` and `INITSCRIPT_PARAMS` instead of the custom `pkg_postinst` and then we wouldn't have had to replace it with `python __anonymous()`.

recipes-ni/sysconfig-settings/files/nisetembeddeduixml.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=NI Embedded UI XML Service
+DefaultDependencies=no
+Before=graphical.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/systemd/scripts/nisetembeddeduixml start
+RemainAfterExit=yes
```
chaitu236: Why `RemainAfterExit=yes` when there is no `ExecStop`? Same question in other recipes

recipes-ni/ni-configpersistentlogs/ni-configpersistentlogs.bb
usercw88: It does appear that the build system is "smart enough" to only install the files based on the init manager (ie I updated a recipe to install both sysv and systemd files, and on a build set up for systemd, the package only contained systemd files) - Given the number of recipes that have already been altered, I will update them to remove these conditionals in a different PR

recipes-ni/sysconfig-settings/files/nisetembeddeduixml.service
usercw88: `RemainAfterExit`, from my understanding, will mark the service as "active" after the "ExecStart" completes.

chaitu236: Yes, but my question is why do we need to mark the service as active?

recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs
usercw88: Fixed

recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs
usercw88: Updated the ni-configpersistentlogs script to modify the systemd version of the volatile file

recipes-ni/ni-configpersistentlogs/files/ni-configpersistentlogs.service
usercw88: Removed all `RemainAfterExit=yes` from `oneshot` services

recipes-ni/sysconfig-settings/files/nisetembeddeduixml.service
usercw88: Removed all `RemainAfterExit` from `oneshot` services

chaitu236 (approved):

PR #830, opened 2025-04-30
Add a systemWebServer workaround for systemd

erickshepherdNI:
### Summary of Changes

Creates a new systemd workaround recipe for systemWebServer that installs the .service file to /usr/lib/systemd/system. The recipe also creates a .conf file in /etc/tmpfiles.d that updates the permissions of /run/natinst and runs chmod to update the permissions of /var/local/natinst.

[AB#3103453](https://dev.azure.com/ni/DevCentral/_workitems/edit/3103453)

### Justification

These changes are required to get systemWebserver running on systemd machines. The permission changes were required because the /run directory is managed differently by systemd and the updates to /var/local/natinst were required because the default systemd permissions stopped the webserv user from writing there.

### Testing

Built the base image and applied it to a cRIO-9053. I confirmed that the systemWebServer service was running automatically as soon as the device booted and I was able to access it remotely from NI MAX. I also tested this when the device was in safemode.

### Procedure

### DIFF
--- a/recipes-ni/ni-sysd-workarounds/files/syswebserv_workaround.service
+++ b/recipes-ni/ni-sysd-workarounds/files/syswebserv_workaround.service
@@ -1,0 +1,28 @@
+[Unit]
+SourcePath=/etc/init.d/systemWebServer
+Before=multi-user.target
+Before=graphical.target
+Before=niminionagent.service
+Before=nirtmdnsd.service
+Before=niwifibledd.service
+Before=nisysmgmtmdnspublisher.service
+After=ni-rtfeatures.service
+After=nisvcloc.service
+After=niauth.service
+After=handle_cpld_ip_reset.service
+After=nisdmond.service
+
+[Service]
+Type=forking
+Restart=no
+TimeoutSec=5min
+IgnoreSIGPIPE=no
+KillMode=process
+GuessMainPID=no
+RemainAfterExit=yes
+ExecStart=/etc/init.d/systemWebServer start
+ExecStop=/etc/init.d/systemWebServer stop
+ExecReload=/etc/init.d/systemWebServer reload
+
+[Install]
+WantedBy=default.target
--- a/recipes-ni/ni-sysd-workarounds/syswebserv-workaround.bb
+++ b/recipes-ni/ni-sysd-workarounds/syswebserv-workaround.bb
@@ -1,0 +1,53 @@
+SUMMARY = "systemWebServer systemD workaround files"
+DESCRIPTION = "Workaround for systemWebServer systemD implementation"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+SECTION = "base"
+
+SRC_URI = " \
+	file://syswebserv_workaround.service \
+	file://run_natinst.conf \
+"
+
+DEPENDS += " \
+	update-rc.d-native \
+	shadow-native \
+	pseudo-native \
+"
+
+inherit systemd
+SYSTEMD_PACKAGES = "${PN}"
+SYSTEMD_SERVICE:${PN} = "syswebserv_workaround.service"
+SYSTEMD_AUTO_ENABLE:${PN} = "enable"
+
+# Create syswebserv_workaround.service and install to /usr/lib/systemd/system
+FILES:${PN} += " \
+	${systemd_system_unitdir}/syswebserv_workaround.service \
+	${sysconfdir}/tmpfiles.d/run_natinst.conf \
+"
+
+S = "${WORKDIR}"
+
+do_install () {
+	install -d ${D}${systemd_system_unitdir}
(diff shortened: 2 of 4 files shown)


recipes-ni/ni-sysd-workarounds/syswebserv-workaround.bb
```diff
@@ -0,0 +1,50 @@
...
+	shadow-native \
+	pseudo-native \
+"
+
+inherit systemd
+SYSTEMD_PACKAGES = "${PN}"
+SYSTEMD_SERVICE:${PN} = "syswebserv_workaround.service"
+SYSTEMD_AUTO_ENABLE:${PN} = "enable"
+
+# Create syswebserv_workaround.service and install to //usr/lib/systemd/system
```
usercw88:
```suggestion
# Create syswebserv_workaround.service and install to /usr/lib/systemd/system
```

usercw88 (approved):

usercw88 (changes requested):
For testing, can you build the base system image an apply it to a sysv safemode - verify the service starts automatically and works as expected

erickshepherdNI:
> For testing, can you build the base system image an apply it to a sysv safemode - verify the service starts automatically and works as expected

I confirmed that the service starts automatically after reimaging when the device was in safemode. I updated the testing section to reflect this.

usercw88 (approved):

recipes-ni/ni-sysd-workarounds/syswebserv-workaround.bb
```diff
@@ -0,0 +1,50 @@
...
+	install -m 0755 ${S}/run_natinst.conf ${D}/etc/tmpfiles.d/
+}
+
+pkg_preinst_on_target:${PN} () {
+	/etc/init.d/systemWebServer stop
+	/sbin/update-rc.d -f systemWebServer remove
+}
+
+pkg_postinst_ontarget:${PN} () {
+	chmod -R 777 /var/local/natinst
```
chaitu236: I'm a bit uneasy about giving world perms for the dir, but I suppose this is only temporary.
I see that this dir is owned by natinst on sysvinit BSI, and on systemd BSI it's owned by admin.. and I suppose that's the root cause? Do we know why the difference?

recipes-ni/ni-sysd-workarounds/files/syswebserv_workaround.service
```diff
@@ -0,0 +1,28 @@
+[Unit]
+SourcePath=/etc/init.d/systemWebServer
+Before=multi-user.target
```
chaitu236, line 3: This is quite a list of deps, and if these are all really required, we should define a [target](https://www.freedesktop.org/software/systemd/man/latest/systemd.target.html#) instead with which to synchronize the services. But I suppose we can wait until webserver team decides how they want to write the service file.

recipes-ni/ni-sysd-workarounds/syswebserv-workaround.bb
```diff
@@ -0,0 +1,50 @@
...
+"
+
+inherit systemd
+SYSTEMD_PACKAGES = "${PN}"
+SYSTEMD_SERVICE:${PN} = "syswebserv_workaround.service"
+SYSTEMD_AUTO_ENABLE:${PN} = "enable"
+
+# Create syswebserv_workaround.service and install to /usr/lib/systemd/system
+FILES:${PN} += " \
+	${systemd_unitdir}system/syswebserv_workaround.service \
```
chaitu236: Maybe just use `${systemd_system_unitdir}` instead of `${systemd_unitdir}system` everywhere in the file.

chaitu236 (changes requested):

recipes-core/images/nilrt-runmode-rootfs.bb
```diff
@@ -10,6 +10,7 @@ IMAGE_INSTALL = "\
 	dkms \
 	nilrt-grub-runmode \
 	niauth-workaround \
+	syswebserv-workaround \
```
chaitu236: Don't we also need this in safemode?

erickshepherdNI: I added the workaround recipe to the safemode IMAGE_INSTALL.

recipes-ni/ni-sysd-workarounds/syswebserv-workaround.bb
erickshepherdNI: We suspect the root cause is that there are still a number of services that have not been converted to systemd. I think Can has a set of changes that may get this fixed. Regardless, I've updated the workaround to be more precise about the permission levels it assigns.

recipes-core/images/nilrt-runmode-rootfs.bb
chaitu236: Can you move this to `packagegroup-ni-base`.. dependencies common to both safemode and runmode should just go in that.

chaitu236 (changes requested):
Just one minor thing, otherwise looks good.

chaitu236 (approved):

PR #832, opened 2025-05-08
linux-nilrt.inc: Add LICENSE to generate SBOM for kernel-modules

pratheekshasn:
### Summary of Changes

The SBOM for safemode image (SPDX for initramfs) was missing numerous kernel-module-* packages. This was because the `LICENSE` was missing from the linux-nilrt recipe, which is now included in linux-nilrt.inc.

### Justification

[AB#2931584](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2931584)

### Testing

I have checked the newly generated SPDX to make sure that the SBOMs have been generated for the kernel-module-* packages. I have also diffed the set of installed package names against the generated SPDX package names.

### Procedure

### DIFF
--- a/recipes-kernel/linux/linux-nilrt.inc
+++ b/recipes-kernel/linux/linux-nilrt.inc
@@ -41,6 +41,7 @@ SRCREV ?= "${AUTOREV}"
 PV = "${LINUX_VERSION}+git${SRCPV}"
 CVE_VERSION = "${KERNEL_VERSION}+git${SRCPV}"
 
+LICENSE = "GPL-2.0-only"
 LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
 LIC_FILES_CHKSUM:xilinx-zynq = "file://COPYING;md5=d7810fab7487fb0aad327b76f1be7cd7"
 


amstewart (approved):

recipes-kernel/linux/linux-nilrt.inc
```diff
@@ -41,6 +41,7 @@ SRCREV ?= "${AUTOREV}"
 PV = "${LINUX_VERSION}+git${SRCPV}"
 CVE_VERSION = "${KERNEL_VERSION}+git${SRCPV}"
 
+LICENSE = "GPL-2.0-only"
```
chaitu236, line 44: `recipes-kernel/linux/linux-yocto.inc` that we include above already contains this `LICENSE` line so we shouldn't require it. I ran `bitbake -e linux-nilrt` to confirm this.
So I'm not sure why we'd need this change to generate spdx for modules. Maybe it's a bug in spdx generation code instead?

pratheekshasn: Yeah, you're right. I did a clean-build, and the kernel-module-* packages get generated without my fix in the initramfs SPDX. I installed it on another newly formatted target and can see that this fix is not necessary. My python parser script had reported a discrepancy because of a mismatch in the commit hashes.

So it seems like the "fix" for the bug is more of a documentation issue where we inform them that the SBOM should be picked up from the initramfs as opposed to the safemode rootfs. Does that sound correct?

PR #834, opened 2025-05-12
linux-nilrt: enable all kernel modules

rajendra-desai-ni:
### Summary of Changes
Created a new config fragment for the default kernel and enabled all the loadable modules so that they would be available for installation using opkg.
Used ubuntu linux kernel config file as a reference to enable all the necessary loadable modules.

### Justification

[AB#2536400](https://dev.azure.com/ni/DevCentral/_workitems/edit/2536400/)

### Testing

* [x] Built the kernel image and the modules with existing changes
* [x] Installed the kernel image and the modules on a cRIO-9030 target
* [x] Verified the installation of modules using modprobe and insmod

### Procedure


gratian:
@rajendra-desai-ni Apologies for the delay. I took a fist pass reviewing this and identified some modules that do not make sense to include (like support for parallel port drivers, mtd support for chips not present on x86_64, test modules etc.). See the diff here: https://github.com/gratian/meta-nilrt/commit/6eb395899a77938769a520563b6de809580fea63

There's probably other things I've missed but let's go with this for now.

recipes-kernel/linux/linux-nilrt_git.bb
```diff
@@ -5,6 +5,10 @@ LINUX_VERSION:xilinx-zynq = "4.14"
 
 require linux-nilrt.inc
 
+SRC_URI += "\
```
chaitu236: Seeing that this `extra.cfg` is x64 specific and we'll have ARM building from same codebase soon, I think we should make it `extra_x64.cfg` or something like that and make this a `SRC_URI:append:x64`.

What do you think @gratian ?

chaitu236 (changes requested):

recipes-kernel/linux/linux-nilrt_git.bb
gratian: @chaitu236 Yes, It think that makes sense.

rajendra-desai-ni: Got it, I have renamed the file accordingly.

chaitu236 (changes requested):
> enabled all the modules as a part of config fragment so that the new
modules are available for installation as IPK packages.

Commit messages should be in imperative mood i.e., something like `Enable` instead of `Enabled` - [ref](https://www.kernel.org/doc/html/v4.12/process/submitting-patches.html).

Also, "all kernel modules" is not accurate.. maybe change it to "Enable additional modules with a config fragment.. " or something along those lines.

chaitu236 (approved):

chaitu236 (approved):

gratian (approved):

PR #838, opened 2025-05-14
linux-nilrt: Remove System.map from boot dir

erickshepherdNI:
The System.map file included in the boot directory takes up 5MB and is able to be safely removed. This commit adds a new bbappend file to remove System.map.

[AB#3112870](https://dev.azure.com/ni/DevCentral/_workitems/edit/3112870)

### Summary of Changes

Adds a new bbappend file for linux-nilrt that removes the System.map file.

### Justification

This file takes up valuable space in the boot partition, can be removed safely,  and is not required for dkms functions. There is also an identical copy of the file that is still included in the /lib/modules directory in case the file is needed for debugging.

### Testing

Built the system image and applied it to a cRIO. I also installed NI driver software, removed it, and verified that I could rebuild the packages from source with dkms.

### Procedure

### DIFF
--- a/recipes-kernel/linux/linux-nilrt.inc
+++ b/recipes-kernel/linux/linux-nilrt.inc
@@ -73,6 +73,10 @@ DEPENDS:append:armv7a = " dtc-native"
 
 RDEPENDS:kernel-dev += "bash binutils gcc gcc-symlinks"
 
+kernel_do_install:append() {
+	rm ${D}/${KERNEL_IMAGEDEST}/System.map-${KERNEL_VERSION}
+}
+
 do_shared_workdir:append() {
 	cp -rf ./* ${STAGING_KERNEL_BUILDDIR}
 }


recipes-kernel/linux/linux-nilrt_git.bbappend
```diff
@@ -0,0 +1,3 @@
+do_install:append() {
```
chaitu236: `recipes-kernel/linux/linux-nilrt_git.bbappend` will only apply this to main kernel and not the variants like debug, nohz.

Also, since `meta-nilrt` layer owns the recipes, it's odd to have a bbappend instead of changing the recipe itself.
Since all variants derive from `recipes-kernel/linux/linux-nilrt.inc`, we should be changing it instead.

I think `do_install:append()` would apply to every package built from recipe (or it would complain that no specific package was specified). I suppose [this ](https://github.com/ni/openembedded-core/blob/nilrt/master/scarthgap/meta/classes-recipe/kernel.bbclass#L494)is the where the file is added? Did you check what pacakges `kernel_do_install` applies to?

chaitu236 (changes requested):

recipes-kernel/linux/linux-nilrt_git.bbappend
erickshepherdNI: Makes sense, I've gone ahead and deleted the bbappend file, moved the change to linux-nilrt.inc, and changed the function to kernel_do_install:append(). I confirmed this properly removes the file from the boot directory.

chaitu236 (approved):

gratian (approved):

amstewart (approved):

PR #854, opened 2025-06-20
Add usbguard to SNAC package group

texasaggie97:
### Summary of Changes

This commit adds the `usbguard` package to `packagegroup-ni-snac` package group, ensuring it is included in relevant builds.

In the `packagefeed-ni-extra` group, removed `tmux` - `tmux` is also in `packagegroup-ni-snac` which ends up in core so is not needed in extra.

### Justification

[AB#3093658](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3093658)

### Testing

* [X] I have installed the newly build ipk
* [X] I have verified that both `usbguard` and `usbguard-daemon` run

### Procedure

### DIFF
--- a/recipes-core/packagegroups/packagefeed-ni-extra.bb
+++ b/recipes-core/packagegroups/packagefeed-ni-extra.bb
@@ -328,7 +328,6 @@ RDEPENDS:${PN} += "\
 	can-utils \
 	tcsh \
 	tipcutils \
-	tmux \
 	zram \
 "
 
--- a/recipes-core/packagegroups/packagegroup-ni-snac.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-snac.bb
@@ -13,5 +13,6 @@ RDEPENDS:${PN} = "\
 	nilrt-snac \
 	ntp \
 	tmux \
+	usbguard \
 	wireguard-tools \
 "


recipes-core/packagegroups/packagefeed-ni-extra.bb
```diff
@@ -358,6 +358,11 @@ RDEPENDS:${PN} += "\
 "
 # vlc, mplayer requires libmad which is on a non-whitelisted license
 
+# meta-openembedded/meta-oe/recipes-security
+RDEPENDS:${PN} += "\
+    usbguard \
```
chaitu236: `packagefeed-ni-extra` is a best effort feed and we don't provide guarantees that everything in this list (or their deps) will build or end up in feed. Is this acceptable for `usbguard`?

If not, this should be added to `packagegroup-ni-desirable` for which we provide stronger guarantees.

AlexHearnNI: Does putting it into `packagefeed-ni-extra` mean it will go into [the extra directory](https://download.ni.com/#ni-linux-rt/feeds/2025Q2/x64/extra/)? If so, then it can't go into `packagefeed-ni-extra`.

texasaggie97: I put this in the same groups as `tmux` which is also used in SNAC. @amstewart, what should I do here?

chaitu236: It appears that `packagegroup-ni-snac` is [part of](https://github.com/ni/meta-nilrt/blob/nilrt/master/scarthgap/recipes-core/packagegroups/packagefeed-ni-core.bb#L23) `packagefeed-ni-core`. So `usbguard` will end up in core feed (as it's added to `packagegroup-ni-snac.bb`) and will be excluded from extra feed even when added to it explicitly because of [this step](https://github.com/ni/nilrt/blob/nilrt/master/scarthgap/scripts/pipelines/build.extra-feeds.sh#L79) that removes packges from extra feed if they're already part of core feed.

So, adding `usbguard` to `packagefeed-ni-extra.bb` is not useful.

AlexHearnNI: Then I recommend we remove `usbguard` and `tmux` from `packagefeed-ni-extra`.

amstewart: Yes; `usbguard` only needs to be in the SNAC packagegroup.

texasaggie97: Should I remove tmux while I am at it?

chaitu236: Yes, I think it's a good idea to remove `tmux` from `packagefeed-ni-extra`
 (in a separate commit).

texasaggie97: Updated commits with these changes.

chaitu236 (approved):

amstewart (approved):

AlexHearnNI (approved):

PR #885, opened 2025-10-10
Re-add ARM specific changes in OE repos

usercw88:
### Summary of Changes

Re-add ARM specific changes that were previously removed from OE repos

### Justification

[AB#3218279](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3218279)

### Testing

* [X] I have built the core package feed, on x64, with this PR in place. (`bitbake packagefeed-ni-core`)
* [X] I have built the core package feed, on ARM, with this PR in place. (`bitbak packagefeed-ni-core`)
* [X] I have built the core base system image, on x64, with this PR in place. (`bitbake nilrt-base-system-image`)
* [X] I have built the core base system image, on ARM, with this PR in place. (`bitbake nilrt-base-system-image`)
* [X] I have installed the x64 BSI on a PXIe Target and verified device boots

### Procedure

### DIFF
--- a/recipes-core/udev/udev-extraconf.bb
+++ b/recipes-core/udev/udev-extraconf.bb
@@ -13,6 +13,8 @@ SRC_URI = " \
 	file://scripts \
 "
 
+SRC_URI:append:xilinx-zynq = "file://fpga.rules"
+
 S = "${WORKDIR}"
 
 
@@ -35,6 +37,9 @@ do_install() {
 	install -m 0755 ${S}/scripts/* ${D}${sysconfdir}/udev/scripts
 }
 
+do_install:append:xilinx-zynq() {
+	install -m 0644 ${WORKDIR}/fpga.rules          ${D}${sysconfdir}/udev/rules.d/fpga.rules
+}
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
 
--- a/recipes-core/udev/udev-extraconf/fpga.rules
+++ b/recipes-core/udev/udev-extraconf/fpga.rules
@@ -1,0 +1,2 @@
+# /dev/fpgaperipheral should be read/write to ni (0660 permissions are default)
+KERNEL=="fpgaperipheral", GROUP="ni"


recipes-ni/base-files-nilrt/base-files-nilrt.bb
```diff
@@ -31,6 +31,8 @@ FILES:${PN} += "README_File_Paths.txt \
 
 S = "${WORKDIR}"
 
+ARCH_ABI_EXT="${ABIEXTENSION}${@bb.utils.contains('TUNE_FEATURES','callconvention-hard','hf','',d)}"
```
chaitu236: > Revert "base-files-nilrt: remove ARM-specific logic"

We are no longer using hard float so this commit can be dropped.

recipes-core/initscripts/initscripts_1.0.bbappend
```diff
@@ -28,4 +28,8 @@ do_install:append() {
 	update-rc.d -f -r ${D} checkroot.sh remove
 	update-rc.d -f -r ${D} mountnfs.sh remove
 	update-rc.d -f -r ${D} read-only-rootfs-hook.sh remove
+
+	if [ "${TARGET_ARCH}" = "arm" ]; then
+		sed -i -e "s/echo \"3\"/echo \"5\"/" ${D}/${sysconfdir}/init.d/alignment.sh
```
chaitu236: > Revert "initscripts: remove ARM logic"

It appears that we have not been using this change in current sumo ARM builds as ni-central's `alignment.sh` overwrites this. Let us drop this revert.

recipes-core/udev/udev-extraconf.bb
```diff
@@ -13,6 +13,9 @@ SRC_URI = " \
 	file://scripts \
 "
 
+SRC_URI:append:xilinx-zynq = "file://fpga.rules"
+SRC_URI:append:xilinx-zynqhf = "file://fpga.rules"
```
chaitu236: > SRC_URI:append:xilinx-zynqhf = "file://fpga.rules"

This line can be dropped as we no longer use hardfloat.

recipes-kernel/perf/perf.bbappend
```diff
@@ -1,5 +0,0 @@
-
```
chaitu236, line 1: Commit https://github.com/ni/meta-nilrt/commit/fa80bd95261ed95def14877cb13c79d03fcee840 seems to suggest that this change is required on ARM as long as kernel is 4.14 which it is now.

Is that no longer the case? Do 4.14 kernel sources now support python3?

chaitu236 (changes requested):

recipes-kernel/perf/perf.bbappend
usercw88: Looking through the history of linux kernel 4.14, there are multiple references to adding support (or addressing) python3 - https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git/log/?h=v4.14-rt&qt=grep&q=python

chaitu236: Some python3 changes were from before when we introduced the commit, so it's possible there isn't full support for python3.

https://github.com/ni/meta-nilrt/pull/213 mentions this

> Without this change, building perf with MACHINE=xilinx-zynq throws a do_compile error saying that python3 is unsupported. With this patch, it builds completely.

If we can build `perf` without this change, then we can remove it.

usercw88: When building perf, the following warning is thrown:
```
Path does not exist: /git/nilrt/build/tmp-glibc/work-shared/xilinx-zynq/kernel-source/arch/arm64/tools. Maybe PERF_SRC lists more files than what your kernel version provides and needs
```
For this reason, I will remove this revert

chaitu236 (approved):

PR #890, opened 2025-10-17
lvrt cgroupsv2: Add cgroups v2 file hierarchy specific to LVRT usage

pratheekshasn:
### Summary of Changes

Since cgroups v1 is going to be deprecated soon, we need to support v2 paths.
Currently LVRT requires a specific hierarchy under `/sys/fs/cgroup` to allocate resources and tasks to specific groups. Since this is a requirement only for LVRT, and not NILRT itself, this can be pulled out of the current initscripts-nilrt that takes care of this. A new recipe for a new package has been created as part of this PR, which takes care of creating the required file hierarchy. This package will be added as a dependency to the LVRT package, so that this gets set up only when LVRT gets installed on the target.

This PR also removes the existing initscripts-nilrt that mounts v1 paths. v1 paths are set up conditionally based on what version of LabVIEW Real Time is installed.

### Justification

[AB#3271314](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3271314)

### Testing

1. Tested by installing the built sysvinit IPK on an RT target, and made sure that the files get added to `/etc/init.d.` Also made sure that the required symlinks are created in the /etc/rc*.d directories.
3. Tested that the right hierarchy is created for the `ni` subgroup under `/sys/fs/cgroup` with the right ownership.
4. Tested that the cgroup.subtree_control file has `cpuset` enabled.
5. I have tested that the right cgroup version paths are mounted as well, here's the table that shows that information:

| LVRT version | Mounted cgroup path version | Image |
|---|---|---|
| No LVRT installed | Mounts v2 paths | image |
| LVRT 26.0 installed | Mounts v2 paths | image |
| LVRT 25.0 insalled | Mounts v1 paths | image |

### Procedure


recipes-ni/cgroups-v2/files/cgroupv2-common.sh
```diff
@@ -0,0 +1,36 @@
...
+    mkdir -p "${NI_GROUP}"
+    chown lvuser:ni "${NI_GROUP}"
+
+    cd "${NI_GROUP}"
+
+    echo "+cpuset" > cgroup.subtree_control
+
+    for group in $SUBGROUPS; do
+        mkdir -p "$group"
+        chown lvuser:ni "$group"
```
gratian: Can you double check that the files under these folders also end up being owned by `lvuser:ni` or at least the ones that LabVIEW needs access to.

Otherwise we may need a recursive chown at the end of this function instead, i.e. `chown -R lvuser:ni "${NI_GROUP}"`

recipes-ni/cgroups-v2/files/cgroupv2-common.sh
```diff
@@ -0,0 +1,36 @@
...
+    done
+}
+
+assign_tasks() {
```
gratian: This could use a more descriptive name, maybe `assign_tasks_from_parent`?

Also it may be cleaner and more self contained if the folder is passed in as an argument instead of having to `cd` into it before calling this function.

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
```diff
@@ -0,0 +1,104 @@
...
+    dt = d.getVar('DISTRO_FEATURES').split()
+    if 'systemd' in dt:
+        d.setVar('INHERIT', (d.getVar('INHERIT') or '') + ' systemd update-rc.d')
+        d.setVar('SYSTEMD_AUTO_ENABLE', 'enable')
+        d.setVar('SYSTEMD_SERVICE_${PN}', 'nicreatecpuacctgroups.service nicreatecpusets.service')
+
+        # Unset SysVinit vars
+        d.setVar('INITSCRIPT_PACKAGES', '')
+        d.setVar('INITSCRIPT_NAME:${PN}-cpusets', '')
+        d.setVar('INITSCRIPT_PARAMS:${PN}-cpusets', '')
```
gratian: Still looking into this but I don't think you have to clear these variables. I believe a systemd image build would ignore them.

gratian (commented):
I still need to finish reviewing the nicreatecgroupsv2_1.bb. At a first glance it looks overly complex. I don't think you need those anonymous python functions to set/unset variables.

You should be able to just assign the vars for both sysv and systemd use-cases. Here's an example from openembedded-core for avahi: https://github.com/ni/openembedded-core/blob/fd678eb2aff91aab4cb34ed0b941d5de2cd34541/meta/recipes-connectivity/avahi/avahi_0.8.bb#L177

recipes-ni/cgroups-v2/files/cgroupv2-common.sh
pratheekshasn: Only the directories under `/sys/fs/cgroup/ni` had the owner set to `lvuser:ni` and not the files. I have changed this to `chown -R` now, and all the files now belong to `lvuser:ni` as well.

recipes-ni/cgroups-v2/files/cgroupv2-common.sh
pratheekshasn: I have fixed this now.

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
pratheekshasn: You're right, I didn't need to do this at all, I have removed this now.

recipes-ni/cgroups-v2/files/cgroupv2-common.sh
```diff
@@ -0,0 +1,41 @@
...
+    mkdir -p "${NI_GROUP}"
+    chown -R lvuser:ni "${NI_GROUP}"
+
+    cd "${NI_GROUP}"
+
+    echo "+cpuset" > cgroup.subtree_control
+
+    for group in $SUBGROUPS; do
+        mkdir -p "$group"
+        chown -R lvuser:ni "$group"
```
gratian: This call to chown could be eliminated if we move the `chown -R lvuser:ni "${NI_GROUP}"` above to the end of the function.

recipes-ni/cgroups-v2/files/createcpuacctgroups.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
gratian: I think we can drop the `createcpuacctgroups.sh` script entirely. With cgroups v2 the cpuacct is not a separate cgroup controller. We can get CPU utilization stats directly from cpu.stat which is available with cpuset.

As it stands this just ends up running the same `setup_mount_and_chown` and `assign_tasks_from_parent` like the other `createcpusets.sh` script.

Sorry, I didn't catch this on the first pass.

recipes-ni/cgroups-v2/files/createcpusets.sh
```diff
@@ -0,0 +1,30 @@
...
+# Provides:          createcpusets
+# Required-Start:    $local_fs
+# Default-Start:     1 4 5
+# Default-Stop:
+# Short-Description: Setup cpuset controller & assign tasks
+### END INIT INFO
+
+set -e
+
+. /usr/lib/nicreatecgroupsv2/cgroupv2-common.sh
```
gratian: Since it turns out the `createcpuacctgroups.sh` is not needed with cgroups v2 I think we can fold the `cgroupv2-common.sh` functions here and not have a separate file. This will end up being the only script that uses those functions.

recipes-ni/cgroups-v2/files/createcpusets.sh
```diff
@@ -0,0 +1,30 @@
...
+        cat ../cpuset.cpus > "$group/cpuset.cpus"
+    fi
+    if [ -f "../cpuset.mems" ]; then
+        cat ../cpuset.mems > "$group/cpuset.mems"
+    fi
+done
+
+assign_tasks_from_parent "${NI_GROUP}"
+
+echo "createcpusets done."
```
gratian: This print is not very descriptive. We can probably just drop it since it is not adding much value.

recipes-ni/cgroups-v2/files/nicreatecpuacctgroups
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+/usr/lib/nicreatecgroupsv2/createcpuacctgroup.sh "$@"
```
gratian: We can drop this. We cgroups v2 we do not have a cpuacct cgroup controller to worry about.

recipes-ni/cgroups-v2/files/nicreatecpusets
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+/usr/lib/nicreatecgroupsv2/createcpusets.sh "$@"
```
gratian: Right now this wrapper script just increases the execution time by spawning an new shell process w/o much benefit. If needed we can drop the `.sh` extension from the actual script this invokes.

Alternatively if the intent of this script is to be a proper SysV init script then we need to extend it to have proper 'start', 'stop', 'restart' handling.

recipes-ni/cgroups-v2/files/nicreatecpuacctgroups.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
```
gratian: We can drop this. We cgroups v2 we do not have a cpuacct cgroup controller to worry about.

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
```diff
@@ -0,0 +1,79 @@
...
+        install -Dm644 ${WORKDIR}/nicreatecpusets.service ${D}${SYSTEMD_UNITDIR}/system/nicreatecpusets.service
+    else
+        install -Dm755 ${WORKDIR}/nicreatecpuacctgroups ${D}${sysconfdir}/init.d/nicreatecpuacctgroups
+        install -Dm755 ${WORKDIR}/nicreatecpusets ${D}${sysconfdir}/init.d/nicreatecpusets
+    fi
+}
+
+RDEPENDS_${PN} += "update-rc.d"
+
+pkg_postinst:${PN} () {
```
gratian: Can you double check this postinst is actually needed? I believe the udate-rc.d class inherited above should handle this based on INITSCRIPT_* variables w/o requiring an explicit postinst.

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
```diff
@@ -0,0 +1,79 @@
...
+    if command -v systemctl >/dev/null 2>&1; then
+        # systemd: no special postinst needed for units usually
```
gratian: If the postinst is actually needed we should rewrite this 'if' condition to eliminate this empty case.

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
```diff
@@ -0,0 +1,79 @@
...
+        exit 0
+    elif [ -x /usr/sbin/update-rc.d ]; then
+        update-rc.d nicreatecpusets defaults
+        update-rc.d nicreatecpuacctgroups defaults
+    fi
+    exit 0
+}
+
+pkg_postrm_${PN} () {
```
gratian: Check if we need an explicit postrm, normally I believe it should be handled by the systemd and update-rc.d Yocto bbclases.

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
```diff
@@ -0,0 +1,79 @@
...
+SYSTEMD_SERVICE:${PN} = "nicreatecpuacctgroups.service nicreatecpusets.service"
+
+INITSCRIPT_PACKAGES = "${PN}-cpusets ${PN}-cpuacct"
+INITSCRIPT_NAME:${PN}-cpusets = "nicreatecpusets"
+INITSCRIPT_PARAMS:${PN}-cpusets = "start 90 1 4 5 . stop 90 0 6 ."
+INITSCRIPT_NAME:${PN}-cpuacct = "nicreatecpuacctgroups"
+INITSCRIPT_PARAMS:${PN}-cpuacct = "start 91 2 4 5 . stop 91 0 6 ."
+
+FILES:${PN} += " /usr/lib/nicreatecgroupsv2/cgroupv2-common.sh \
+                  /usr/lib/nicreatecgroupsv2/createcpusets.sh \
```
gratian: After we drop the cpuacct script as noted above we will end up with a single script which I think we should just install in `/usr/bin`. We can name it something more descriptive to show it's a NI script or just call it `nicreatecgroupsv2`

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
pratheekshasn: postinst seems to be required, it never created the symlinks into `/etc/rc4.d` and `/etc/rc5.d` without it.

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
pratheekshasn: Without the postrm, the initscript at `/etc/init.d` gets deleted when I uninstall the package, but the symlinks at `/etc/rc4.d/` and `/etc/rc5.d` don't. I thought it would be better not to have broken symlinks.

recipes-ni/cgroups-v2/files/cgroupv2-common.sh
pratheekshasn: Fixed.

recipes-ni/cgroups-v2/files/createcpuacctgroups.sh
pratheekshasn: Fixed.

recipes-ni/cgroups-v2/files/createcpusets.sh
pratheekshasn: Done.

recipes-ni/cgroups-v2/files/createcpusets.sh
pratheekshasn: Removed the print.

recipes-ni/cgroups-v2/files/nicreatecpuacctgroups
pratheekshasn: Removed.

recipes-ni/cgroups-v2/files/nicreatecpuacctgroups.service
pratheekshasn: Removed.

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
pratheekshasn: Rewrote the if block.

recipes-ni/cgroups-v2/nicreatecgroupsv2_1.0.bb
pratheekshasn: I have fixed this now.

gratian (approved):
I have made a couple of small tweaks, mostly white space fixes, fixing missing runtime dependencies, and handling the stop on runlevels 0 and 6.

My changes to @pratheekshasn commit are captured in this branch: https://github.com/gratian/meta-nilrt/commits/dev/pratheeksha/lvrt-cgroupsv2-tweaks/

In the interest of saving a round-trip through another PR review I have tested this changes for both cgroups v1 and v2 scenarios and pushed the amended commit as: https://github.com/ni/meta-nilrt/commit/f890c3bdbdbe477d6f9d7331f728afefbc111ec8

Also tweaked the commit message to add more details.

gratian:
Closing this PR. Changes pulled as: https://github.com/ni/meta-nilrt/commit/f890c3bdbdbe477d6f9d7331f728afefbc111ec8

PR #893, opened 2025-10-24
docker-moby: Remove cgroup-lite as a dependency

pratheekshasn:
### Summary of Changes

cgroups v1 paths will soon be deprecated, we need docker containers to be runnable on cgroup v2 paths. The
meta-virtualization recipe adds the dependency cgroup-lite for SysVInit images. Since that is specific to cgroup v1 paths, it needs to be removed from the docker-moby pacakge dependencies.

The docker-moby package is now capable of using cgroup v2 paths by adding the new recipe `ni-cgroups` as a dpeendency.

### Justification

[AB#3441396](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3441396)

### Testing

This change has been tested locally by building and installing the docker-moby package on a SysVInit target. The docker-functional-tests have also been run on the target to make sure nothing is broken.

### Procedure

### DIFF
--- a/recipes-containers/docker/docker-moby_%.bbappend
+++ b/recipes-containers/docker/docker-moby_%.bbappend
@@ -2,3 +2,7 @@
 # So use a shortened SRCREV_moby in PV.
 SRCREV_moby_short = "${@d.getVar('SRCREV_moby')[:10]}"
 PV = "${DOCKER_VERSION}+git${SRCREV_moby_short}"
+
+# Remove cgroup-lite dependency so that docker can work on systems with cgroup v2 paths.
+RDEPENDS:${PN}:remove = "cgroup-lite"
+RDEPENDS:${PN}:append = " ni-cgroups"


recipes-containers/docker/docker-moby_%.bbappend
```diff
@@ -2,3 +2,14 @@
 # So use a shortened SRCREV_moby in PV.
 SRCREV_moby_short = "${@d.getVar('SRCREV_moby')[:10]}"
 PV = "${DOCKER_VERSION}+git${SRCREV_moby_short}"
+
+# File: docker-moby_%.bbappend
+
+# Remove cgroup-lite dependency explicitly since the base recipe
+# adds it instead of appending it.
+# This is required so that docker can work on systems with cgroup v2 paths.
+RDEPENDS:${PN} = "util-linux util-linux-unshare iptables \
```
gratian: `RDEPENDS:remove` is not limited to dependencies added by `RDEPENDS:append` it also works if the parent recipe assigns RDEPENDS directly.

I've tried the following in this bbappend and it works fine:

`RDEPENDS:${PN}:remove = "cgroup-lite"`

This is better than overwriting RDEPENDS because the list of dependencies in the parent recipe may change in the future w/o us noticing.

gratian (commented):
The change itself looks good but the commit message needs to be updated since it still talks about not being able to use RDEPENDS:remove.

Also I would move the section about testing from the commit message into the PR message.

recipes-containers/docker/docker-moby_%.bbappend
pratheekshasn: I have fixed this now.

gratian (approved):
@pratheekshasn Can you mark this PR as ready for review? It is currently a draft. Looks fine otherwise.

PR #897, opened 2025-11-04
Add xorgxrdp and xrdp package builds to nilrt extra feed

jhershbe:
### Summary of Changes

Fix a few things for our context and make sure the build will support hardware acceleration of the RDP session on our controllers.

### Justification

This is generally a useful thing to be able to use for accessing your nilrt targets.

### Testing

Installed on a 9042 and verified that the RDP was functional and that the Intel GPU was accelerating the RDP session.

* [ X ] I have built the core package feed with this PR in place. (`bitbake packagefeed-ni-core`)

### Procedure

* [ X ] I certify that the contents of this pull request complies with the [Developer Certificate of Origin](https://github.com/ni/nilrt/blob/HEAD/docs/CONTRIBUTING.md#developer-certificate-of-origin-dco).

### DIFF
--- a/recipes-support/xorg-xrdp/xorgxrdp_%.bbappend
+++ b/recipes-support/xorg-xrdp/xorgxrdp_%.bbappend
@@ -1,0 +1,15 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+SRC_URI += "file://0001-xorg.conf-Make-sure-the-glamor-module-is-loaded.patch"
+
+DEPENDS += "libdrm"
+
+FILES:${PN}:remove = "${libdir}/xorg/modules/*"
+FILES:${PN} += "${libdir}/xorg/modules/*.so"
+FILES:${PN} += "${libdir}/xorg/modules/*/*.so"
+FILES:${PN}-staticdev += "${libdir}/xorg/modules/*.a"
+FILES:${PN}-staticdev += "${libdir}/xorg/modules/*/*.a"
+
+CFLAGS:prepend = "-I${STAGING_INCDIR}/libdrm "
+EXTRA_OECONF =+ "--enable-glamor"
+
--- a/files/group
+++ b/files/group
@@ -6,6 +6,7 @@ ni:x:500:
 openvpn:x:499:
 niwscerts:x:498:
 # free space
+xrdp:x:403:
 arpwatch:x:402:
 ptest:x:401:
 ossec:x:400:
--- a/files/passwd
+++ b/files/passwd
@@ -6,6 +6,7 @@ webserv:x:501::::
 lvuser:x:500::::
 openvpn:x:499::::
 # free space
+xrdp:x:403::::
 arpwatch:x:402::::
 ptest:x:401::::
 ossec:x:400::::
--- a/recipes-core/packagegroups/packagefeed-ni-extra.bb
+++ b/recipes-core/packagegroups/packagefeed-ni-extra.bb
@@ -4,7 +4,7 @@ LICENSE = "MIT"
 inherit packagegroup
 
 # Graphical extra packages
-RDEPENDS:${PN}:append = "\
+RDEPENDS:${PN}:append:x64 = "\
 	packagegroup-self-hosted \
 	packagegroup-xfce-extended \
 	consolekit \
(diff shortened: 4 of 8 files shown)


jhershbe:
I guess this should have a startup level

jhershbe:
Confirmed intel gpu acceleration on cRIO-9042

```
[506470.673] (II) Loading sub module "glamoregl"
[506470.673] (II) LoadModule: "glamoregl"
[506470.673] (II) Loading /usr/lib/xorg/modules/libglamoregl.so
[506470.673] (II) Module glamoregl: vendor="X.Org Foundation"
[506470.673]    compiled for 1.21.1.13, module version = 1.0.1
[506470.673]    ABI class: X.Org ANSI C Emulation, version 0.4
[506470.673] rdpPreInit: glamor module load ok
[506470.709] (II) XRDPDEV(0): glamor X acceleration enabled on Mesa Intel(R) HD Graphics 500 (APL 2)
[506470.709] rdpPreInit: glamor init ok
```

recipes-core/packagegroups/packagefeed-ni-core.bb
```diff
@@ -24,6 +24,8 @@ RDEPENDS:${PN} = "\
 
 RDEPENDS:${PN}:append = "\
 	docker \
+	xorgxrdp \
```
chaitu236: We are now building ARM from same codebase and I believe these packages are not applicable to ARM targets.

Can you move them into [packagegroup-ni-graphical](https://github.com/ni/meta-nilrt/blob/nilrt/master/scarthgap/recipes-core/packagegroups/packagegroup-ni-graphical.bb#L1) assuming these changes are not xfce specific; if they are, they should go into [packagegroup-ni-xfce](https://github.com/ni/meta-nilrt/blob/nilrt/master/scarthgap/recipes-core/packagegroups/packagegroup-ni-xfce.bb#L1).

chaitu236 (changes requested):
Thanks for the PR!

recipes-support/xorg-xrdp/xorgxrdp_%.bbappend
```diff
@@ -0,0 +1,15 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
```
chaitu236, line 1: > xorgxrdp_0.9.%.bbappend
> xrdp_0.9.%.bbappend

Also, can you make the bbappend not version specific? i.e.,

> xorgxrdp_%.bbappend
> xrdp_%.bbappend

If the version changes, we would still want these changes applied. And if they don't apply cleanly, I think it's preferable that it fail the build rather than the warning we get on orphaned bbappend which may not get noticed.

recipes-core/packagegroups/packagefeed-ni-core.bb
jhershbe: Sure... it's not specific to xfce, I just added the ability to start it if it's there. Moved to graphical

recipes-support/xorg-xrdp/xorgxrdp_%.bbappend
jhershbe: Done

recipes-core/packagegroups/packagefeed-ni-core.bb
chaitu236: Sorry for the bad advice, but I just realized that `packagegroup-ni-graphical` is included in BSI by default.
I believe `xorgxrdp` and `xrdp` are optional packages that only need to be present in feeds.

In that case [packagefeed-ni-extra](https://github.com/ni/meta-nilrt/blob/nilrt/master/scarthgap/recipes-core/packagegroups/packagefeed-ni-extra.bb) is probably the better place for these. [This section](https://github.com/ni/meta-nilrt/blob/nilrt/master/scarthgap/recipes-core/packagegroups/packagefeed-ni-extra.bb#L6) already contains some remote desktop packages so is a good place to put these two (in alphabetical order); but that section is `RDEPENDS:${PN}:append` which is not correct for graphical programs so if you can change it to `RDEPENDS:${PN}:append:x64`, that'd be appreciated.

chaitu236 (approved):

gratian (approved):

PR #904, opened 2025-11-20
clamav: Add memory-aware scan wrapper for NILRT systems

hemant-jadhav-emerson:
### Summary of Changes
Add clamav-scan wrapper script and bbappend for NILRT-specific memory management when scanning on resource-constrained embedded systems.

Added DNS backup and restore logic for SNAC Mode.

The wrapper script:
- Automatically creates swap space when system RAM is below 3GB
- Configurable swap size (default 1024MB) via --swap-size option

Added tech debt [AB#3560775](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3560775) to track the DNS backup and restore workaround.

### Justification

This addresses memory exhaustion issues on devices with limited RAM when running virus database updates and scans.
[AB#3286105](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3286105)

### Testing
Tested on NI cRIO-9038 when combined with meta-nilrt changes
- ClamAV installs without errors without SNAC mode too
- freshclam successfully downloads virus databases without any human intervention
- clamscan and clamav-scan scans for malware
- Swap wrapper manages memory automatically

Tested following scenarios
1. install clamAV - Run freshclam and clamscan
2. install clamAV - Configure Snac Mode - Run freshclam and clamscan (DNS backup restore is required)
3. Configure Snac Mode - install clamAV - Run freshclam and clamscan (DNS backup restore is required)

### Procedure

### DIFF
--- a/recipes-scanners/clamav/files/clamav-scan
+++ b/recipes-scanners/clamav/files/clamav-scan
@@ -1,0 +1,109 @@
+#!/bin/sh
+#
+# ClamAV scan wrapper
+# Automatically manages swap memory for systems with limited RAM
+#
+
+# Check for root privileges
+if [ "$(id -u)" -ne 0 ]; then
+  echo "Error: This script must be run as root." >&2
+  exit 1
+fi
+
+# Configuration
+MIN_MEMORY_MB=3072
+SWAP_SIZE_MB=1024
+# Use /var/cache which should be on root filesystem
+# Avoid /tmp and /var/volatile which are tmpfs (RAM-based)
+TEMP_SWAP="/var/cache/clamav_swap"
+
+# Parse wrapper-specific options
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --swap-size)
+      SWAP_SIZE_MB="$2"
+      shift 2
+      ;;
+    -h|--help)
+      echo "ClamAV Scan Wrapper"
+      echo ""
+      echo "Usage: clamav-scan [--swap-size MB] [clamscan options...]"
+      echo ""
+      echo "Wrapper Options:"
+      echo "  --swap-size MB    Set swap file size in MB (default: 1024)"
+      echo "  -h, --help        Show this help"
+      echo ""
+      echo "All other options are passed to clamscan."
+      exit 0
+      ;;
+    *)
+      # Not a wrapper option, pass to clamscan
+      break
+      ;;
+  esac
+done
+
+# Memory and swap management
+check_and_create_swap() {
+  # Check total memory in KB and convert to MB
+  total_mem_kb=$(grep MemTotal /proc/meminfo | awk '{print $2}')
+  total_mem_mb=$((total_mem_kb / 1024))
+  
+  # Check if any swap is currently active
+  swap_active=$(swapon --show 2>/dev/null | wc -l)
+  
+  # Create and enable swap file only if memory < threshold and no swap is active
+  if [ "$total_mem_mb" -lt "$MIN_MEMORY_MB" ] && [ "$swap_active" -eq 0 ]; then
+    
+    # Clean up any existing swap file
+    if [ -f "$TEMP_SWAP" ]; then
+      swapoff "$TEMP_SWAP" 2>/dev/null || true
+      rm -f "$TEMP_SWAP"
+    fi
+    
+    # Create swap file using dd (works on all filesystems and architectures)
+    if ! dd if=/dev/zero of="$TEMP_SWAP" bs=1M count=$SWAP_SIZE_MB 2>&1; then
+      echo "Warning: Failed to create ${SWAP_SIZE_MB}MB swap file. Continuing without swap." >&2
+      echo "         ClamAV may fail or produce errors due to insufficient memory." >&2
+      rm -f "$TEMP_SWAP" 
+      return 0
+    fi
+    
+    chmod 600 "$TEMP_SWAP"
+    if ! mkswap "$TEMP_SWAP" >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


hemant-jadhav-emerson:
@amstewart Please review

recipes-scanners/clamav/clamav_1.4.3.bbappend
```diff
@@ -0,0 +1,10 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
```
amstewart: Nonblocking. If you rename the bbappend to `clamav_1.%.bbappend`, it will match against any clamav recipe that has major-version 1.X.X. That will keep this bbappend from throwing an error at the RTOS team when the underlying clamav recipe gets updated - as long as it isn't to a new major version.

amstewart (approved):

recipes-scanners/clamav/clamav_1.4.3.bbappend
hemant-jadhav-emerson: I have some additional changes. I will push those together.

recipes-scanners/clamav/files/clamav-scan
```diff
@@ -0,0 +1,111 @@
...
+    if [ -f "$TEMP_SWAP" ]; then
+      swapoff "$TEMP_SWAP" 2>/dev/null || true
+      rm -f "$TEMP_SWAP"
+    fi
+    
+    # Create swap file
+    if ! fallocate -l "${SWAP_SIZE_MB}M" "$TEMP_SWAP" 2>&1; then
+      # Fallback to dd if fallocate not available
+      if ! dd if=/dev/zero of="$TEMP_SWAP" bs=1M count=$SWAP_SIZE_MB 2>&1; then
+        echo "Warning: Failed to create ${SWAP_SIZE_MB}MB swap file. Continuing without swap." >&2
```
chaitu236: If dd fails due to disk filling up, I think the partially created swap file may need to be cleaned up sooner (to prevent runtime issues due to full disk) than when `trap` runs on exit.

recipes-scanners/clamav/files/clamav-scan
```diff
@@ -0,0 +1,111 @@
...
+    
+    # Clean up any existing swap file
+    if [ -f "$TEMP_SWAP" ]; then
+      swapoff "$TEMP_SWAP" 2>/dev/null || true
+      rm -f "$TEMP_SWAP"
+    fi
+    
+    # Create swap file
+    if ! fallocate -l "${SWAP_SIZE_MB}M" "$TEMP_SWAP" 2>&1; then
+      # Fallback to dd if fallocate not available
```
chaitu236: I see that `fallocate` is already installed on scarthgap BSI; not sure which package pulls it in as we don't explicitly add it.

And `fallocate` seems to check that disk space is available before creating file - unlike `dd`. Can we just remove the dd below and declare an explicit dependency on `util-linux-fallocate` from this bbappends?

chaitu236 (changes requested):

recipes-scanners/clamav/clamav_1.4.3.bbappend
chaitu236: I don't think this script is doing anything version specific so we may as well go with `clamav_%.bbappend`.

amstewart: My opinion is that one shouldn't wildcard out the major version without a very good reason. If clamav bumps their major: it is the maintainers communicating that there is a really substantial break to the API. I think that should warrant the warning in OE and having someone do a quick check that the wrapper script still makes sense.

I'll yield to your decision however.

chaitu236: It may be a good practice but all this script is doing that's clam specific is call `clamscan` and unless this name is changed, I don't see how this script can break.
I don't have any strong opinions on this so either way is fine with me.

recipes-scanners/clamav/files/clamav-scan
chaitu236: Just noticed that fallocate doesn't work on ARM devices - possibly due to unsupported filesystem; so we may need `dd` after all.

Probably still better to declare explicit dependency on `util-linux-fallocate` on x64 targets (like [this](https://github.com/ni/meta-nilrt/blob/nilrt/master/next/recipes-core/packagegroups/packagegroup-ni-base.bb#L97)).

recipes-scanners/clamav/files/clamav-scan
```diff
@@ -0,0 +1,108 @@
...
+    # Clean up any existing swap file
+    if [ -f "$TEMP_SWAP" ]; then
+      swapoff "$TEMP_SWAP" 2>/dev/null || true
+      rm -f "$TEMP_SWAP"
+    fi
+    
+    # Create swap file using dd (works on all filesystems and architectures)
+    if ! dd if=/dev/zero of="$TEMP_SWAP" bs=1M count=$SWAP_SIZE_MB 2>&1; then
+      echo "Warning: Failed to create ${SWAP_SIZE_MB}MB swap file. Continuing without swap." >&2
+      echo "         ClamAV may fail or produce errors due to insufficient memory." >&2
```
chaitu236, line 67: > If dd fails due to disk filling up, I think the partially created swap file may need to be cleaned up sooner (to prevent runtime issues due to full disk) than when trap runs on exit.

@hemant-jadhav-emerson This is still an issue right?

chaitu236 (changes requested):

PR #906, opened 2025-12-05
Adapt Salt recipe for the Salt 3006.13 upgrade

AlexDanDuna:
### Summary of Changes

Adapt BitBake file for salt for the Salt 3006.13 upgrade:
- Add new dependencies used by salt
- Remove salt-pip as it triggered errors due to it not being used by any packages
- Remove tornado; salt includes its own tornado; This, however, has changed in newer salt versions.

### Justification

As we're upgrading salt to 3006.13, we need to make appropriate modifications to the bb file that we use in order to make the salt minion for LinuxRT.

### Testing

We performed testing using a locally built LinuxRT image.

### Procedure

### DIFF
--- a/recipes-support/salt/salt_3006.13.bb
+++ b/recipes-support/salt/salt_3006.13.bb
@@ -1,7 +1,7 @@
 HOMEPAGE = "http://saltstack.com/"
 SECTION = "admin"
 LICENSE = "Apache-2.0"
-LIC_FILES_CHKSUM = "file://LICENSE;md5=c996f5a78d858a52c894fa3f4bec68c1"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=3ab0f85f9c3b3eb8184e17b99fd4dfb2"
 DEPENDS = "\
     python3-msgpack \
     python3-pyyaml \
@@ -14,11 +14,11 @@ DEPENDS += "\
 "
 
 PACKAGECONFIG = "tcp zeromq"
-PACKAGECONFIG[tcp] = ",,python3-pycryptodome"
-PACKAGECONFIG[zeromq] = ",,python3-pycryptodome python3-pyzmq"
+PACKAGECONFIG[tcp] = ",,python3-pycryptodomex"
+PACKAGECONFIG[zeromq] = ",,python3-pycryptodomex python3-pyzmq"
 
 SRC_URI = "\
-    git://github.com/ni/salt.git;protocol=https;branch=ni/master/3000.2 \
+    git://github.com/ni/salt.git;protocol=https;branch=ni/master/3006.13 \
     file://minion \
     file://salt-minion \
     file://salt-common.bash_completion \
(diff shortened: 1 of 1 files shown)


recipes-support/salt/salt_3006.13.bb
```diff
@@ -1,7 +1,7 @@
 HOMEPAGE = "http://saltstack.com/"
```
chaitu236, line 1: > Signed-off-by: Alpar Perini <12642481+palpar@users.noreply.github.com>

Please sign off with emerson email.
And if multiple people have worked on a commit, it is customer to add all of their sign off tags.

recipes-support/salt/salt_3006.13.bb
```diff
@@ -1,7 +1,7 @@
...
 SECTION = "admin"
```
chaitu236, line 2: > Add the changes needed for the salt upgrade to 3006.13

The the format most openembedded commits use is `recipe/file: <change>`. So commit message could be updated to something like `salt: Upgrade to 3006.13`.

recipes-support/salt/salt_3006.13.bb
```diff
@@ -118,13 +121,20 @@ RDEPENDS:${PN}-common = "\
     python3-backports-ssl-match-hostname \
```
chaitu236, line 121: As discussed offline, with these changes BSI has become big enough that installation on smaller ARM32 targets (i.e., most of them) fails.
We should take a closer look at the dependencies and remove anything that's not required.

recipes-support/salt/salt_3006.13.bb
```diff
@@ -118,13 +121,20 @@ RDEPENDS:${PN}-common = "\
...
     python3-charset-normalizer \
     python3-core \
+    python3-croniter (>= 0.3.23) \
```
chaitu236: python3-croniter -> python3-natsort -> python3-icu -> libicui18n74 -> libicudata74

This dependency chain pulls in `libicudata74` which takes up 30MB of disk space.

It appears that croniter is [only required](https://github.com/ni/salt/blob/ni/master/3006.13/salt/utils/schedule.py#L1182) for handling cron jobs; do we really need it for systemlink?

Removing `python3-croniter` saves a lot of space but this is still not enough for ARM installation to succeed. We should look at other dependencies similarly (preferably even the existing ones before this PR).

chaitu236 (changes requested):

recipes-support/salt/salt_3006.13.bb
AlexDanDuna: Removed croniter and tornado. Also added other deps (which are pulled anyway, for example through requests, but it's important to have them match salt's requirements files).

recipes-support/salt/salt_3006.13.bb
AlexDanDuna: Removed croniter and tornado. Also added other deps (which are pulled anyway, for example through requests, but it's important to have them match salt's requirements files).

chaitu236 (approved):
Commits need to be squashed

chaitu236:
Needs to be merged after https://github.com/ni/meta-nilrt/pull/907

chaitu236:
Squashed commits and merged as b6b4a76a05108a84f19ce146f8fe76b79a0492c3

PR #917, opened 2026-01-05
pkgfeed-ni-core: Enable TPM2 Tools

usercw88:
### Summary of Changes
Enable TPM2 user-space tools in packagefeed-ni-core
TPM2 CONFIG modules enabled in nati_x86_64_defconfig: https://github.com/ni/linux/pull/265

### Justification
[AB#3643436](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3643436)

### Testing
* [x] I have installed the built base system image onto a target machine
* [x] I have created a python server of the core package feed with this PR in place
* [x] I have installed libtss2-tcti-device onto a target machine
* [x] I have installed tpm2-tools onto a target machine
* [x] I have executed tpm2 commands

### Procedure

### DIFF
--- a/conf/distro/nilrt.inc
+++ b/conf/distro/nilrt.inc
@@ -22,6 +22,11 @@ DISTRO_FEATURES += "\
         virtualization \
 "
 
+# Enable TPM2 features for x64 architectures
+DISTRO_FEATURES:append:x64 = "\
+        tpm2 \
+"
+
 # Because NIRLT uses opkg so intimately, assert that the rootfs should always
 # be made of IPKs.
 PACKAGE_CLASSES =+ "package_ipk"
--- a/recipes-core/packagegroups/packagegroup-ni-tpm.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-tpm.bb
@@ -1,0 +1,15 @@
+# (C) Copyright 2026,
+#  National Instruments Corporation.
+#  All rights reserved.
+
+SUMMARY = "TPM packages for NI Linux Realtime distribution"
+LICENSE = "MIT"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+inherit packagegroup
+
+RDEPENDS:${PN} = "\
+	packagegroup-security-tpm2 \
+	libtss2-tcti-device \
+"
--- a/conf/templates/default/bblayers.conf.sample
+++ b/conf/templates/default/bblayers.conf.sample
@@ -25,6 +25,7 @@ BBLAYERS ?= " \
 	${GIT_REPODIR}/meta-rauc \
 	${GIT_REPODIR}/meta-sdr \
 	${GIT_REPODIR}/meta-security \
+	${GIT_REPODIR}/meta-security/meta-tpm \
 	${GIT_REPODIR}/meta-selinux \
 	${GIT_REPODIR}/meta-virtualization \
 	${GIT_REPODIR}/openembedded-core/meta \
--- a/recipes-core/packagegroups/packagefeed-ni-core.bb
+++ b/recipes-core/packagegroups/packagefeed-ni-core.bb
@@ -34,6 +34,7 @@ RDEPENDS:${PN}:append:x64 = "\
 	packagegroup-ni-graphical \
 	packagegroup-ni-nohz-kernel \
 	packagegroup-ni-snac \
+	packagegroup-ni-tpm \
 	bolt \
 	onboard \
 "


recipes-core/packagegroups/packagegroup-ni-tpm.bb
```diff
@@ -0,0 +1,15 @@
...
+#  All rights reserved.
+
+SUMMARY = "TPM packages for NI Linux Realtime distribution"
+LICENSE = "MIT"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+inherit packagegroup
+
+RDEPENDS:${PN}:x64 = "\
```
chaitu236: Maybe leave out x64 as you're already including this packagegroup in packagefeed-ni-core only for x64.

We do the same for other packagegroups e.g., https://github.com/ni/meta-nilrt/blob/nilrt/master/scarthgap/recipes-core/packagegroups/packagegroup-ni-graphical.bb#L1

conf/distro/nilrt.inc
```diff
@@ -22,6 +22,11 @@ DISTRO_FEATURES += "\
         virtualization \
 "
 
+# ENABLE TPM
+DISTRO_FEATURES:append:x64 = "\
+    tpm2 \
```
chaitu236: This seems to enable [these](https://github.com/ni/meta-security/blob/nilrt/master/scarthgap/meta-tpm/recipes-kernel/linux/linux-yocto/tpm2.cfg) kernel config options. Can you diff the new `.config` with old one to see there are no unwanted changes (like existing `y`'s becoming `m`'s or `n`'s).

chaitu236 (changes requested):

conf/distro/nilrt.inc
usercw88: Diff'd the .config with this change against the old one and there were no differences

recipes-kernel/linux/files/extra_x64.cfg
```diff
@@ -3075,6 +3084,7 @@ CONFIG_TCG_TIS_SPI=m
 CONFIG_TCG_TIS_ST33ZP24=m
 CONFIG_TCG_TIS_ST33ZP24_I2C=m
 CONFIG_TCG_TIS_ST33ZP24_SPI=m
+CONFIG_TCG_TPM=m
```
chaitu236: Already part of [nati_x86_64_defconfig](https://github.com/ni/linux/blob/nilrt/master/6.12/arch/x86/configs/nati_x86_64_defconfig#L451).

recipes-kernel/linux/files/extra_x64.cfg
```diff
@@ -3064,8 +3070,11 @@ CONFIG_TAHVO_USB=m
 CONFIG_TAP=m
 CONFIG_TARGET_CORE=m
 CONFIG_TCG_ATMEL=m
+CONFIG_TCG_CRB=m
 CONFIG_TCG_INFINEON=m
 CONFIG_TCG_NSC=m
+CONFIG_TCG_TIS=m
```
chaitu236: Already part of [nati_x86_64_defconfig](https://github.com/ni/linux/blob/nilrt/master/6.12/arch/x86/configs/nati_x86_64_defconfig#L452).

recipes-kernel/linux/files/extra_x64.cfg
```diff
@@ -1072,6 +1072,7 @@ CONFIG_HWMON_VID=m
 CONFIG_HWPOISON_INJECT=m
 CONFIG_HW_RANDOM_BA431=m
 CONFIG_HW_RANDOM_TIMERIOMEM=m
+CONFIG_HW_RANDOM_TPM=m
```
chaitu236: [FYI] For modules in this file we have a test [here](https://dev.azure.com/ni/DevCentral/_git/Tools?path=/test/rtos_oetest/suite/rtos_oetest/testextrakernelmodules_files/testextramodules.sh&version=GBmain&_a=contents). Consider if we want these these modules added to the test.

chaitu236: `extra_x64.cfg` is only applied when building default kernel and not `-nohz`, `-next` or `-debug` kernel; which means the modules won't be loadable on the alternate kernels (if their generated `.config` doesn't contain them that is). Is that acceptable for the feature?

recipes-kernel/linux/files/extra_x64.cfg
```diff
@@ -2655,6 +2660,7 @@ CONFIG_SCSI_WD719X=m
 CONFIG_SDIO_UART=m
 CONFIG_SDR_MAX2175=m
 CONFIG_SD_ADC_MODULATOR=m
+CONFIG_SECURITYFS=m
```
chaitu236: image

Even without adding these CONFIG options explicitly, most of them end up being either `y` or `m` in generated `.config`. I'm trying to figure out what our policy should be regarding `extra_x64.cfg`; should we only allow *new* CONFIG options not already enabled by default or if this is fine.

Another approach is to have a separate .cfg file for these with the expectation that duplicates are ok; and include that in `linux-nilrt.inc` so all variants get it (or modify `nati_x86_64_defconfig` directly in which case local builds get them as well).

@gratian any opinions?

gratian: Just to close the loop here - we've talked offline and the conclusion was to include these changes directly in the main [nati_x86_64_defconfig](https://github.com/ni/linux/blob/nilrt/master/6.12/arch/x86/configs/nati_x86_64_defconfig) unless we see a large disk size penalty.

usercw88: @gratian @chaitu236 I built the linux kernel with and without these TPM CONFIGs and the difference in bzImage size is appx 40KB

chaitu236 (approved):

PR #918, opened 2026-01-06
Start ni-cgroups init script earlier in the boot order

gratian:
### Summary of Changes

Move the ni-cgroups init script to start earlier in the boot. Also remove unnecessary 'cp' calls from v2 script.

### Justification

Currently the 'ni-cgroups' init script is started at rc5/S04 run level. This is fairly late in the boot process and it creates unnecessary cgroup migration work for the kernel. Move it earlier in the boot process to avoid this and make sure everything is migrated by the time lvrt starts. Note that any processes started after this script runs will automatically inherit the cgroup setup from their parent (which is the desired behavior). Also clean-up redundant 'cp' calls.

[AB#3515413](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3515413)

### Testing

* [x] `bitbake ni-cgroups`
* [x] Installed `ni-cgroups` from local feed on a cRIO-9037 system
* [x] Validated that the cgroups v1 code path works as expected and the LV ATS test for CPU pool assignments passes
* [x] Validated the cgroups v2 mount point, folder hierarchy, and cgroup assignment for a timed loop is correct. Note that the cgroups v2 LV ATS test is still in work on the LV side.

### Procedure

### DIFF
--- a/recipes-ni/ni-cgroups/files/ni-cgroups
+++ b/recipes-ni/ni-cgroups/files/ni-cgroups
@@ -3,7 +3,7 @@
 # Provides:          ni-cgroups
 # Required-Start:    $local_fs
 # Required-Stop:
-# Default-Start:     4 5
+# Default-Start:     S
 # Default-Stop:      0 6
 # Short-Description: Setup cgroups LabVIEW RT & assign tasks
 ### END INIT INFO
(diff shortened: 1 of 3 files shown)


chaitu236 (approved):

chaitu236:
Cherry-picked both commits to `nilrt/26.0/scarthgap` - e08b7a5a1660ce3d88c3c1a8283068a29dd8daef, 51b83b51b099fdfcc820e0dcba21a78c16f17aae

PR #928, opened 2026-01-20
Fix DKMS postinst path to restore kernel module auto-build

shubhanshumt26:
### Summary of Changes

- This change fixes an issue where DKMS post-install scripts were not executed correctly, leading to missing kernel modules and failures in NI-PAL and syscfg.
- This PR also Cherry-picks this PR [commit](https://github.com/ni/meta-nilrt/pull/927) as that was causing failure in `packagegroup-kernel-modules-essential.bb`

### Justification

[AB#3039672](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3039672)
FYI: [PR Link](https://dev.azure.com/ni/DevCentral/_git/921942b5-b2be-4a87-b97c-79c194f82203/pullrequest/1143089)

### Testing

* [x] I have built the BSI and installed successfully on target.
* [x] I have tested that installing packages also installs the relevant dependencies &  `common-postinst` script executes successfully after that(`lsni` , `system-config`, etc. is working as expected).

### Procedure

### DIFF
--- a/recipes-kernel/dkms/dkms.bb
+++ b/recipes-kernel/dkms/dkms.bb
@@ -27,7 +27,7 @@ do_compile[noexec] = "1"
 
 
 do_install() {
-	oe_runmake install DESTDIR=${D} LIBDIR=${libdir} 
+	oe_runmake install DESTDIR=${D} LIBDIR=${libdir}/dkms
 }
 
 
@@ -37,7 +37,12 @@ EXTRA_OEMAKE += " -o tarball"
 INSANE_SKIP:${PN} += "dev-deps"
 
 
-FILES:${PN} += " ${datadir}/bash-completion/* ${datadir}/zsh/* ${libdir}/dkms_autoinstaller ${libdir}/common.postinst"
+FILES:${PN} += " \
+	${datadir}/bash-completion/* \
+	${datadir}/zsh/* \
+	${libdir}/dkms/common.postinst \
+	${libdir}/dkms/dkms_autoinstaller \
+"
 
 RDEPENDS:${PN} += " \
 	bash \
--- a/recipes-kernel/packagegroups/packagegroup-kernel-modules-essential.bb
+++ b/recipes-kernel/packagegroups/packagegroup-kernel-modules-essential.bb
@@ -408,9 +408,6 @@ RDEPENDS:${PN} = "\
 	kernel-module-thunderbolt \
 	kernel-module-tls \
 	kernel-module-tmp421 \
-	kernel-module-tpm \
-	kernel-module-tpm-tis \
-	kernel-module-tpm-tis-core \
 	kernel-module-ts-bm \
 	kernel-module-ts-fsm \
 	kernel-module-ts-kmp \


rajendra-desai-ni (approved):

recipes-kernel/packagegroups/packagegroup-kernel-modules-essential.bb
```diff
@@ -408,9 +408,6 @@ RDEPENDS:${PN} = "\
 	kernel-module-thunderbolt \
 	kernel-module-tls \
 	kernel-module-tmp421 \
-	kernel-module-tpm \
```
chaitu236, line 411: > pkgrp-kern-mod-ess: Remove built-in TPM modules

Cherry-pick commits should be cherry-picked with `-x` option (i.e., `git cherry-pick -x`) so that there is a cherry-pick from line in the commit. Also, use `-s` option to add signoff tag.

recipes-kernel/dkms/dkms.bb
```diff
@@ -27,7 +27,10 @@ do_compile[noexec] = "1"
 
 
 do_install() {
-	oe_runmake install DESTDIR=${D} LIBDIR=${libdir}
```
chaitu236, line 30: nit: this file uses tabs and your changes are using spaces; let's not mix formatting - just keep to the original format and use tabs instead of spaces in this file.

chaitu236 (changes requested):
The commit message needs fixing for clarity/accuracy

> DKMS expects common.postinst to be available at
/usr/lib/dkms/common.postinst during package postinst execution.

I think you mean that `NI drivers expect dkms common.postinst to be available at
/usr/lib/dkms/common.postinst`.

> Previously, the file was installed only as:
/usr/lib/common.postinst

It is not clear from the commit message why the recipe suddenly stopped installing the file to `/usr/lib/dkms`. It appears that it was [our change](https://github.com/ni/meta-nilrt/commit/41994cc8a74b31fd7e08cd0e5b14af045898334a) that changed the install path. Looking at upstream dkms' Makefile, it [appears](https://github.com/dkms-project/dkms/blob/aefb03c66ac8ea0f4fa1bcd43d39cf97bfb014a8/Makefile#L11) that the file is supposed to be installed to `/usr/lib/dkms` and not `/usr/lib/`. The [package in debian](https://packages.debian.org/trixie/all/dkms/filelist) also doesn't show the file being installed in `/usr/lib`.

So, do we really need this to be installed to `/usr/lib` or did we perhaps not notice that dkms' Makefile [uses](https://github.com/dkms-project/dkms/blob/aefb03c66ac8ea0f4fa1bcd43d39cf97bfb014a8/Makefile#L11) `/usr/lib/dkms` for `LIBDIR` and not `/usr/lib`. If the latter, we need to fix this recipe.

Please compare filepaths of files in this package to what we ship in scarthgap and what upstream dkms' Makefile creates / debian package. We should generally stick to scarthgap's layout (as our drivers rely on this as we just found out) unless upstream is different and there is a good reason for us to follow upstream.

chaitu236 (approved):
If we've verified that package generated here has same file structure as scarthgap one, feel free to merge.

shubhanshumt26:
> If we've verified that package generated here has same file structure as scarthgap one, feel free to merge.

Yes, I verified with a older BSI that has scarthgap now the file structure is exactly same for next distribution as it was in scarthgap.

PR #933, opened 2026-01-26
initscripts-nilrt: Add script to sync root

erickshepherdNI:
### Summary of Changes
Add a new init.d script that synchronizes the root entries in /etc/passwd and /etc/shadow between runmode and safemode using shared /etc/natinst/share/passwd and /etc/natinst/share/shadow files.

### Justification
With the removal of niauth we need a way of sharing the root password between runmode and safemode. This init script will automatically copy the root information to files in the natinst share partition when the system shuts down and then copy it back when the system starts.

### Testing
Local testing
-  Confirmed that the root data is only copied to /etc/natinst/share on shutdown if the passwd/shadow files exist there and there is a difference in the root data
- Confirmed that the root data is only copied to /etc/ on boot if the passwd/shadow files exist, pwck succeeds on the shared files, and there is a difference in the root data

### Procedure

### DIFF
--- a/recipes-ni/ni-acctsync/files/ni-acctsync
+++ b/recipes-ni/ni-acctsync/files/ni-acctsync
@@ -1,0 +1,105 @@
+#! /bin/sh
+#
+
+SYS_PASSWD="/etc/passwd"
+SYS_SHADOW="/etc/shadow"
+
+if [ -z "$2" ]; then
+    SHARED_PASSWD="/boot/.passwd"
+    SHARED_SHADOW="/boot/.shadow"
+else
+    SHARED_PASSWD="$2/.passwd"
+    SHARED_SHADOW="$2/.shadow"
+fi
+
+log() {
+    echo "[acct-sync] $1"
+}
+
+get_root_line() {
+    grep "^root:" "$1"
+}
+
+
+replace_root_line() {
+    SRC="$1"
+    DEST="$2"
+
+    SRC_ROOT=$(get_root_line "$SRC")
+    DEST_ROOT=$(get_root_line "$DEST")
+
+    if [ -z "$SRC_ROOT" ]; then
+        log "No entry for root found in $SRC. Nothing to sync."
+        return 1
+    fi
+
+    # Escape slashes for sed
+    ESCAPED_SRC_ROOT=$(printf '%s\n' "$SRC_ROOT" | sed 's/[\/&]/\\&/g')
+    ESCAPED_DEST_ROOT=$(printf '%s\n' "$DEST_ROOT" | sed 's/[\/&]/\\&/g')
+
+    if [ "$ESCAPED_SRC_ROOT" != "$ESCAPED_DEST_ROOT" ]; then
+        log "Syncing root info from $SRC to $DEST"
+        if grep -q "^root:" "$DEST"; then
+            sed -i "s/^root:.*/$ESCAPED_SRC_ROOT/" "$DEST"
+        else
+            sed -i "1i $ESCAPED_SRC_ROOT" "$DEST"
+            sed -i "/^[[:space:]]*$/d" "$DEST"
+        fi
+    else
+        log "Root entry in $SRC matches the one in $DEST, nothing to sync."
+    fi
+}
+
+log "Shared passwd path: $SHARED_PASSWD"
+log "Shared shadow path: $SHARED_SHADOW"
+
+case "$1" in
+    start)
+        log "Syncing root info from boot partition..."
+
+        if [ -f "$SHARED_PASSWD" ] && [ -f "$SHARED_SHADOW" ]; then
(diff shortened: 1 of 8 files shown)


recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,97 @@
...
+                log "Sync complete."
+            else
+                log "pwck returned $STATUS. Stopping sync."
+            fi
+        else
+            log "Shared passwd or shadow files not found. Stopping sync."
+        fi
+        ;;
+
+    stop)
```
chaitu236, line 80: If the system crashes, syncing back to configfs won't happen. Is there a way to know/detect when a password changed and sync it right away?

recipes-ni/initscripts-nilrt/files/nirootsync
```diff
@@ -0,0 +1,97 @@
...
+        $0 start
+        ;;
+
+    *)
+        echo "Usage: $0 {start|stop|restart|force-reload}"
+        exit 1
+        ;;
+esac
+
+exit 0
```
chaitu236: Needs empty line at end

Image

recipes-ni/initscripts-nilrt/initscripts-nilrt.bb
```diff
@@ -19,6 +19,7 @@ SRC_URI = "\
 	file://nicleanstalelinks \
 	file://nidisablecstates \
 	file://nipopulateconfigdir \
+	file://nirootsync \
```
chaitu236: These changes end up on ARM BSI but ARM safemode is still shipping sumo. Isn't that a problem? Do we also need to update sumo safemode?

recipes-ni/initscripts-nilrt/initscripts-nilrt.bb
```diff
@@ -47,6 +48,7 @@ do_install () {
 	install -m 0755 ${WORKDIR}/nicleanstalelinks     ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/nidisablecstates      ${D}${sysconfdir}/init.d
 	install -m 0755 ${WORKDIR}/nipopulateconfigdir   ${D}${sysconfdir}/init.d
+	install -m 0755 ${WORKDIR}/nirootsync             ${D}${sysconfdir}/init.d
```
chaitu236: Spacing not same as rest of lines.

recipes-ni/initscripts-nilrt/initscripts-nilrt.bb
```diff
@@ -69,6 +71,7 @@ do_install () {
 	update-rc.d -r ${D} nicleanstalelinks     start 5  S .
 	update-rc.d -r ${D} nidisablecstates      start 2 3 4 5 S .
 	update-rc.d -r ${D} nipopulateconfigdir   start 35 S .
+	update-rc.d -r ${D} nirootsync            defaults
```
chaitu236: What are the defaults? Where does this script end up in init order? It needs to be before ssh starts or console allows logins.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,97 @@
+#! /bin/sh
```
chaitu236, line 1: I think we need to have a ptest that runs as part of our test suite to make sure this functionality never breaks as it's critical.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,97 @@
...
+#
+
+SHARED_PASSWD="/etc/natinst/share/passwd"
```
chaitu236: These files don't exist now. When are they supposed to be created?

recipes-ni/ni-acctsync/files/ni-acctsync
erickshepherdNI: This is a good question, I'll do some research and find out if we can get more precise control over when the script runs.

recipes-ni/initscripts-nilrt/initscripts-nilrt.bb
erickshepherdNI: The correct update-rc.d arguments are something I'm still not clear on. I definitely think it should be changed from the default settings before this goes in but it will take some looking into.

recipes-ni/ni-acctsync/files/ni-acctsync
erickshepherdNI: I'm currently planning to update the USB provisioning script to create the files when it sets up the share partition.

chaitu236 (commented):
PR needs to be targeted to `nilrt/master/next`

PR #936, opened 2026-02-03
Include only required i915 firmware in safemode

gratian:
### Summary of Changes

Reduce the size of the safemode image by only including the i915 firmware which is currently in use.

In order to accomplish that:

- Commit caf6dcf ("linux-firmware: split i915 firmware into sub-packages") creates separate packages based on the
Intel CPU/GPU abbreviated code-name and subsystem.
- Commit cd0e25b ("packagegroup-ni-safemode: include only minimum required i915 firmware")
  - Moves the `linux-firmware-i915` package from `packagegroup-ni-base.bb` into the `packagegroup-ni-runmode.bb`. We do want to continue to include the full i915 firmware set in runmode for maximum flexibility.
  - Only includes the required sub-packages in `packagegroup-ni-safemode.bb`. See commit details for supported hardware information.
- While we are at it commit 5380b90 ("linux-nilrt: remove obsolete i915 module parameter") does some clean-up to remove noise from the kernel log related to the `enable_rc6` module parameter which got removed upstream.

### Justification

[AB#3704250](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3704250)

### Testing

* [x] `bitbake nilrt-safemode-rootfs` and deployed the resulting safemode files to cRIO-9042, and PXIe-8862. Validated the i915 firmware is loaded correctly and only specified i915 firmware sub-packages are installed.
* [x] `bitbake nilrt-base-system-image`, extracted the resulting runmode image and validated that a full set of firmware files is present under `/lib/firmware/i915`
* [x] In order to validate what firmware is in use a modified kernel with additional logging was used on: cRIO-903x, cRIO-904x, PXIe-8821, PXIe-8840, and PXIe-8862. Also validated that PXIe-8861, PXIe-8880, and PXIe-8881 do not use i915 firmware (i.e., they use a radeon GPU) and checked that specs for cRIO-905x do not include video output.

### Procedure

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-safemode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-safemode.bb
@@ -21,6 +21,35 @@ RDEPENDS:${PN} = " \
 	sysconfig-settings-ssh \
 "
 
+# GPU firmware, included as split packages to conserve space
+#
+# Intel Valleyview family, no i915 firmware in use
+#   cRIO-903x, CVS-1458RT, CVS-1459RT, IC-312x
+#   sbRIO-960x, sbRIO-962x, sbRIO-9638, cDAQ-913x
+#
+# Intel Broadwell family, no i915 firmware in use
+#   IC-317x
+#
+# Intel Haswell family, no i915 firmware in use
+#   PXIe-8821, PXIe-8840
+#
+# Intel Broxton family, uses 'bxt' dmc firmware, GuC and HuC disabled
+#   cRIO-904x
+#
+# Intel Tiger Lake family, uses 'tgl' dmc firmware, GuC and HuC disabled
+#   PXIe-8822, PXIe-8842, PXIe-8862
+#
+# N/A, no video output
+#   cRIO-905x
+#
+# N/A, radeon GPU
+#   PXIe-8861, PXIe-8880, PXIe-8881
+#
+RDEPENDS:${PN}:append:x64 = "\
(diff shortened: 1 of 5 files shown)


recipes-core/packagegroups/packagegroup-ni-safemode.bb
```diff
@@ -21,6 +21,24 @@ RDEPENDS:${PN} = " \
 	sysconfig-settings-ssh \
 "
 
+# GPU firmware, included as split packages to conserve space
+#
+# cRIO-903x: no firmware in use
```
chaitu236: There are more x64 model defined [here](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/rtos/rtosinstallersd/nilrt_onert_system_image/source/rss/cdf/linux-rt-rss-x64.cdf) (e.g., sbCombo, CVS, IC, etc). Do you think we need to check those as well or is that not necessary?

gratian: I'll check. Thanks.

gratian: @chaitu236 I've checked all the additional models listed in the link above. No new firmware packages are needed. I've updated the recipe source comment to list the additional models (and re-organized it by CPU code name to make it more readable).

chaitu236 (approved):

PR #925, opened 2026-01-14
Fixed build warnings in dist-next NILRT images

AshwinUjjwal:
### Summary of Changes
All warnings encountered while building images have been resolved.
### Justification
[AB#3039668](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3039668)

### Testing

- [x] I have built the core package feed.(bitbake packagefeed-ni-core)
- [x] I have built the core base system image. (bitbake nilrt-base-system-image)
- [x] I have build the core safemode rootfs.(bitbake nilrt-safemode-rootfs)
- [x] I have build the core recovery media.( bitbake nilrt-recovery-media)

### Procedure

* [x] I certify that the contents of this pull request complies with the [Developer Certificate of Origin](https://github.com/ni/nilrt/blob/HEAD/docs/CONTRIBUTING.md#developer-certificate-of-origin-dco).

PR #939, opened 2026-02-06
ni-acctsync: Create init script to sync account info

erickshepherdNI:
### Summary of Changes

Creates a new init script called ni-acctsync that synchronizes the root password hash between runmode and safemode using a shared .shadow file in the configfs partition. This commit also includes changes to the provisioning scripts to prompt users to set a root password and store it in the shared files as well as in an EFI variable. The formatting script was also updated to preserve the shared .shadow file when the configfs partition is formatted.

### Justification

With the removal of niauth we need a way of sharing the root password between runmode and safemode. The ni-acctsync init script will automatically copy the root password hash to a file in the configfs partition when the system shuts down and then insert the hash back into the /etc/shadow file when the system boots. This synchronization is only done if the hash stored in the shared .shadow file and the hash in the /etc/shadow are not equal.

This commit also includes changes to the NI provisioning scripts so that they will prompt users to set a root password when provisioning a device. This password can also be set using a command line argument which will allow manufacturing to automate the password setting process. Additionally the provisioning script stores the new password in an EFI variable so it can be recovered if the device is reset.

### Testing

Local testing:
* [X] Tested that the provisioning script prompts users for a password and that password is recorded in the bootfs partition and an EFI variable
* [X] Tested that using the -p argument for the provisioning script allows a password to be set automatically
* [X] Confirmed that the root hash is copied to the configfs .shadow file if there is a difference between the hashes and that the configfs .shadow file is created if it did not already exist.
* [X] Confirmed that the root hash is copied to the local shadow file if the shared .shadow file exists and has different information than the local shadow file. Also confirmed that the hash needs to be in the expected format to be copied.
* [X] Tested the full workflow by running the provisioning script and setting a password, booting into safemode, installing runmode, booting into runmode, and confirmed that the passwd and shadow files were synchronized properly at each step and the EFI variable persisted
* [X] Confirmed that the shared shadow file and updated local shadow file have the 400 permissions at each step.

### Procedure

### DIFF
--- a/recipes-core/initrdscripts/files/ni_provisioning.safemode.common
+++ b/recipes-core/initrdscripts/files/ni_provisioning.safemode.common
@@ -404,6 +404,29 @@ fixup_configfs()
 	umount $CONFIG_MOUNTPOINT
 }
 
+set_root_password()
+{
+	# Set the restore mode password
+	if [[ -n $ROOT_PASSWORD ]]; then
+		echo "root:$ROOT_PASSWORD" | chpasswd
+	else
+		passwd
+	fi
+
+	# Get the root password hash
+	ROOT_ENTRY=$(grep "^root:" /etc/shadow)
+	ROOT_HASH=$(echo "$ROOT_ENTRY" | cut -d':' -f2)
+
+	# Create the shared shadow file and add root entry
+	CONFIG_MOUNTPOINT=/var/volatile/configfs
+	mkdir -p $CONFIG_MOUNTPOINT
+	mount -L $PART3_LABEL $CONFIG_MOUNTPOINT
+	touch "$CONFIG_MOUNTPOINT/.shadow"
+	chmod 400 "$CONFIG_MOUNTPOINT/.shadow"
+	echo "$ROOT_HASH" > "$CONFIG_MOUNTPOINT/.shadow"
+	umount $CONFIG_MOUNTPOINT
+}
+
 sanity_check()
 {
 	print_info "Sanity check TARGET_DISK=$TARGET_DISK..."
--- a/recipes-ni/ni-acctsync/files/ni-acctsync
+++ b/recipes-ni/ni-acctsync/files/ni-acctsync
@@ -1,0 +1,143 @@
+#! /bin/sh
+#
+
+SYS_SHADOW="/etc/shadow"
+
+if [ -z "$2" ]; then
+    SHARED_SHADOW="/etc/natinst/share/.shadow"
+else
+    SHARED_SHADOW="$2/.shadow"
+fi
+
+log() {
+    logger -p auth.info "[acct-sync] $1"
+}
+
+get_root_line() {
+    grep "^root:" "$1"
+}
+
+check_hash_equality()
+{
+    SHARED_HASH_PATH="$1"
+    LOCAL_HASH_PATH="$2"
+    LOCAL_ROOT=$(get_root_line "$LOCAL_HASH_PATH")
+
+    LOCAL_HASH=$(echo "$LOCAL_ROOT" | cut -d':' -f2)
+    read -r SHARED_HASH < "$SHARED_HASH_PATH"
+
+    if [ "$LOCAL_HASH" = "$SHARED_HASH" ]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+hash_sanity_check()
+{
+    SHADOW_HASH="$1"
+    
+    if [ -z "$SHADOW_HASH" ] || [ "$SHADOW_HASH" = "*" ] || [ "$SHADOW_HASH" = "!" ]; then
+        return 0
+    fi
+    
+    if [ -f "/etc/login.defs" ]; then
+        ENCRYPT_METHOD=$(awk '$1=="ENCRYPT_METHOD" { print $2; exit }' /etc/login.defs)
+        if [ -z "$ENCRYPT_METHOD" ]; then
+            log "ENCRYPT_METHOD not defined, defaulting to SHA512."
+            ENCRYPT_METHOD="SHA512"
+        fi
+    else
+        log "login.defs file not found. Assuming encryption method is SHA512."
+        ENCRYPT_METHOD="SHA512"
+    fi
+    
+    case "$ENCRYPT_METHOD" in
+        SHA512)
+            log "Checking sanity of SHA512 hash..."
+            if [[ "$SHADOW_HASH" =~ ^\$6\$[A-Za-z0-9./]{1,16}\$[A-Za-z0-9./]{86}$ ]]; then
+                log "Sanity check passed. Continue to sync."
+                return 0
+            else
+                log "Sanity check failed. Stopping sync."
+                return 1
+            fi
+            ;;
+        *)
+            log "Unexpected encrypt method. Stopping sync."
+            return 1
+            ;;
+    esac
+}
+
+insert_root_hash()
+{
+    SRC="$1"
+    DEST="$2"
+
+    if ! check_hash_equality "$SRC" "$DEST"; then
+        TMP_SHADOW="$(mktemp /tmp/shadow.XXXXXX)"
+        trap 'rm -f "$TMP_SHADOW"' EXIT HUP INT TERM
+        read -r SHADOW_HASH < "$SRC"
+
+        if hash_sanity_check "$SHADOW_HASH"; then
+            chmod --reference="$DEST" "$TMP_SHADOW"
+            chown --reference="$DEST" "$TMP_SHADOW"
+            awk -F: -v OFS=":" -v shadow_hash="$SHADOW_HASH" '$1 == "root" { $2 = shadow_hash } { print }' "$DEST" > "$TMP_SHADOW"
+            mv "$TMP_SHADOW" "$DEST"
+        fi
+    fi
+}
+
+extract_root_hash()
+{
+    SRC="$1"
+    DEST="$2"
+    SRC_ROOT=$(get_root_line "$SRC")
+
+    if ! check_hash_equality "$DEST" "$SRC"; then
+        TMP_HASH="$(mktemp /tmp/root_hash.XXXXXX)"
+        trap 'rm -f "$TMP_HASH"' EXIT HUP INT TERM
+        chmod --reference="$DEST" "$TMP_HASH"
+        chown --reference="$DEST" "$TMP_HASH"
+        echo "$SRC_ROOT" | cut -d':' -f2 > "$TMP_HASH"
+        mv "$TMP_HASH" "$DEST"
+    fi
+}
+
+case "$1" in
+    start)
+        log "Syncing root info from shared shadow file..."
+
+        if [ -f "$SHARED_SHADOW" ]; then
+            log "Found shared shadow file."
+            insert_root_hash "$SHARED_SHADOW" "$SYS_SHADOW"
+            log "Sync complete."
+        else
+            log "Shared shadow file not found. Stopping sync."
+        fi
+        ;;
+
+    stop)
+        log "Syncing root info to shared shadow file..."
+
+        if [ ! -f "$SHARED_SHADOW" ]; then
+            log "Creating shared shadow file..."
+            touch "$SHARED_SHADOW"
+            chmod 400 "$SHARED_SHADOW"
+        fi
+        extract_root_hash "$SYS_SHADOW" "$SHARED_SHADOW"
+        ;;
+
+    restart|force-reload)
+        $0 stop
+        $0 start
+        ;;
+
+    *)
+        echo "Usage: $0 {start|stop|restart|force-reload}"
+        exit 1
+        ;;
+esac
+
+exit 0
--- a/recipes-ni/ni-acctsync/ni-acctsync_1.0.bb
+++ b/recipes-ni/ni-acctsync/ni-acctsync_1.0.bb
@@ -1,0 +1,21 @@
+SUMMARY = "NI account sync setup"
+DESCRIPTION = "Installs init.d setup scripts that synchronize account information"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+LICENSE = "MIT"
+SECTION = "base"
+
+SRC_URI = "\
+	file://ni-acctsync \
+"
+
+S = "${UNPACKDIR}"
+
+inherit update-rc.d
+
+INITSCRIPT_NAME = "ni-acctsync"
+INITSCRIPT_PARAMS = "start 36 S . stop 11 0 6 ."
+
+do_install () {
+	install -d ${D}${sysconfdir}/init.d/
+	install -m 0700 ${S}/ni-acctsync ${D}${sysconfdir}/init.d/
+}
--- a/recipes-core/initrdscripts/files/ni_provisioning.common
+++ b/recipes-core/initrdscripts/files/ni_provisioning.common
@@ -136,7 +136,7 @@ elif [ "$ARCH" = "x86_64" ]; then
 
 	show_help()
 	{
-		>&3 echo -e "\nUsage: $0 [-hav] [-t DEVICE] [-c y|n] [-b SECONDS] [-F RESTART_OPTION]\n"
+		>&3 echo -e "\nUsage: $0 [-hav] [-t DEVICE] [-c y|n] [-b SECONDS] [-p PASSWORD] [-F RESTART_OPTION]\n"
 		>&3 echo -e "Options\n"
 		>&3 echo "   -h : Show help."
 		>&3 echo "   -a : Ask at every error if continue. Default is N, so at the first error abort."
@@ -146,6 +146,7 @@ elif [ "$ARCH" = "x86_64" ]; then
 		>&3 echo "               that is not the recovery tool is used."
 		>&3 echo "   -c y|n : Initial value of the console out enable flag."
 		>&3 echo "   -b SECONDS : Initial value of bootdelay -- how long (seconds) to show boot menu."
+		>&3 echo "   -p ROOT_PASSWORD : Set the value of the root password."
 		>&3 echo "   -F RESTART_OPTION : Force provisioning (surpress the promts). RESTART_OPTION specifies"
 		>&3 echo "                       what should be done after a succesful provisioning and can have the"
 		>&3 echo "                       following values: reboot, poweroff or shell."
@@ -488,7 +489,7 @@ early_setup()
 
 	OPTIND=1
 
-	while getopts "havt:c:b:F:" opt; do
+	while getopts "havt:c:b:p:F:" opt; do
 		case "$opt" in
 			h)  show_help
 				exit 1
@@ -509,6 +510,8 @@ early_setup()
 				fi
 				;;
 			b)  grubenv_bootdelay="$OPTARG"
+				;;
+			p)  ROOT_PASSWORD="$OPTARG"
 				;;
 			F)  FORCE_PROVISIONING=1
 				if [ $OPTARG == "reboot" -o  "$OPTARG" == "poweroff" -o "$OPTARG" == "shell" ]; then
--- a/recipes-core/initrdscripts/files/ni_provisioning.safemode
+++ b/recipes-core/initrdscripts/files/ni_provisioning.safemode
@@ -37,4 +37,6 @@ trap - ERR
 exec 1>&3
 exec 2>&4
 
+set_root_password
+
 provision_successfull
--- a/recipes-core/packagegroups/packagegroup-ni-restoremode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-restoremode.bb
@@ -21,6 +21,7 @@ RDEPENDS:${PN} += "\
 	grep \
 	init-restore-mode \
 	kmod \
+	ni-acctsync \
 	ni-systemreplication \
 	parted \
 	procps \
--- a/recipes-core/packagegroups/packagegroup-ni-runmode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-runmode.bb
@@ -25,6 +25,7 @@ RDEPENDS:${PN} = "\
 	lldpd \
 	mpfr \
 	nftables \
+	ni-acctsync \
 	ni-cgroups \
 	ni-configpersistentlogs \
 	ni-locale-alias \
--- a/recipes-core/packagegroups/packagegroup-ni-safemode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-safemode.bb
@@ -15,6 +15,7 @@ RDEPENDS:${PN} = " \
 	e2fsprogs-e2fsck \
 	e2fsprogs-mke2fs \
 	e2fsprogs-tune2fs \
+	ni-acctsync \
 	ni-netcfgutil \
 	ni-shutdown-guard-safemode \
 	ni-systemimage \
--- a/recipes-ni/ni-systemformat/files/nisystemformat
+++ b/recipes-ni/ni-systemformat/files/nisystemformat
@@ -23,6 +23,8 @@ RESET_OVERLAY=no
 NETCFG_TMP=/tmp/netconfig.$$
 declare -r BASENAME=${0##*/}
 
+ACCTINFO_TMP=/tmp/acctinfo
+
 NINETCFGUTIL=/usr/local/natinst/bin/ninetcfgutil
 
 if [ "$arch" = "x86_64" ]; then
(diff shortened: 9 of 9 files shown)


recipes-core/initrdscripts/files/ni_provisioning.safemode.common
```diff
@@ -404,6 +404,43 @@ fixup_configfs()
 	umount $CONFIG_MOUNTPOINT
 }
 
+store_default_password()
+{
+	if [[ $ARCH == "x86_64" ]]; then
+		VAR_NAME="DefaultPassword-7c8c526b-2c07-4189-9495-b331d3b8d2cc"
+		VAR_PATH="/sys/firmware/efi/efivars/$VAR_NAME"
+		printf "\x07\x00\x00\x00%s" "$1" > "$VAR_PATH"
```
chaitu236: So we're not using `fw_setenv` on x64 because on x64 `fw_setenv` stores vars in grubenv instead of EFI (like in arm), and we want to retain default pass even after reprovisioning - is that right?

recipes-core/initrdscripts/files/ni_provisioning.safemode.common
```diff
@@ -404,6 +404,43 @@ fixup_configfs()
...
+	# Create the shared passwd and shadow files and add root entry
+	mount_bootfs_partition
+	echo > $BOOTFS_MOUNTPOINT/.passwd "root:x:0:0:root:/home/admin:/bin/sh"
+	echo > $BOOTFS_MOUNTPOINT/.shadow "root:*::0:99999:7:::"
+
+	chmod 600 $BOOTFS_MOUNTPOINT/.passwd
+	chmod 600 $BOOTFS_MOUNTPOINT/.shadow
+
+	# Manually call the acctsync script
+	service ni-acctsync stop $BOOTFS_MOUNTPOINT
```
chaitu236: Does `ni-acctsync`'s initscript start it in restore mode and sync passwd to restoremode's /etc/* file? That doesn't change the provisioning process in any way (e.g., asking for root passwd, etc) right?

recipes-core/initrdscripts/files/ni_provisioning.safemode.common
```diff
@@ -404,6 +404,43 @@ fixup_configfs()
...
+	fi
+}
+
+set_root_password()
+{
+	# Set the restore mode password
+	if [[ -n $ROOT_PASSWORD ]]; then
+		echo "root:$ROOT_PASSWORD" | chpasswd
+	else
+		passwd
```
chaitu236, line 413: Does ni-acctsync's initscript start it in restore mode and sync passwd to restoremode's /etc/* file? That doesn't change the provisioning process in any way (e.g., asking for previous root passwd, etc) right?

And does calling `passwd` change provisioning process in any way other than accepting password from user? e.g., if provisioning process fails (or when it succeeds and user selects shell at the end), do we drop into shell without being prompted for password?

Would it be safer to make changes to non-default passwd, shadow files instead (e.g., by using `--prefix` argument or such)?

recipes-ni/ni-acctsync/ni-acctsync_1.0.bb
```diff
@@ -0,0 +1,23 @@
...
+
+SRC_URI = "\
+	file://ni-acctsync \
+"
+
+S = "${WORKDIR}"
+
+do_install () {
+	install -d ${D}${sysconfdir}/init.d/
+	install -Dm 0755 ${WORKDIR}/ni-acctsync ${D}${sysconfdir}/init.d/
```
chaitu236: Previous line creates `${D}${sysconfdir}/init.d/` so `-D` option right isn't needed right?

recipes-core/initrdscripts/files/ni_provisioning.safemode.common
chaitu236: Ok, I see below that default passwd is modified during reprovisioning - so is there a reason `fw_setenv` isn't used for x64 as well?

recipes-ni/ni-acctsync/ni-acctsync_1.0.bb
```diff
@@ -0,0 +1,23 @@
...
+LICENSE = "MIT"
+SECTION = "base"
+
+DEPENDS += "update-rc.d-native"
+
+SRC_URI = "\
+	file://ni-acctsync \
+"
+
+S = "${WORKDIR}"
```
chaitu236: We put in [this ](https://github.com/ni/meta-nilrt/commit/3cdf2d6dfec7d7898eff403b29ee30eeeee279c6)and [this ](https://github.com/ni/meta-nilrt/pull/925/changes) commit - going from `WORKDIR` to `UNPACKDIR`. I haven't looked at those changes very closely but check if you need to update this recipe with those changes too.

recipes-ni/ni-acctsync/ni-acctsync_1.0.bb
erickshepherdNI: Yes, I think that's right, I'll remove it.

recipes-core/initrdscripts/files/ni_provisioning.safemode.common
erickshepherdNI: ni-acctsync does get installed to the restoremode image but the bootfs partition is not mounted so the passwords do not get synced automatically. The bootfs partition is also mounted in a different location (var/volatile/bootfs instead of /boot) on the restoremode image which prevents it from finding the shared passwd and shadow files. I tested this case and the provisioning process was not affected.

Calling the passwd command does change the root password for the current provisioning image so if you select 'shell' after provisioning it will require the password. I had tried calling passwd with the --prefix option to target the shared files but had a problem where calling 'passwd --prefix' in the script would not prompt the user for input while calling just 'passwd' works. @gratian suggested we use 'passwd' and then call the init script to sync the information.

recipes-core/initrdscripts/files/ni_provisioning.safemode.common
erickshepherdNI: I didn't use fw_setenv for x64 devices because I thought that they did not support U-Boot but only just realized we alter fw_setenv on x64 devices to affect GRUB and EFI vars. I'll work on an update to use fw_setenv for x64 and arm and see how it works.

erickshepherdNI: I ended up changing the fw_setenv wrapper scripts to include the DefaultPassword variable so fw_setenv can be used for x64 and ARM.

erickshepherdNI:
Current open questions:

- Do we store the shared password information in the bootfs or configfs partitions?
    - This PR is set up to use the bootfs partition but configfs would be more secure. Currently, reformatting a device via the HWCU always wipes the configfs partition but I'm researching what changes would be required to fix this. I think changes relating to use of configfs could be done in a separate PR under the HWCU user story.
- An automated test should be created for the new scripts.
    - Would this be ok to do in a separate PR under a new user story?

chaitu236:
> Current open questions:
> 
> * Do we store the shared password information in the bootfs or configfs partitions?
>   
>   * This PR is set up to use the bootfs partition but configfs would be more secure. Currently, reformatting a device via the HWCU always wipes the configfs partition but I'm researching what changes would be required to fix this. I think changes relating to use of configfs could be done in a separate PR under the HWCU user story.

Do you think most of the changes in this PR will be applicable when we switch to configfs as well? If so, that should be fine.

> * An automated test should be created for the new scripts.
>   
>   * Would this be ok to do in a separate PR under a new user story?

Yes, that should be fine.

recipes-ni/ni-acctsync/ni-acctsync_1.0.bb
```diff
@@ -0,0 +1,23 @@
+SUMMARY = "NI account sync setup"
+DESCRIPTION = "Installs init.d setup scripts that synchronize account information"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+LICENSE = "MIT"
+SECTION = "base"
+
+DEPENDS += "update-rc.d-native"
```
amstewart: Prefer to use the `update-rc.d` [bbclass](https://github.com/ni/openembedded-core/blob/nilrt/master/scarthgap/meta/classes-recipe/update-rc.d.bbclass).

amstewart (changes requested):

chaitu236 (approved):

recipes-core/initrdscripts/files/ni_provisioning.safemode.common
```diff
@@ -404,6 +404,37 @@ fixup_configfs()
...
+	fi
+
+	# Store the new root password hash
+	ROOT_ENTRY=$(grep "^root:" /etc/shadow)
+	ROOT_HASH=$(echo "$ROOT_ENTRY" | cut -d':' -f2)
+	store_default_password $ROOT_HASH
+
+	# Create the shared passwd and shadow files and add root entry
+	mount_bootfs_partition
+	echo > $BOOTFS_MOUNTPOINT/.passwd "root:x:0:0:root:/home/admin:/bin/sh"
```
gratian: Do we need to do this? This line (i.e. /etc/passwd) should not change when you change the password. The only thing we need to save and keep in sync is the hash in the shadow file.

recipes-core/initrdscripts/files/ni_provisioning.safemode.common
```diff
@@ -404,6 +404,37 @@ fixup_configfs()
...
+	echo > $BOOTFS_MOUNTPOINT/.shadow "root:*::0:99999:7:::"
+
+	chmod 600 $BOOTFS_MOUNTPOINT/.passwd
+	chmod 600 $BOOTFS_MOUNTPOINT/.shadow
+
+	# Manually call the acctsync script
+	service ni-acctsync stop $BOOTFS_MOUNTPOINT
```
gratian: If we write the hash directly to the `.shadow` like I've suggested above I think we can drop this manual call.

recipes-core/initrdscripts/files/ni_provisioning.safemode.common
```diff
@@ -404,6 +404,37 @@ fixup_configfs()
...
+
+	# Store the new root password hash
+	ROOT_ENTRY=$(grep "^root:" /etc/shadow)
+	ROOT_HASH=$(echo "$ROOT_ENTRY" | cut -d':' -f2)
+	store_default_password $ROOT_HASH
+
+	# Create the shared passwd and shadow files and add root entry
+	mount_bootfs_partition
+	echo > $BOOTFS_MOUNTPOINT/.passwd "root:x:0:0:root:/home/admin:/bin/sh"
+	echo > $BOOTFS_MOUNTPOINT/.shadow "root:*::0:99999:7:::"
```
gratian: It would be preferable to only save the root hash and leave the other fields untouched when we restore it. You already have the root hash above so you should be able to `echo "$ROOT_HASH" > "$BOOTFS_MOUNTPOINT/.shadow"`

recipes-core/initrdscripts/files/ni_provisioning.safemode.common
```diff
@@ -404,6 +404,37 @@ fixup_configfs()
...
+
+	chmod 600 $BOOTFS_MOUNTPOINT/.passwd
+	chmod 600 $BOOTFS_MOUNTPOINT/.shadow
```
gratian: This leaves an open window between the hash getting written and the permission change where in theory somebody else could read it.

To close the gap you could use `umask` or create the `.shadow` file first, set the permissions, and then write the hash to it. Also usually the shadow file has permissions set to read-only for root and nothing else.

Something like
```
touch "$BOOTFS_MOUNTPOINT/.shadow"
chmod 400 "$BOOTFS_MOUNTPOINT/.shadow"
echo "$ROOT_HASH" > "$BOOTFS_MOUNTPOINT/.shadow"
```

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,105 @@
+#! /bin/sh
+#
+
+SYS_PASSWD="/etc/passwd"
```
gratian: We shouldn't need to touch this file.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,105 @@
...
+if [ -z "$2" ]; then
+    SHARED_PASSWD="/boot/.passwd"
+    SHARED_SHADOW="/boot/.shadow"
+else
+    SHARED_PASSWD="$2/.passwd"
+    SHARED_SHADOW="$2/.shadow"
+fi
+
+log() {
+    echo "[acct-sync] $1"
```
gratian: It would be better if these log messages are logged in one of the system logs like `/var/log/auth.log` instead of getting printed to the console.

You can use `logger -p auth.info 'your custom log string'` for that.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,105 @@
...
+    fi
+
+    # Escape slashes for sed
+    ESCAPED_SRC_ROOT=$(printf '%s\n' "$SRC_ROOT" | sed 's/[\/&]/\\&/g')
+    ESCAPED_DEST_ROOT=$(printf '%s\n' "$DEST_ROOT" | sed 's/[\/&]/\\&/g')
+
+    if [ "$ESCAPED_SRC_ROOT" != "$ESCAPED_DEST_ROOT" ]; then
+        log "Syncing root info from $SRC to $DEST"
+        if grep -q "^root:" "$DEST"; then
+            sed -i "s/^root:.*/$ESCAPED_SRC_ROOT/" "$DEST"
```
gratian: Doing an in-place replace with sed is not going to be atomic with respect to power loss, resets, or other interruptions. It can leave the file in a corrupted state.

The usual pattern is to

- use `cp` to make a copy of the file you want to edit
- make the changes - we should only update the hash field in the shadow file and leave the rest of the root line untouched
- use `mv` to replace the original file with the modified copy

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,105 @@
...
+        if [ -f "$SHARED_PASSWD" ] && [ -f "$SHARED_SHADOW" ]; then
+            log "Found shared passwd and shadow files. Checking integrity with pwck..."
+
+            pwck -r "$SHARED_PASSWD" "$SHARED_SHADOW" >/dev/null 2>&1
+            STATUS=$?
+
+            if [ $STATUS -eq 0 ]; then
+                log "Integrity OK. Syncing..."
+                
+                replace_root_line "$SHARED_PASSWD" "$SYS_PASSWD"
```
gratian: I don't think we need to change the passwd file at all.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,105 @@
...
+log() {
+    echo "[acct-sync] $1"
+}
+
+get_root_line() {
+    grep "^root:" "$1"
+}
+
+
+replace_root_line() {
```
gratian: We should replace only the hash and not the whole line.

recipes-ni/ni-acctsync/ni-acctsync_1.0.bb
```diff
@@ -0,0 +1,23 @@
...
+
+SRC_URI = "\
+	file://ni-acctsync \
+"
+
+S = "${UNPACKDIR}"
+
+do_install () {
+	install -d ${D}${sysconfdir}/init.d/
+	install -m 0755 ${S}/ni-acctsync ${D}${sysconfdir}/init.d/
```
gratian: Evaluate if we need to tighten the restrictions here. Only root should have access, so `0700`?

gratian (commented):
An additional enhancement would be to use PAM to detect password changes for root and save the hash to the synchronized location at that time instead of just on shutdown.

That can be accomplished by:

- adding `pam-plugin-exec` to images
- adding a script that runs on password changes by adding a line like the one below
 ```
password        required                        pam_exec.so seteuid /sbin/mysyncscript
```
before
```
password	required			pam_permit.so
```
in
```
/etc/pam.d/common-password
```
- have the script check that the action is a password change and save the hash
```
#!/bin/sh

[ "$PAM_TYPE" = "password" ] || exit 0
[ "$PAM_USER" = "root" ] || exit 0

# extract hash and save to location used for sync
```

File: recipes-core/initrdscripts/files/ni_provisioning.safemode.common (hunk 7 above)
Comment (erickshepherdNI) [reply recipes-core/initrdscripts/files/ni_provisioning.safemode.common]: I've updated the scripts to only store the shadow hash in .shadow and have removed any reference to the passwd file.

File: recipes-ni/ni-acctsync/ni-acctsync_1.0.bb (hunk 6 above)
Comment (erickshepherdNI) [reply recipes-ni/ni-acctsync/ni-acctsync_1.0.bb]: When I replace update-rc.d-native with update-rc.d I get a build error for "update-rc.d: not found". It looks like every recipe that directly calls update-rc.d depends on update-rc.d-native.

File: recipes-ni/ni-acctsync/files/ni-acctsync [hunk 17]
```diff
@@ -0,0 +1,106 @@
...
+}
+
+check_hash_equality()
+{
+    SHARED_HASH_PATH="$1"
+    LOCAL_HASH_PATH="$2"
+    LOCAL_ROOT=$(get_root_line "$LOCAL_HASH_PATH")
+
+    LOCAL_HASH=$(echo "$LOCAL_ROOT" | cut -d':' -f2)
+    read -r SHARED_HASH < $SHARED_HASH_PATH
```
gratian: Minor nitpick. We know the path is not going to have spaces or other weird chars in this case but is still recommended to quote variables containing paths (and other expansions), it is cheap insurance i.e.,
```
read -r SHARED_HASH < "$SHARED_HASH_PATH"
```
Check other variables, there are several instances in this file, and add quotes. When in doubt, quote everything :)

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,106 @@
...
+    else
+        return 1
+    fi
+}
+
+insert_root_hash()
+{
+    SRC="$1"
+    DEST="$2"
+    TMP_SHADOW="/tmp/.shadow.tmp"
```
gratian: Copilot says we should use less predictable names here. I agree.
I have not looked into the trap part, seems like a good idea, but I don't know if it makes sense with the rest of the script.
```
2) Use a safer temporary-file pattern (avoid predictable /tmp names)

/tmp/.shadow.tmp and /tmp/root_hash.tmp are predictable. That’s a classic symlink/hardlink attack surface (especially risky because you later mv into /etc/shadow).

Use mktemp and a trap:

    TMP_SHADOW="$(mktemp /tmp/shadow.XXXXXX)"
    trap 'rm -f "$TMP_SHADOW" "$TMP_HASH"' EXIT HUP INT TERM

Also set restrictive permissions for temp files (umask).
```

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,106 @@
...
+
+    if check_hash_equality $SRC $DEST; then
+        log "Root hash in $SRC matches the one in $DEST, nothing to sync."
+    else
+        log "Syncing root info from $SRC to $DEST"
+        read -r SHADOW_HASH < $SRC
+        awk -F: -v OFS=":" -v shadow_hash="$SHADOW_HASH" '$1 == "root" { $2 = shadow_hash } { print }' "$DEST" > "$TMP_SHADOW"
+        mv $TMP_SHADOW $DEST
```
gratian: Another Copilot suggestion. We need to make sure we do not accidentally make access to `/etc/shadow` more permissive by doing the move from a temp file (especially if we don't have umask set). I think the way it is now that awk call will create a `$TMP_SHADOW` file with the default umask which is probably more permissive than `0600`
The suggestion is to create temp, then `chmod --reference="$DEST" "$TMP_SHADOW"` and `chown --reference="$DEST" "$TMP_SHADOW"` before replacing.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,106 @@
...
+
+    stop)
+        log "Syncing root info to shared shadow file..."
+
+        if [ -f "$SHARED_SHADOW" ]; then
+            log "Found shared shadow file. Syncing..."
+            extract_root_hash "$SYS_SHADOW" "$SHARED_SHADOW"
+            log "Sync complete."
+        else
+            log "Shared shadow file not found. Stopping sync."
```
gratian: Are there going to be situations where the shared shadow file doesn't exist, for example because the target was upgraded as opposed to re-provisioned?

Should we create the shared shadow file on `stop` if it doesn't exist?

gratian (approved):
Looks good in general. Approved conditionally pending testing results and resolving the couple nitpicks below.

amstewart (approved):

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,147 @@
...
+hash_sanity_check()
+{
+    SHADOW_HASH="$1"
+    
+    if [ -z "$SHADOW_HASH" ] || [ "$SHADOW_HASH" = "*" ] || [ "$SHADOW_HASH" = "!" ]; then
+        return 0
+    fi
+    
+    if [ -f "/etc/login.defs" ]; then
+        ENCRYPT_METHOD=$(grep -E '^[[:space:]]*ENCRYPT_METHOD[[:space:]]+' "/etc/login.defs" | awk '{print $2}')
```
gratian: Nitpick, you can avoid the extra call to grep and the pipe if you use awk directly for parsing:

```
ENCRYPT_METHOD=$(awk '$1=="ENCRYPT_METHOD" { print $2; exit }' /etc/login.defs)
```

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,147 @@
...
+            return 1
+            ;;
+    esac
+}
+
+insert_root_hash()
+{
+    SRC="$1"
+    DEST="$2"
+    TMP_SHADOW="$(mktemp /tmp/shadow.XXXXXX)"
```
gratian: We should move creating the temp file in the 'else' case below since it is only used if the hashes don't match.  Alternatively we can check_hash_equality first and exit early.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,147 @@
...
+            mv "$TMP_SHADOW" "$DEST"
+        fi
+    fi
+}
+
+extract_root_hash()
+{
+    SRC="$1"
+    DEST="$2"
+    TMP_HASH="$(mktemp /tmp/root_hash.XXXXXX)"
```
gratian: Same as above the TMP_HASH file is only used when the hashes don't match. I'm not sure if the trap cleans it up on exit but it's probably better to avoid creating it in the first place if it is not needed i.e. if the hashes match.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,147 @@
...
+    trap 'rm -f "$TMP_HASH"' EXIT HUP INT TERM
+    SRC_ROOT=$(get_root_line "$SRC")
+
+    if check_hash_equality "$DEST" "$SRC"; then
+        log "Local root hash matches the shared one, nothing to sync."
```
gratian: Nitpick, we should probably avoid unnecessary logging for the "nothing happened" cases, especially since we don't have a corresponding log call for the else case where we are doing the sync.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -0,0 +1,147 @@
...
+    else
+        echo "$SRC_ROOT" | cut -d':' -f2 > "$TMP_HASH"
```
gratian: Should this be moved after the chmod/chown calls to avoid exposing the hash accidentally if permissions are not set correctly at this point?

gratian (approved):
Looks good, just a couple nitpicks.

recipes-ni/ni-acctsync/files/ni-acctsync
erickshepherdNI: Applied this change and tested that ENCRYPT_METHOD is still correctly parsed by changing the login.defs files.

recipes-ni/ni-acctsync/files/ni-acctsync
erickshepherdNI: Makes sense, updated the script to create the file in the if structure.

recipes-ni/ni-acctsync/files/ni-acctsync
erickshepherdNI: I updated the script so that the "if" case is executed when there are actions to take and removed the else case entirely since the only thing that would go there is this logging.

recipes-ni/ni-acctsync/files/ni-acctsync
erickshepherdNI: Applied this change.

gratian (approved):

PR #942, opened 2026-02-11
meta-nilrt: sysapi uses SSH for communications in all modes

Nasrulla-NI:
### Summary of Changes
ashd initscript (or unit file) ignores sshd.enabled in /etc/natinst/share/ni.rt.ini in runmode and safemode. 
SSH uixml file removed from safemode so that it can no longer be configured from MAX. 

### Justification
https://dev.azure.com/ni/DevCentral/_workitems/edit/3470276

### Testing
Test Report: 
[Feature_3470276_Test_Report.docx](https://github.com/user-attachments/files/25225573/Feature_3470276_Test_Report.docx)

- [x] - I have done manual testing on Crio 9045 and 9030. 
- [x] - Verified sysapi is included in safe mode and shh is enabled in all modes
- [x] - Verified, as shh shall be no longer configured from MAX.

### Note
After installing the "nilrt-safemode-rootfs-x64.tar.gz", a power cycle may be needed for updates to be seen on the Target Hardware

### Procedure
- [x]  I certify that the contents of this pull request complies with the [Developer Certificate of Origin](https://github.com/ni/nilrt/blob/HEAD/docs/CONTRIBUTING.md#developer-certificate-of-origin-dco).

PR #945, opened 2026-02-19
meta-nilrt: sysapi, ssh, uixml changes

Nasrulla-NI:
### Summary of Changes
This PR is to make sures the below snac feature is implemented.
- [x] SSH daemon is enabled in NILRT safemode.
- [x] ni-sysapi-sshcli package is included.
- [x] SSH uixml file removed from safemode so that it can no longer be configured from MAX.

### Justification
This is one of the snac feature, kindly refer below feature azure link for more details
https://dev.azure.com/ni/DevCentral/_workitems/edit/3470276

### Testing
Uploaded the latest image on CRIO 9045  and Verified SSH is default enabled in all modes, and SSH is no longer configured from MAX.

Safe_Mode_sysapi_sshcli_installed

Ni_Max_SSH_Not_Configurable_Safe_Mode

### Procedure

* [x] I certify that the contents of this pull request complies with the [Developer Certificate of Origin]

### DIFF
--- a/recipes-connectivity/openssh/openssh_%.bbappend
+++ b/recipes-connectivity/openssh/openssh_%.bbappend
@@ -16,13 +16,9 @@ do_patch_oe_source () {
 addtask patch_oe_source after do_patch before do_configure
 
 do_install:append () {
-	sed 's|check_for_no_start() {|&\
-	# if sshd is not enabled in ni-rt.ini, do not start sshd\
-	enable=`/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=sshd.enabled,value="false" \|tr "[:upper:]" "[:lower:]"`\
-	if [ "$enable" != "true" ]; then\
-		[ "${VERBOSE}" != "no" ] \&\& echo "SSHD not enabled in ni-rt.ini"\
-		exit 0\
-	fi|' -i ${D}${sysconfdir}/init.d/sshd
+	sed -i '/check_for_no_start() {/a \
+	# Ignore sshd.enabled in /etc/natinst/share/ni.rt.ini\
+	enable="true"' ${D}${sysconfdir}/init.d/sshd
 
 	# customize sshd_config
 	sed -e 's|^[#[:space:]]*Banner .*|Banner /etc/issue.net|' \
--- a/recipes-core/packagegroups/packagegroup-ni-safemode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-safemode.bb
@@ -18,7 +18,6 @@ RDEPENDS:${PN} = " \
 	ni-netcfgutil \
 	ni-shutdown-guard-safemode \
 	ni-systemimage \
-	sysconfig-settings-ssh \
 "
 
 RDEPENDS:${PN}:append:xilinx-zynq = " \
--- a/recipes-core/images/includes/nilrt-proprietary.inc
+++ b/recipes-core/images/includes/nilrt-proprietary.inc
@@ -16,6 +16,7 @@ NI_PROPRIETARY_COMMON_PACKAGES = "\
         ni-software-installation-websvc \
         ni-ssl-webserver-support \
         ni-sysapi \
+        ni-sysapi-sshcli \
         ni-sysapi-remote \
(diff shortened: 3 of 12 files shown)


Nasrulla-NI (commented):
meta-nilrt: sysapi, ssh, uixml changes

review comments from "https://github.com/ni/meta-nilrt/pull/942" are fixed here. as old PR is closed due wrong branch merge request.

recipes-connectivity/openssh/openssh_%.bbappend
```diff
@@ -16,13 +16,9 @@ do_patch_oe_source () {
...
 
 do_install:append () {
-	sed 's|check_for_no_start() {|&\
-	# if sshd is not enabled in ni-rt.ini, do not start sshd\
-	enable=`/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=sshd.enabled,value="false" \|tr "[:upper:]" "[:lower:]"`\
-	if [ "$enable" != "true" ]; then\
-		[ "${VERBOSE}" != "no" ] \&\& echo "SSHD not enabled in ni-rt.ini"\
-		exit 0\
-	fi|' -i ${D}${sysconfdir}/init.d/sshd
+	sed -i '/check_for_no_start() {/a \
```
chaitu236, line 19: I don't understand the point of this - it prepends `enable="true"` to `check_for_no_start` but right on the next line `enable` is set to `/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=sshd.enabled,value="false" |tr "[:upper:]" "[:lower:]"`.

What are we trying to do here?

recipes-core/packagegroups/packagegroup-ni-safemode.bb
```diff
@@ -18,7 +18,6 @@ RDEPENDS:${PN} = " \
 	ni-netcfgutil \
 	ni-shutdown-guard-safemode \
 	ni-systemimage \
-	sysconfig-settings-ssh \
```
chaitu236, line 21: What happens when [nisyscfg](https://dev.azure.com/ni/DevCentral/_git/Tools?version=GC1855b697801f7ba79fe58cb0e58be9585a25bbdb&path=/shared/nisyscfg/nisyscfg/nisyscfg.py)'s `enable_ssh` or `disable_ssh` are called on a target with these changes?

recipes-connectivity/openssh/openssh_%.bbappend
Nasrulla-NI: - check_for_no_start() is the function which decides wheather ssh should be enabled or not in /etc/init.d/sshd file based on the "enable" variable.

- the older code was reading the /etc/natinst/share/ni.rt.in based and store the status into "enable" variable, then decision was taken to start or stop ssh after reboot. 

- as the requirement says to start ssh default during the safemode I have removed this check to verify "/etc/natinst/share/ni.rt.in" and forcefully setting it to "True" by appending "enable = True" inside this check_for_no_start().

chaitu236: Ah I ran the sed command on an already processed file (which was already setting `enable` to `/usr/local/natinst/bin/nirtcfg ..`) so came to wrong conclusion. But we're removing that altogether here - makes sense.

recipes-core/packagegroups/packagegroup-ni-safemode.bb
Nasrulla-NI: We tested this on a VM using both the host machine and the mobilize command. After enabling and disabling SSH using nisyscfg, all test cases passed successfully.
Enabling or disabling SSH simply involves updating the enable = TRUE/FALSE parameter in /etc/natinst/share/ni-rt.ini.
This modification does not affect the system because SSH is enabled by default and bypasses the configuration in /etc/natinst/share/ni-rt.ini, as required.
For reference, please see the dashboard link below:
https://dashboard.ni.systems/dashboard/#/mobilize?plans=699ec382a2a09a2e1f61bd56&plan=699ec382a2a09a2e1f61bd56

chaitu236 (approved):

PR #946, opened 2026-02-19
Move /var/log permission changes to BSI

texasaggie97:
### Summary of Changes

- Added `su root adm` to the global logrotate configuration so that logrotate
  accepts `/var/log` being owned by group `adm` with mode 0770, fixing the error:
  `skipping "/var/log/dmesg" because parent directory has insecure permissions`.
- Updated `/etc/default/volatiles/00_core` entries in `ni-configpersistentlogs`
  to create `/var/log` with owner `root:adm` and mode `0770` (both persistent and
  volatile configurations).
- Added a new SysV init script `set_log_permissions.sh` (runs at start priority 3)
  that enforces `root:adm 0770` ownership and default ACLs on `/var/log` at boot
  time, keeping the base image consistent with the security hardening applied by
  nilrt-snac.
- Added `acl` as a runtime dependency of `ni-configpersistentlogs` to ensure
  `setfacl` is available.

### Justification

[AB#3698573](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3698573)

The `nilrt-snac` security hardening script sets `/var/log` to `root:adm 0770`
with default ACLs. Without this change the base image is inconsistent with that
policy and logrotate fails on any log under `/var/log` with a misleading
"insecure permissions" error.

See this [internal Teams chat](https://teams.microsoft.com/l/message/19:00d87ceefac44440821b903817b341c3@thread.tacv2/1770926793743?tenantId=eb06985d-06ca-4a17-81da-629ab99f6505&groupId=b028d9eb-0284-4d3f-a159-a4b340eb62df&parentMessageId=1770926793743&teamName=TM-LabVIEW%3A%20Engineering%20%2B%20Services&channelName=Engineering&createdTime=1770926793743) for discussion regarding this change and how it would affect LabVIEW RT.

### Testing

### Procedure

### DIFF
--- a/recipes-ni/ni-configpersistentlogs/files/set_log_permissions.sh
+++ b/recipes-ni/ni-configpersistentlogs/files/set_log_permissions.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+LOG_DIR="/var/log"
+
+if [ -d "${LOG_DIR}" ]; then
+	chown root:adm "${LOG_DIR}"
+	chmod 0770 "${LOG_DIR}"
+	if command -v setfacl >/dev/null 2>&1; then
+		setfacl -d -m g:adm:rwx "${LOG_DIR}"
+		setfacl -d -m o::0 "${LOG_DIR}"
+	fi
+fi
+
+exit 0
--- a/recipes-ni/ni-configpersistentlogs/ni-configpersistentlogs.bb
+++ b/recipes-ni/ni-configpersistentlogs/ni-configpersistentlogs.bb
@@ -8,6 +8,7 @@ PV = "1.0"
 
 SRC_URI = "\
 	file://ni-configpersistentlogs \
+	file://set_log_permissions.sh \
 "
 
 S = "${WORKDIR}"
@@ -20,11 +21,18 @@ inherit update-rc.d
 do_install () {
 	install -d ${D}${sysconfdir}/init.d/
 	install -m 0755 ${S}/ni-configpersistentlogs ${D}${sysconfdir}/init.d/
+	install -m 0755 ${S}/set_log_permissions.sh ${D}${sysconfdir}/init.d/
 }
 
 
 FILES:${PN} += "\
 	${sysconfdir}/init.d/ni-configpersistentlogs \
+	${sysconfdir}/init.d/set_log_permissions.sh \
 "
 
-RDEPENDS:${PN} += "bash initscripts"
+RDEPENDS:${PN} += "bash initscripts acl"
+DEPENDS += "update-rc.d-native"
+
+do_install:append () {
+	update-rc.d -r ${D} set_log_permissions.sh start 3 S .
+}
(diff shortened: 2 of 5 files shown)


recipes-ni/ni-configpersistentlogs/files/set_log_permissions.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+LOG_DIR="/var/log"
+
+if [ -d "${LOG_DIR}" ]; then
+	chown root:adm "${LOG_DIR}"
+	chmod 0770 "${LOG_DIR}"
```
chaitu236, line 7: Is it ok removing rx permissions for others? I know that [technical support report](https://nilrt-docs.ni.com/troubleshooting/technicalreport.html) reads from this location and I don't remember if it's done as webserv or admin; have we tested that workflow?

@jpautler  may also have more context on persistent logs implementation.

chaitu236 (changes requested):

recipes-ni/ni-configpersistentlogs/files/set_log_permissions.sh
texasaggie97: We make this change in ownership in SNAC mode which is then causing issues with logrotate. This change moves that ownership change into the BSI so logrotate can be configured so that it will work in SNAC mode as well. If this is an issue for something, know that it will also be an issue when the target is configured for SNAC.

recipes-ni/ni-configpersistentlogs/ni-configpersistentlogs.bb
```diff
@@ -8,6 +8,7 @@ PV = "1.0"
 
 SRC_URI = "\
 	file://ni-configpersistentlogs \
+	file://set_log_permissions.sh \
```
amstewart, line 11: The original `/var/log` permissions are applied by the [populate-volatile](https://github.com/ni/openembedded-core/blob/f6022dd5d0171b09adb8daa1c3829355d93c17af/meta/recipes-core/initscripts/initscripts-1.0/populate-volatile.sh) initscript already - which reads its configuration from `/etc/default/volatiles`. Instead of adding another initscript that resets it later, we should just add a conf snippet to that location changing the ownership.

amstewart: And you should make that change in [our initscripts bbappend](https://github.com/ni/meta-nilrt/tree/nilrt/master/scarthgap/recipes-core/initscripts).

amstewart (changes requested):
Remember to clean up your commit messages to include the recipe in the commit subject and have a detailed description.

recipes-ni/ni-configpersistentlogs/files/set_log_permissions.sh
chaitu236: I think we need to test technical support report workflow to ensure it's not broken. We should also spend some time figuring out if any other workflows will be broken (or already broken in SNAC).

PR #959, opened 2026-03-05
Use u-boot's fw_printenv instead of libubootenv's

chaitu236:
### Summary of Changes

For MTD partitions that use UBI, `fw_printenv` from `libubootenv` only works at UBI layer and acquires exclusive lock of the volume during `fw_setenv` which causes `fw_printenv` to fail. This causes issues for other programs that expect these programs to always succeed.

`fw_printenv` from `u-boot` can operate at `MTD` layer and doesn't lock volume. So concurrent `fw_setenv` and `fw_printenv` calls can work.

So switch to `u-boot`'s `fw_printenv` to keep the same behavior we shipped with sumo.

Also update all `u-boot-fw-utils` dependencies to `fw-printenv`.

### Justification

WI: [AB#3737069](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3737069)

### Testing

- [x] I have built the core package feed with this PR in place. (`bitbake packagefeed-ni-core`)
- [x] Built BSI and installed on cRIO-9068.
- [x] Ran `fw_printenv` in a loop while `fw_setenv` was running in a loop and didn't see `fw_printenv` failing. This would fail with `libubootenv`-provided `fw_printenv` and `fw_setenv`.
- [x] `opkg install kernel-containerized-performance-tests-ptest kernel-performance-tests-ptest ni-test-boot-time-ptest` works.

### Procedure

### DIFF
--- a/recipes-bsp/u-boot/libubootenv_%.bbappend
+++ b/recipes-bsp/u-boot/libubootenv_%.bbappend
@@ -1,7 +1,5 @@
 COMPATIBLE_MACHINE = "xilinx-zynq"
-PROVIDES += "fw-printenv"
 DEPENDS += "niacctbase"
-RPROVIDES:${PN}-bin += "fw-printenv"
 RDEPENDS:${PN}-bin += "u-boot-env"
 
 do_install:append() {
--- a/recipes-bsp/u-boot/files/fw_env.config
+++ b/recipes-bsp/u-boot/files/fw_env.config
@@ -3,5 +3,5 @@
 # environment sector is assumed present.
 
 # Device name		Device offset	Env. size	Flash sector size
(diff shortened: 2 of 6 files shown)


recipes-bsp/u-boot/libubootenv_%.bbappend
```diff
@@ -1,7 +1,5 @@
 COMPATIBLE_MACHINE = "xilinx-zynq"
-PROVIDES += "fw-printenv"
```
chaitu236, line 2: Ideally we'd have a virtual provider for this and we'd choose which one is default in a conf file but I'll do that in a future PR time permitting.

gratian (approved):

PR #962, opened 2026-03-06
libpam: Add pam hook to sync passwords

erickshepherdNI:
### Summary of Changes

Update libpam_1.%.bbappend to add this line to /etc/pam.d/common-password:
`password	optional			pam_exec.so /usr/sbin/ni-acctsync-pam`
 This will execute '/etc/init.d/ni-acctsync stop' whenever a password is changed. Also add the pam-plugin-exec package to images.

[AB#3698760](https://dev.azure.com/ni/DevCentral/_workitems/edit/3698760)

### Justification

This will allow ni-acctsync to immediately update the shared .shadow file when the root password is changed instead of waiting for the system to shut down. This makes the password data stored in the configs partition more secure against cases where the device loses power before a new password can be synced.

### Testing

* [X] Built the base runmode and safemode images and confirmed that /etc/pam.d/common-password contains the expected line
* [X] Updated the root password via the passwd command and confirmed /etc/natinst/share/.shadow was immediately updated as well. Tested in runmode and safemode.
* [X] Tested that changing the password for non-root users does not trigger the ni-acctsync script

### Procedure

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-safemode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-safemode.bb
@@ -15,7 +15,6 @@ RDEPENDS:${PN} = " \
 	e2fsprogs-e2fsck \
 	e2fsprogs-mke2fs \
 	e2fsprogs-tune2fs \
-	ni-acctsync \
 	ni-netcfgutil \
 	ni-shutdown-guard-safemode \
 	ni-systemimage \
--- a/recipes-extended/pam/libpam_1.%.bbappend
+++ b/recipes-extended/pam/libpam_1.%.bbappend
@@ -1,11 +1,16 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
 
+RDEPENDS:${PN} += "ni-acctsync pam-plugin-exec"
+
 SRC_URI += "\
 	file://security/faillock.conf \
+	file://scripts/ni-acctsync-pam \
 "
 
 do_install:append() {
 	install -m 644 ${UNPACKDIR}/security/faillock.conf ${D}${sysconfdir}/security/faillock.conf
+	install -m 700 ${UNPACKDIR}/scripts/ni-acctsync-pam ${D}${sbindir}/ni-acctsync-pam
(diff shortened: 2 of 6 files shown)


recipes-core/packagegroups/packagegroup-ni-safemode.bb
```diff
@@ -19,6 +19,7 @@ RDEPENDS:${PN} = " \
 	ni-netcfgutil \
 	ni-shutdown-guard-safemode \
 	ni-systemimage \
+	pam-plugin-exec \
```
chaitu236: Packages common to safemode and runmode could be put in `packagegroup-ni-base`. I just realized that we could've done the same for `ni-acctsync`.

gratian (approved):
Approved, conditional on testing the script only runs on password changes and not every login.

recipes-extended/pam/libpam_1.%.bbappend
```diff
@@ -6,6 +6,7 @@ SRC_URI += "\
 
 do_install:append() {
 	install -m 644 ${UNPACKDIR}/security/faillock.conf ${D}${sysconfdir}/security/faillock.conf
+	sed -E -i '/^password[[:space:]]+requisite[[:space:]]+pam_deny\.so$/a password\toptional\t\t\tpam_exec.so /etc/init.d/ni-acctsync stop' "${D}${sysconfdir}/pam.d/common-password"
```
chaitu236: If libpam is going to expect `/etc/init.d/ni-acctsync` being present, does it make more sense to add RDEPENDS in libpam to `ni-netcfgutil` and `pam-plugin-exec` instead of adding  `ni-netcfgutil` and `pam-plugin-exec` explicitly to different packagegroups?

erickshepherdNI: Do you mean `ni-acctsync` instead of `ni-netcfgutil`? If we add `ni-acctsync` and `pam-plugin-exec` as dependencies for `libpam`, would we still need to add them to the base package group like you mention in your other comment?

chaitu236: Yeah sorry I meant `ni-acctsync` and not `ni-netcfgutil`.

erickshepherdNI: I've updated the PR to exclude ni-acctsync and pam-plugin-exec from the packagegroup files and added them to the libpam RDEPENDS instead.

chaitu236 (approved):

gratian (approved):

PR #965, opened 2026-03-09
meta-nilrt: udev-extraconf: update worldwritable permissions

Nasrulla-NI:
### Summary of Changes

Currently NILRT will automatically mount block and TTY devices as world-writable and should be updated to no longer do this.
We should try to avoid just changing the permissions and should instead have some way to toggle the world-writable permissions back on if needed. Hence to do this - /etc/natinst/share/ni-rt.ini - worldwritable.enabled="true" or worldwritable.enabled="false" is appended as per information provided from hardware config team/

- Controls whether VFAT/TTY mounts are world-writable.
- 0 = default restrictive (fmask=0002,dmask=0002)
- 1 = world-writable (fmask=0000,dmask=0000)
- Reads /** /etc/natinst/share/ni-rt.ini - worldwritable.enabled="true" or worldwritable.enabled="false" ** 

### Justification

Please refer below feature
https://dev.azure.com/ni/DevCentral/_workitems/edit/3470239

### Testing
Safemode:
VFAT and TTY: 
**- worldwritable.enabled="false"**
worldwritable enabled_false

**worldwritable.enabled="true"**
worldwritable enabled_true

Runmode:
VFAT and TTY: 
**- worldwritable.enabled="false"**
worldwritable_runmode_false

**worldwritable.enabled="true"**
worldwritable_runmode_true

### Procedure

### DIFF
--- a/recipes-core/udev/udev-extraconf/rules.d/usb-serial-permissions.rules
+++ b/recipes-core/udev/udev-extraconf/rules.d/usb-serial-permissions.rules
@@ -1,2 +1,10 @@
-SUBSYSTEM=="tty", KERNEL=="ttyUSB*", GROUP="tty", MODE="0666"
-SUBSYSTEM=="tty", KERNEL=="ttyACM*", GROUP="tty", MODE="0666"
+# Import runtime toggle (NI_WORLDWRITABLE) from /etc/udev/scripts/ni-worldwritable.sh
+IMPORT{program}="/etc/udev/scripts/ni-worldwritable.sh"
+
+# When NI_WORLDWRITABLE=1, keep world-writable (0666)
+SUBSYSTEM=="tty", KERNEL=="ttyUSB*", ENV{NI_WORLDWRITABLE}=="1", MODE:="0666", GROUP="dialout"
+SUBSYSTEM=="tty", KERNEL=="ttyACM*", ENV{NI_WORLDWRITABLE}=="1", MODE:="0666", GROUP="dialout"
+
+# Otherwise default to 0664 (no world-write)
+SUBSYSTEM=="tty", KERNEL=="ttyUSB*", ENV{NI_WORLDWRITABLE}!="1", MODE:="0664", GROUP="dialout"
+SUBSYSTEM=="tty", KERNEL=="ttyACM*", ENV{NI_WORLDWRITABLE}!="1", MODE:="0664", GROUP="dialout"
--- a/recipes-core/udev/udev-extraconf/scripts/mount.sh
+++ b/recipes-core/udev/udev-extraconf/scripts/mount.sh
@@ -51,10 +51,23 @@ automount() {
 		MOUNT="$MOUNT -o silent"
 	fi
 
+	# --- Runtime configuration for VFAT permissions ---
+	# Read worldwritable.enabled from /etc/natinst/share/ni-rt.ini, if available.
+	# False (default) => fmask=0002,dmask=0002  (group-writable, not world-writable)
+	# True            => fmask=0000,dmask=0000  (world-writable)
+	enable=$(/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=worldwritable.enabled,value="false" | tr "[:upper:]" "[:lower:]")
+
+	# Default VFAT options (restrictive)
+	VFAT_OPTS="fmask=0002,dmask=0002,quiet"
+	case "${enable}" in
+	    true)
+			VFAT_OPTS="fmask=0000,dmask=0000,quiet" ;;
+	esac
+
 	case "$ID_FS_TYPE" in
 		# If mounting vfat, set partition to be world-writable, and
 		# disable permissions warnings
-		vfat) MOUNT="$MOUNT -o fmask=0000,dmask=0000,quiet" ;;
+		vfat) MOUNT="$MOUNT -o $VFAT_OPTS" ;;
 	esac
 
 	if ! $MOUNT -t auto $DEVNAME "/media/$name"
--- a/recipes-core/udev/udev-extraconf/scripts/ni-worldwritable.sh
+++ b/recipes-core/udev/udev-extraconf/scripts/ni-worldwritable.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+# Echo udev environment keys for IMPORT{program}
+# Reads /etc/natinst/share/ni-rt.ini and outputs worldwritable.enabled=<true|false>
+
+enable=$(/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=worldwritable.enabled,value="false" | tr "[:upper:]" "[:lower:]")
+
+case "$enable" in
+     true) NI_WORLDWRITABLE_VAL=1 ;;
+    *)     NI_WORLDWRITABLE_VAL=0 ;;
+esac
+
+# udev IMPORT{program} expects KEY=VALUE pairs on stdout
+echo "NI_WORLDWRITABLE=$NI_WORLDWRITABLE_VAL"
+exit 0


amstewart (changes requested):
>We should try to avoid just changing the permissions and should instead have some way to toggle the world-writable permissions back on if needed.

Why? Because of VISA? VISA has to run as a user that isn't root or a member of the `dialout` group?

Nasrulla-NI:
> > We should try to avoid just changing the permissions and should instead have some way to toggle the world-writable permissions back on if needed.
> 
> Why? Because of VISA? VISA has to run as a user that isn't root or a member of the `dialout` group?

Thanks for the comment. Indeed, the change was aimed at supporting VISA, since it often runs as a non‑root user without dialout membership and needs access to these device nodes. So, added a config/feature flag so the relaxed permissions are enabled only when needed. This will be controlled from front end Hwconfig.

recipes-core/udev/udev-extraconf/scripts/ni-worldwritable.sh
```diff
@@ -0,0 +1,16 @@
...
+
+case "$enable" in
+    1|yes|true|TRUE|Yes|True) NI_WORLDWRITABLE_VAL=1 ;;
+    *)                   NI_WORLDWRITABLE_VAL=0 ;;
+esac
+
+# udev IMPORT{program} expects KEY=VALUE pairs on stdout
+echo "NI_WORLDWRITABLE=$NI_WORLDWRITABLE_VAL"
+exit 0
+SH
```
rajendra-desai-ni: Missing whiteline

recipes-core/udev/udev-extraconf/scripts/ni-worldwritable.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+# Echo udev environment keys for IMPORT{program}
+# Reads /etc/natinst/share/ni-rt.ini and outputs worldwritable.enabled=<0|1>
+
+
+enable=`/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=worldwritable.enabled,value="false" \|tr "[:upper:]" "[:lower:]"`\
+
+case "$enable" in
+    1|yes|true|TRUE|Yes|True) NI_WORLDWRITABLE_VAL=1 ;;
```
rajendra-desai-ni: What is the need of checking various cases when you are already converting upper case to lower case string above?
Also, it is better to have just "true/false" Boolean type instead of other values like 1/0 or yes/no. See if you can update the case statement accordingly.

recipes-core/udev/udev-extraconf/scripts/ni-worldwritable.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+# Echo udev environment keys for IMPORT{program}
+# Reads /etc/natinst/share/ni-rt.ini and outputs worldwritable.enabled=<0|1>
```
rajendra-desai-ni: `worldwritable.enabled=<true|false>` would be better I guess.

recipes-core/udev/udev-extraconf/scripts/mount.sh
```diff
@@ -51,10 +52,21 @@ automount() {
...
 
+	# --- Runtime configuration for VFAT permissions ---
+	# Read worldwritable.enabled from /etc/natinst/share/ni-rt.ini, if available.
+	# False (default) => fmask=0002,dmask=0002  (group-writable, not world-writable)
+	# True            => fmask=0000,dmask=0000  (world-writable)
+	enable=`/usr/local/natinst/bin/nirtcfg --get section=SystemSettings,token=worldwritable.enabled,value="false" \|tr "[:upper:]" "[:lower:]"`\
+	# Default VFAT options (restrictive)
+	VFAT_OPTS="fmask=0002,dmask=0002,quiet"
+	case "${enable}" in
+	    1|yes|true|TRUE|Yes|True) VFAT_OPTS="fmask=0000,dmask=0000,quiet" ;;
```
rajendra-desai-ni: What is the need of checking various cases when you are already converting upper case to lower case string above?
Also, it is better to have just "true/false" Boolean type instead of other values like 1/0 or yes/no. See if you can update the case statement accordingly.

recipes-core/udev/udev-extraconf/scripts/mount.sh
```diff
@@ -12,6 +12,7 @@ sgdisk_sed="s/^\/dev\/((sd[a-z]+)|(mmcblk[0-9]+)p)([0-9]+)$/sgdisk -i \4 \/dev\/
 blacklisted_gpt_part_guids_egrep="(C12A7328-F81F-11D2-BA4B-00A0C93EC93B)|(4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709)"
 export PATH="$PATH:/usr/sbin"
 
+
```
rajendra-desai-ni: Remove this extra line

recipes-core/udev/udev-extraconf/rules.d/usb-serial-permissions.rules
```diff
@@ -1,2 +1,10 @@
...
-SUBSYSTEM=="tty", KERNEL=="ttyACM*", GROUP="tty", MODE="0666"
+# Import runtime toggle (NI_WORLDWRITABLE) from /etc/udev/scripts/ni-worldwritable.sh
+IMPORT{program}="/etc/udev/scripts/ni-worldwritable.sh"
+
+# When NI_WORLDWRITABLE=1, keep world-writable (0666)
+SUBSYSTEM=="tty", KERNEL=="ttyUSB*", ENV{NI_WORLDWRITABLE}=="1", MODE:="0666", GROUP="dialout"
+SUBSYSTEM=="tty", KERNEL=="ttyACM*", ENV{NI_WORLDWRITABLE}=="1", MODE:="0666", GROUP="dialout"
+
+# Otherwise default to 0664 (no world-write)
+SUBSYSTEM=="tty", KERNEL=="ttyUSB*", ENV{NI_WORLDWRITABLE}!="1", MODE:="0664", GROUP="dialout"
```
rajendra-desai-ni, line 9: Since NI_WORLDWRITABLE would only have 2 values 1 or 0, can we just be clear with `=="0"` instead of `!="1"`?

rajendra-desai-ni (changes requested):
Commit message changes:
- Change commit title to be more specific, something like:
`udev-extraconf: update worldwritable permissions`
- Update commit message to have more info about the changes
- Signed-off tag should be at the end

recipes-core/udev/udev-extraconf/scripts/mount.sh
```diff
@@ -51,10 +51,22 @@ automount() {
...
+	    true)
+			VFAT_OPTS="fmask=0000,dmask=0000,quiet" ;;
+	esac
+
 	case "$ID_FS_TYPE" in
 		# If mounting vfat, set partition to be world-writable, and
 		# disable permissions warnings
-		vfat) MOUNT="$MOUNT -o fmask=0000,dmask=0000,quiet" ;;
+		vfat) MOUNT="$MOUNT -o $VFAT_OPTS" ;;
 	esac
```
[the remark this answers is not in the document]
chaitu236: I think moving it within vfat block would make the code less readable. So what you've here is fine.

recipes-core/udev/udev-extraconf/rules.d/usb-serial-permissions.rules
```diff
@@ -1,2 +1,10 @@
...
+# Import runtime toggle (NI_WORLDWRITABLE) from /etc/udev/scripts/ni-worldwritable.sh
+IMPORT{program}="/etc/udev/scripts/ni-worldwritable.sh"
+
+# When NI_WORLDWRITABLE=1, keep world-writable (0666)
+SUBSYSTEM=="tty", KERNEL=="ttyUSB*", ENV{NI_WORLDWRITABLE}=="1", MODE:="0666", GROUP="dialout"
+SUBSYSTEM=="tty", KERNEL=="ttyACM*", ENV{NI_WORLDWRITABLE}=="1", MODE:="0666", GROUP="dialout"
+
+# Otherwise default to 0664 (no world-write)
+SUBSYSTEM=="tty", KERNEL=="ttyUSB*", ENV{NI_WORLDWRITABLE}=="0", MODE:="0664", GROUP="dialout"
+SUBSYSTEM=="tty", KERNEL=="ttyACM*", ENV{NI_WORLDWRITABLE}=="0", MODE:="0664", GROUP="dialout"
```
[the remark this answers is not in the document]
chaitu236: @Nasrulla-NI can you comment on whether `dialout` is required or if it's ok to switch to `tty` like copilot says? If we're not sure and `dialout` is what we already had, we don't need to dig too deep and can drop the issue.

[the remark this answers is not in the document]
chaitu236: I know you changed this from `!=1` previously but I think it's better - like copilot says when the script is missing or not runnable.

chaitu236 (changes requested):
Other than copilot suggestions, looks good. I don't have any other feedback.

recipes-core/udev/udev-extraconf/scripts/ni-worldwritable.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+# Echo udev environment keys for IMPORT{program}
+# Reads /etc/natinst/share/ni-rt.ini and outputs worldwritable.enabled=<true|false>
```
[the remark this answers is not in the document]
Nasrulla-NI: As the comment suggests "#Reads /etc/natinst/share/ni-rt.ini and outputs worldwritable.enabled=<true|false>"  this is an attribute used by Hardware Config and stored in /etc/natinst/share/ni-rt.ini.  A common convection is followed as per the "ni-rt.ini"

recipes-core/udev/udev-extraconf/rules.d/usb-serial-permissions.rules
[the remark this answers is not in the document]
Nasrulla-NI: Hi @chaitu236,
dialout was chosen intentionally to allow non‑root user access to serial devices. For our setup, this is tied to NI‑VISA, which relies on user‑level permissions for serial access. Based on that, keeping dialout seems safer to avoid breaking existing usage.

recipes-core/udev/udev-extraconf/rules.d/usb-serial-permissions.rules
```diff
@@ -1,2 +1,10 @@
...
-SUBSYSTEM=="tty", KERNEL=="ttyACM*", GROUP="tty", MODE="0666"
+# Import runtime toggle (NI_WORLDWRITABLE) from /etc/udev/scripts/ni-worldwritable.sh
+IMPORT{program}="/etc/udev/scripts/ni-worldwritable.sh"
+
+# When NI_WORLDWRITABLE=1, keep world-writable (0666)
+SUBSYSTEM=="tty", KERNEL=="ttyUSB*", ENV{NI_WORLDWRITABLE}=="1", MODE:="0666", GROUP="dialout"
+SUBSYSTEM=="tty", KERNEL=="ttyACM*", ENV{NI_WORLDWRITABLE}=="1", MODE:="0666", GROUP="dialout"
+
+# Otherwise default to 0664 (no world-write)
+SUBSYSTEM=="tty", KERNEL=="ttyUSB*", ENV{NI_WORLDWRITABLE}!="0", MODE:="0664", GROUP="dialout"
```
chaitu236: `!="1"`

chaitu236 (changes requested):

recipes-core/udev/udev-extraconf/rules.d/usb-serial-permissions.rules
```diff
@@ -1,2 +1,10 @@
...
+SUBSYSTEM=="tty", KERNEL=="ttyACM*", ENV{NI_WORLDWRITABLE}!="0", MODE:="0664", GROUP="dialout"
```
chaitu236: `!="1"`

chaitu236 (approved):

rajendra-desai-ni (approved):

PR #973, opened 2026-03-18
postinst: avoid false positives when scanning logs for errors

AshwinUjjwal:
### Summary of Changes

Replace deprecated egrep usage with grep -E and improve the log matching pattern to avoid false positives.

Previously, the pattern 'warn|error|fatal|fail' matched substrings such as 'str_error_r.o' and triggered failures due to the egrep deprecation warning.

Updated the command to use:
grep -Eiw 'warn(ing)?|error|fatal|fail'

This ensures only whole words are matched and includes both 'warn' and 'warning', reducing incorrect test failures.

### Justification

[AB#3429565](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3429565)

### Testing

TODO: Detail what testing has been done to ensure this submission meets requirements.

* [x] I validated this change by running the test script on target(RTOS-8862): /usr/lib/run-postinsts/ptest/test-run-postinsts-log.sh
image

### Note

This change should also be cherry‑picked into the dist-next and the relevant release branches.
 
### Procedure

### DIFF
--- a/recipes-devtools/run-postinsts/files/test-run-postinsts-log.sh
+++ b/recipes-devtools/run-postinsts/files/test-run-postinsts-log.sh
@@ -20,7 +20,7 @@ source "$CONFIG_FILE" || fail "Failed to run $CONFIG_FILE"
 
 # only present on first boot
 if [ -e "$LOGFILE" ]; then
-    if egrep -i 'warn|error|fatal|fail' "$LOGFILE"; then
+    if grep -Eiw 'warn(ing)?|error(s|ed)?|err|fatal|fail(ed|ure|ures)?' "$LOGFILE"; then
         fail "One or more errors in LOGFILE=$LOGFILE"
     fi
 fi


rajendra-desai-ni (changes requested):
Change looks good, but since the change is just in the test and not in the actual postinst script, it is better to update the commit title to `postinst-test:` or something instead of `postinst:`

Also, I guess this commit needs to be cherry-picked to dist-next and release branches. Can you update the same in the PR description?

recipes-devtools/run-postinsts/files/test-run-postinsts-log.sh
```diff
@@ -20,7 +20,7 @@ source "$CONFIG_FILE" || fail "Failed to run $CONFIG_FILE"
 
 # only present on first boot
 if [ -e "$LOGFILE" ]; then
-    if egrep -i 'warn|error|fatal|fail' "$LOGFILE"; then
+    if grep -Eiw 'warn(ing)?|error|fatal|fail' "$LOGFILE"; then
```
chaitu236: What about strings like `errors`, `failed`, `failure`, etc? I think it's better to have false positives rather than false negatives.

AshwinUjjwal: Updated the log scanning pattern to include additional keyword variations such as err, errored, errors, failed, and failure to improve coverage. Since log outputs are not fixed, included as many relevant keyword forms as possible to minimize false negatives.

Attempted to remove -w (using -Ei) for broader matching, but it introduced false positives from file names such as /lib/modules/.../str_error_r.o. Therefore, retained -Eiw to ensure meaningful matches .

chaitu236 (approved):
lgtm. But we won't be cherry-picking this into release branch. And we should wait until release is complete to cherry-pick into -next branch.

rajendra-desai-ni (approved):

PR #986, opened 2026-04-08
ni_provisioning: target disk from answers file

erickshepherdNI:
### Summary of Changes

There are two primary changes:

- Make disk selection possible using the answers file
    - This was done by adding a new PROVISION_TARGET_DISK that can be added to the answers file and will overwrite the current TARGET_DISK variable when present.
    - I also added some slight refactoring to read the answers file immediately after reading the script arguments instead of doing it in each branch of the following if-block.
- Add a interactive disk selection mode
    - This can be enablede by using a new -i argument for the ni_provisioning script when running it manually.
    - When enabled, a list of valid provisioning targets will be gathered using the same method we currently use to automatically select a target and each option will be printed followed by a prompt to select one.
        - There is a slight change to which block devices are valid, which is that removable devices are displayed in the interactive option.

The default behavior is unchanged, which is to say that if the default provisioning option is selected and no answer file is present, then the first valid block device that is found will be selected for provisioning.

### Justification

These changes will help support the upcoming Kinabalu devices that have multiple internal disks and require a way to select which one should be provisioned.

[AB#3813755](https://dev.azure.com/ni/DevCentral/_workitems/edit/3813755)

### Testing

For all of the testing I set up a cRIO-9033 and added an SD card that could be provisioned.

- [X] Tested that the behavior if PROVISION_TARGET_DISK is undefined or if the interactive case is not used matches our current default behavior. This is that the first valid disk is selected and any removable disks are skipped.
- [X] Tested that PROVISION_TARGET_DISK can be added to an answer file and the disk specified there is selected.
- [X] Tested that using the -i script argument provides a list of valid targets and allows the user to select one.
    - Confirmed that removable disks are displayed as options.
    - Checked that the user prompt does not accept any inputs besides the displayed numbers.
    - provisioning 1
    - provisioning 2

### Procedure

### DIFF
--- a/recipes-core/initrdscripts/files/ni_provisioning
+++ b/recipes-core/initrdscripts/files/ni_provisioning
@@ -24,17 +24,17 @@ early_setup "$@"
 # Source default answer file to prompt user for all questions
 source_default_answer_file
 
-# Source user's answer file, if present
-source_answer_file
-echo
-
-set_target_from_answer_file
-
 if in_recovery_mode; then
 	echo "Automatic provision and reboot in recovery mode"
 	PROVISION_REPARTITION_TARGET="y"
 	PROVISION_REBOOT_METHOD="reboot"
 fi
+
+# Source user's answer file, if present
+source_answer_file
+echo
+
+set_target_from_answer_file
 
 if [[ $restore == "provision-safe" ]]; then
 	if [[ "$PROVISION_REPARTITION_TARGET" != "y" ]]; then
--- a/recipes-core/initrdscripts/files/ni_provisioning
+++ b/recipes-core/initrdscripts/files/ni_provisioning
@@ -24,6 +24,12 @@ early_setup "$@"
 # Source default answer file to prompt user for all questions
 source_default_answer_file
 
+# Source user's answer file, if present
+source_answer_file
+echo
+
+set_target_from_answer_file
+
 if in_recovery_mode; then
 	echo "Automatic provision and reboot in recovery mode"
 	PROVISION_REPARTITION_TARGET="y"
@@ -31,10 +37,6 @@ if in_recovery_mode; then
 fi
 
 if [[ $restore == "provision-safe" ]]; then
-	# source user's answer file, if present
-	source_answer_file
-	echo
-
 	if [[ "$PROVISION_REPARTITION_TARGET" != "y" ]]; then
 		ask_for_continue
 	fi
@@ -51,10 +53,6 @@ elif [[ $restore == "backward-migrate" ]]; then
 	ASK_BEFORE_REBOOT=1
 	PROVISION_REBOOT_METHOD="reboot"
 elif [[ $restore == "provision" ]]; then
-	# source user's answer file, if present
-	source_answer_file
-	echo
-
 	if [[ "$PROVISION_PART_NILRT_ID" = "auto" ]]; then
 		ask_for_continue
 
@@ -108,10 +106,6 @@ elif [[ $restore == "migrate" ]]; then
 elif [[ $restore == "get-image" ]]; then
 	echo "Get Image"
 
-	# source user's answer file, if present
-	source_answer_file
-	echo
-
 	source /sbin/nisystemreplication get
 	if [[ "${FORCE_PROVISIONING}" -eq 1 ]]; then
 		PROVISION_REBOOT_METHOD=${RESTART_OPTION}
@@ -119,10 +113,6 @@ elif [[ $restore == "get-image" ]]; then
 	ASK_BEFORE_REBOOT=1
 elif [[ $restore == "set-image" ]]; then
 	echo "Set Image"
-
-	# source user's answer file, if present
-	source_answer_file
-	echo
 
 	source /sbin/nisystemreplication set
 	if [[ "${FORCE_PROVISIONING}" -eq 1 ]]; then
--- a/recipes-core/initrdscripts/files/ni_provisioning.common
+++ b/recipes-core/initrdscripts/files/ni_provisioning.common
@@ -70,6 +70,16 @@ provision_successfull()
 	echo -e ${GREEN}"PROVISIONING SUCCESSFUL!"${NC} 1>&3 2>&4
 }
 
+set_target_from_answer_file()
+{
+	if [[ -n "$PROVISION_TARGET_DISK" ]]; then
+		TARGET_DISK="$PROVISION_TARGET_DISK"
+		if [[ "${TARGET_DISK: -1}" =~ [0-9] ]]; then
+			PART_SEPARATOR='p'
+		fi
+	fi
+}
+
 umount_partition()
 {
 	if [ ! -z "`mount | grep $1`" ]; then
--- a/recipes-core/initrdscripts/files/ni_provisioning.common
+++ b/recipes-core/initrdscripts/files/ni_provisioning.common
@@ -24,6 +24,7 @@ ARCH=$(uname -m)
 VERBOSE_ARGS=${VERBOSE_ARGS:-"-q"}
 MKFS_ARGS=${MKFS_ARGS:-"$VERBOSE_ARGS -F -I 256"}
 verbose_mode=${verbose_mode:-0}
+interactive_disk_selection=${interactive_disk_selection:-0}
 
 if [ "$verbose_mode" -eq 0 ]; then
 	do_silent() { "$@" &>/dev/null; };
@@ -74,6 +75,7 @@ set_target_from_answer_file()
 {
 	if [[ -n "$PROVISION_TARGET_DISK" ]]; then
 		TARGET_DISK="$PROVISION_TARGET_DISK"
+		PART_SEPARATOR=''
 		if [[ "${TARGET_DISK: -1}" =~ [0-9] ]]; then
 			PART_SEPARATOR='p'
 		fi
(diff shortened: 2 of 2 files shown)


recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -263,6 +274,49 @@ elif [ "$ARCH" = "x86_64" ]; then
...
+
+		# Build a regex pattern for the available devices
+		max_index=$((${#block_devices[@]} - 1))
+		regex="^("
+		for i in $(seq 0 "$max_index"); do
+			regex+="$i|"
+		done
+		regex="${regex%|})$"
+
+		prompt_user "Select a block device to provision" "$regex" "0" block_index
```
chaitu236, line 318: Can you post a screenshot of this step so we can better understand how the devices show up.

recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -263,6 +274,49 @@ elif [ "$ARCH" = "x86_64" ]; then
...
+		for path in /sys/block/*; do
+			path_abbrev=${path##/sys/block/}
+			local block=$(udevadm info -r --query=name --path=$path)
+			check_block_device $path $block $mmc_ok || continue
+
+			# Add each valid block device to block_devices
+			block_devices+=( "$block" )
+		done
+
+		for i in "${!block_devices[@]}"; do
```
chaitu236, line 302: Don't we have to do anything to exclude the provisioning USB drive?

erickshepherdNI: The check_block_devices function should filter out block devices that are no valid provisioning targets. I've done some testing and when a USB drive and SD card are plugged in, only the internal drive gets listed. If I alter the code to allow removable block devices then the provisioning USB was still ignored due to the NIRECOVERY label.

chaitu236: > I've done some testing and when a USB drive and SD card are plugged in, only the internal drive gets listed.

So the current code can never select SD card? In that case, the testing mentioned in PR description with SD card present is not valid right?

> If I alter the code to allow removable block devices then the provisioning USB was still ignored due to the NIRECOVERY label.

For Kinabalu, all block-devices will be non-removable correct? Even so, I think allowing non-removable block devices will be useful for other usecases (ATS tests that require a lot of storage e.g., full feed installation) - but if this is something that is time-consuming wrt implementation/validation, it's ok to skip.

recipes-core/initrdscripts/files/ni_provisioning.common
erickshepherdNI: I added screenshots of the user prompt to the testing section.

recipes-core/initrdscripts/files/ni_provisioning.common
erickshepherdNI: Specifically, the current code will not allow a removable SD card from being selected and yes, my understanding is that Kinabalu will contain multiple non-removable block devices. This is why I commented out the check for if block devices are removable in order to test a case where there are multiple provisioning targets.

erickshepherdNI: If we want to allow users to provision removable devices I could change check_block_device to have an argument that allows for bypassing that check. That would allow the interactive case to select removable devices while the default case would still ignore them.

recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -263,6 +274,49 @@ elif [ "$ARCH" = "x86_64" ]; then
...
+	{
+		local -n selected_block=$1
+		local block_index="?"
+		local mmc_ok=true
+		local block_devices=()
+
+		for path in /sys/block/*; do
+			path_abbrev=${path##/sys/block/}
+			local block=$(udevadm info -r --query=name --path=$path)
+			check_block_device $path $block $mmc_ok || continue
```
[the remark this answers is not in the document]
erickshepherdNI: Manual disk selection should avoid the need to ignore MMC devices since the whitelist was only added because automatic selection would sometimes choose them over the internal disk. In the interactive case they should be available at all times.

recipes-core/initrdscripts/files/ni_provisioning
```diff
@@ -24,17 +24,19 @@ early_setup "$@"
 # Source default answer file to prompt user for all questions
 source_default_answer_file
 
+# Source user's answer file, if present
+source_answer_file
+echo
+
+set_target_from_answer_file
+
```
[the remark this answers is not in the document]
erickshepherdNI: I think this is ok, currently if the answer file contain PRIVISION_TARGET_DISK and the interactive case is selected the interactive case will go first and then be overwritten by the answer file. The part separator gets updated in the set_target_from_answer_file function so it should remain consistent.

recipes-core/initrdscripts/files/ni_provisioning.common
chaitu236: > If we want to allow users to provision removable devices I could change check_block_device to have an argument that allows for bypassing that check. That would allow the interactive case to select removable devices while the default case would still ignore them.

If it's simple enough to do, let's do it.

erickshepherdNI: I've gone ahead and added this functionality. I confirmed that in interactive mode the SD card appears and in the default mode it gets skipped.

recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -263,6 +274,49 @@ elif [ "$ARCH" = "x86_64" ]; then
...
+			block_name=$(lsblk -dn -o NAME "$dev")
+			block_model=$(lsblk -dn -o MODEL "$dev")
+			block_serial=$(lsblk -dn -o SERIAL "$dev")
+			printf "%s) %s\t%s\t%s\n" "$i" "$block_name" "$block_model" "$block_serial"
+		done
+
+		# Build a regex pattern for the available devices
+		max_index=$((${#block_devices[@]} - 1))
+		regex="^("
+		for i in $(seq 0 "$max_index"); do
```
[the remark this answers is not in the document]
erickshepherdNI: Applied

recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -263,6 +274,49 @@ elif [ "$ARCH" = "x86_64" ]; then
...
+			# Add each valid block device to block_devices
+			block_devices+=( "$block" )
+		done
+
+		for i in "${!block_devices[@]}"; do
+			dev="${block_devices[$i]}"
+			block_name=$(lsblk -dn -o NAME "$dev")
+			block_model=$(lsblk -dn -o MODEL "$dev")
+			block_serial=$(lsblk -dn -o SERIAL "$dev")
+			printf "%s) %s\t%s\t%s\n" "$i" "$block_name" "$block_model" "$block_serial"
```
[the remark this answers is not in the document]
erickshepherdNI: Applied

recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -70,6 +70,16 @@ provision_successfull()
 	echo -e ${GREEN}"PROVISIONING SUCCESSFUL!"${NC} 1>&3 2>&4
 }
 
+set_target_from_answer_file()
+{
+	if [[ -n "$PROVISION_TARGET_DISK" ]]; then
+		TARGET_DISK="$PROVISION_TARGET_DISK"
```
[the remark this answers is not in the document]
erickshepherdNI: Partially applied. I reset the PART_SEPARATOR to blank before setting it again when reading from the answer file.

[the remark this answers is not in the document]
erickshepherdNI: Updated PR description

recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -536,18 +592,22 @@ early_setup()
...
-			if [[ -z $TARGET_DISK ]]; then
-				(( verbose_mode )) && echo "not found" >&2
-				sleep 1
-			else
-				(( verbose_mode )) && echo "found TARGET_DISK=$TARGET_DISK" >&2
-				break
-			fi
-		done
+		if [[ $interactive_disk_selection -eq 1 ]]; then
+			select_target_block_device TARGET_DISK
```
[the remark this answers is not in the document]
erickshepherdNI: Set the variable using ${interactive_disk_selection:-0} at the top of the file

recipes-core/initrdscripts/files/ni_provisioning.common
```diff
@@ -263,6 +274,49 @@ elif [ "$ARCH" = "x86_64" ]; then
...
+		# Build a regex pattern for the available devices
+		max_index=$((${#block_devices[@]} - 1))
+		regex="^("
+		for i in $(seq 0 "$max_index"); do
+			regex+="$i|"
+		done
+		regex="${regex%|})$"
+
+		prompt_user "Select a block device to provision" "$regex" "0" block_index
+
```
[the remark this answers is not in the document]
erickshepherdNI: Added a check to return if no block devices are found. The calling function is already set up to handle this case.

recipes-core/initrdscripts/files/ni_provisioning
```diff
@@ -24,17 +24,19 @@ early_setup "$@"
 # Source default answer file to prompt user for all questions
 source_default_answer_file
 
+# Source user's answer file, if present
+source_answer_file
+echo
+
+set_target_from_answer_file
```
gratian: Can we move this block after the `if in_recovery_mode; then` below? Not sure if the answers file can overwrite `PROVISION_REPARTITION_TARGET` or `PROVISION_REBOOT_METHOD` but that way we're sure the behavior did not change from what we've had before.

Also do we need that empty `echo`? We can leave it, it just looks weird.

gratian (approved):
LGTM, just a small nitpick on the location for sourcing the answer file.

recipes-core/initrdscripts/files/ni_provisioning
erickshepherdNI: Done, I've gone ahead and kept the empty echo for now. It does look strange but I want to preserve the current console output as much as possible.

chaitu236 (approved):

PR #989, opened 2026-04-20
meta-nilrt: make next-BSI buildable on arm

rajendra-desai-ni:
### Summary of Changes

bitbake nilrt-base-system-image on arm arch fails because of the
 new FIT image implementation in the upstream https://github.com/openembedded/openembedded-core/commit/05d0c7342d7638dbe8a9f2fd3d1c709ee87d6579.

In order to make dist-next BSI buildable on arm arch, we need to create a new recipe that uses the new FIT implementation and builds the required FIT image.

The PR also fixes few build issues encountered while building the BSI.

### Justification

[AB#3711980](https://dev.azure.com/ni/DevCentral/_workitems/edit/3711980/)

### Testing

- [x] I have built the core package feed with this PR in place. (`bitbake packagefeed-ni-core`)
- [x]  bitbake nilrt-base-system-image on ARM next
- [x] Tested the changes on NI-cRIO 9147 and NI-cRIO 9068 arm targets by copying the BSI tar file into the target and installing it.

image

image

### Procedure

### Note
Dependent PR: https://github.com/ni/openembedded-core/pull/193

### DIFF
--- a/recipes-core/images/includes/nilrt-proprietary.inc
+++ b/recipes-core/images/includes/nilrt-proprietary.inc
@@ -16,7 +16,6 @@ NI_PROPRIETARY_COMMON_PACKAGES = "\
         ni-software-installation-websvc \
         ni-ssl-webserver-support \
         ni-sysapi \
-        ni-sysapi-sshcli \
         ni-sysapi-remote \
         ni-sysapi-webservice \
         ni-sysmgmt-auth-utils \
@@ -36,6 +35,7 @@ NI_PROPRIETARY_COMMON_PACKAGES = "\
 NI_PROPRIETARY_COMMON_PACKAGES:append:x64 = "\
         ni-sdmon \
         ni-wifibledd \
+        ni-sysapi-sshcli \
 "
 
 NI_PROPRIETARY_RUNMODE_PACKAGES = "\
--- a/recipes-kernel/linux/linux-nilrt-fitimage.bb
+++ b/recipes-kernel/linux/linux-nilrt-fitimage.bb
@@ -1,0 +1,23 @@
+SUMMARY = "The Linux kernel as a FIT image (optionally with initramfs) for NILRT"
+SECTION = "kernel"
+
+# If an initramfs is included in the FIT image more licenses apply.
+# But also the kernel uses more than one license (see Documentation/process/license-rules.rst)
+LICENSE = "GPL-2.0-with-Linux-syscall-note"
+LIC_FILES_CHKSUM = "file://${COREBASE}/meta/files/common-licenses/GPL-2.0-with-Linux-syscall-note;md5=0bad96c422c41c3a94009dcfe1bff992"
+
+inherit linux-kernel-base kernel-fit-image
+
+# Set the version of this recipe to the version of the included kernel
+# (without taking the long way around via PV)
+PKGV = "${@get_kernelversion_file("${STAGING_KERNEL_BUILDDIR}")}"
+
+FIT_DESC = "Kernel fitImage for ${DISTRO_NAME}/${PKGV}/${MACHINE}"
+
+DEPENDS += "u-boot"
+
+do_prepare_bootscript() {
+    # Copy bootscript.txt from u-boot’s deploy dir into our sources
+    cp ${DEPLOY_DIR_IMAGE}/bootscript.txt ${UNPACKDIR}/bootscript.txt
+}
+addtask prepare_bootscript before do_compile after do_configure
--- a/conf/machine/xilinx-zynq.conf
+++ b/conf/machine/xilinx-zynq.conf
@@ -5,8 +5,8 @@
 TARGET_ARCH = "arm"
 DEFAULTTUNE = "cortexa9-vfpv3"
 
-KERNEL_CLASSES += "kernel-fitimage"
-KERNEL_IMAGETYPE = "fitImage"
+KERNEL_CLASSES += "kernel-fit-extra-artifacts"
+KERNEL_IMAGETYPE = "zImage"
 
 KERNEL_FEATURES:remove = "cfg/fs/vfat.scc"
 
@@ -18,11 +18,13 @@ MACHINE_FEATURES = "apm ext2 vfat ethernet screen usbgadget usbhost wifi"
 SERIAL_CONSOLES = "115200;ttyS0"
 
 INITRAMFS_FSTYPES = "cpio.gz.u-boot"
+FIT_SUPPORTED_INITRAMFS_FSTYPES = "cpio.gz.u-boot"
 
 # To generate bootscript section in FIT image, u-boot needs to be built.
 UBOOT_ARCH = "arm"
 UBOOT_LOADADDRESS ?= "0x8000"
 UBOOT_ENTRYPOINT ?= "${UBOOT_LOADADDRESS}"
+FIT_UBOOT_ENV = "bootscript.txt"
 UBOOT_ENV = "bootscript"
 UBOOT_ENV_BINARY ?= "bootscript.txt"
 # Set to "" to not generate unnecessary /etc/${UBOOT_INITIAL_ENV}* files.
--- a/recipes-connectivity/ti-wifi-utils/files/0001-fix-efuse_param_type_enm.patch
+++ b/recipes-connectivity/ti-wifi-utils/files/0001-fix-efuse_param_type_enm.patch
@@ -2,6 +2,7 @@ From c4d382a73a1c96c4c1568d4b33363d3ce91c42ba Mon Sep 17 00:00:00 2001
 From: Can Wong <can.wong@emerson.com>
 Date: Wed, 22 Oct 2025 14:55:52 -0500
 Subject: [PATCH] fix efuse_param_type_enm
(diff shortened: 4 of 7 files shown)


recipes-core/images/includes/nilrt-proprietary.inc
```diff
@@ -16,7 +16,6 @@ NI_PROPRIETARY_COMMON_PACKAGES = "\
         ni-software-installation-websvc \
         ni-ssl-webserver-support \
         ni-sysapi \
-        ni-sysapi-sshcli \
```
chaitu236, line 19: Have we communicated to the owners of `ni-sysapi-sshcli` that ARM package is needed? Is there a WI#?

recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -4,3 +4,12 @@ SRC_URI:append:xilinx-zynq = "\
     file://${UBOOT_ENV_BINARY} \
     file://fw_env.config \
 "
+
+do_compile:append:xilinx-zynq() {
+    install -m 0644 ${UNPACKDIR}/bootscript.txt ${B}/bootscript.txt
```
chaitu236, line 9: Why are the `do_compile`, `do_deploy` changes needed? Shouldn't u-boot already have support for handling `bootscript.txt`?

recipes-kernel/linux/linux-nilrt-fitimage.bb
```diff
@@ -0,0 +1,23 @@
...
+# Set the version of this recipe to the version of the included kernel
+# (without taking the long way around via PV)
+PKGV = "${@get_kernelversion_file("${STAGING_KERNEL_BUILDDIR}")}"
+
+FIT_DESC = "Kernel fitImage for ${DISTRO_NAME}/${PKGV}/${MACHINE}"
+
+DEPENDS += "u-boot"
+
+do_prepare_bootscript() {
+    # Copy bootscript.txt from u-boot’s deploy dir into our sources
```
chaitu236, line 20: Why do we need to handle bootscript.txt manually? Shouldn't OE already have support for handling it?

recipes-core/images/includes/nilrt-proprietary.inc
rajendra-desai-ni: I am not sure if we have communicated to the owners of ni-sysapi-sshcli about it. But I have created a work item for that.
https://dev.azure.com/ni/DevCentral/_workitems/edit/3839430

recipes-kernel/linux/linux-nilrt-fitimage.bb
rajendra-desai-ni: The recipe linux-nilrt-fitimage requires bootscript.txt file to be present in its WORKDIR while building the fitimage without which we get this error:

image

recipes-bsp/u-boot/u-boot_%.bbappend
rajendra-desai-ni: Though the u-boot upstream recipe includes bootscript.txt files staged in its WORKDIR, [this](https://github.com/openembedded/openembedded-core/commit/62ff94ff40e823065178318133d54e44a3d8a46d) change in the upstream logic now expects the file to be copied from build directory instead. Due to this we get below error. Hence it is required to copy the file into build directory to resolve this error.

`| install: cannot stat '/../../nilrt/build/tmp-glibc/work/xilinx-zynq-nilrt-linux-gnueabi/u-boot/2026.01/build/bootscript.txt': No such file or directory`

I have removed `do_deploy` as it is already taken care by u-boot build but `do_compile` might still be needed.

chaitu236: https://docs.yoctoproject.org/dev/ref-manual/variables.html#term-UBOOT_ENV says 

> If the variable [UBOOT_ENV_SUFFIX](https://docs.yoctoproject.org/dev/ref-manual/variables.html#term-UBOOT_ENV_SUFFIX) is set to scr the script is packaged as a uImage (mkimage -T script..) otherwise it gets installed verbatim.

In our case, since we're not setting `UBOOT_ENV_SUFFIX`, it defaults to `txt` in which case according to this documentation, `bootscript.txt` should be installed verbatim.

It appears that the [upstream change you mentioned](https://github.com/openembedded/openembedded-core/commit/62ff94ff40e823065178318133d54e44a3d8a46d) only handled the `scr` part but broke the `txt` functionality; there should be something along the lines of

```
if [ -n "${UBOOT_ENV}" ] && [ "${UBOOT_ENV_SUFFIX}" = "scr" ]
then
    if [ "${UBOOT_ENV_SUFFIX}" = "scr" ]
    then
        ${UBOOT_MKIMAGE} -C none -A ${UBOOT_ARCH} -T script -d ${UNPACKDIR}/${UBOOT_ENV_SRC} ${B}/${UBOOT_ENV_BINARY}
    elif [ "${UBOOT_ENV_SUFFIX}" = "txt" ]
    then
        install -m 0644 ${UNPACKDIR}/${UBOOT_ENV_BINARY} ${B}/${UBOOT_ENV_BINARY}
    fi
fi
```

Can you fix this directly in oe-core repo and (after we merge the change in our oe-core fork) try to upstream the fix? Let's avoid carrying these changes in meta-nilrt when it's an issue with upstream code and not something NI specific.

recipes-core/images/includes/nilrt-proprietary.inc
chaitu236: Please assign the bug to sysapi team and let them know.

recipes-kernel/linux/linux-nilrt-fitimage.bb
chaitu236: Adding a new task to copy bootscript.txt doesn't seem like the correct fix. We need to figure out why this failure occurs and find the correct fix.

recipes-core/images/includes/nilrt-proprietary.inc
rajendra-desai-ni: Done

recipes-kernel/linux/linux-nilrt-fitimage.bb
rajendra-desai-ni: FIT_UBOOT_ENV (bootscript.txt) is passed to [this](https://github.com/ni/openembedded-core/blob/094b4b6c7f57c382ef63e325dde22ab8d655d985/meta/classes-recipe/kernel-fit-image.bbclass#L113) function and the [function definition](https://github.com/ni/openembedded-core/blob/094b4b6c7f57c382ef63e325dde22ab8d655d985/meta/lib/oe/fitimage.py#L337) takes the same FIT_UBOOT_ENV's value (bootscript.txt) as both bootscr_id and bootscr_path. The issue occurs when the [called function](https://github.com/ni/openembedded-core/blob/094b4b6c7f57c382ef63e325dde22ab8d655d985/meta/lib/oe/fitimage.py#L337) expects the file to be present in linux-nilrt-fitimage's WORKDIR but since this is a part of u-boot, it is only present in DEPLOYDIR at this time.

One way to fix it is to have a separate variable for bootscr_path to include bootscript.txt by pointing it to the correct path where it exists. This way, we won't have to create a task to copy the file. But this would again introduce some changes to the upstream code to handle the behavior. As we are just using bootscript.txt verbatim, I felt there is not harm copying it from DEPLOYDIR to UNPACKDIR.

Please let me know if there is any other better way than this.

recipes-bsp/u-boot/u-boot_%.bbappend
rajendra-desai-ni: Fixed it here:
https://github.com/ni/openembedded-core/pull/193/commits/4a7a056a4998e889aaef31b305f9800291accd22

recipes-kernel/linux/linux-nilrt-fitimage.bb
chaitu236: > The issue occurs when the called function expects the file to be present in linux-nilrt-fitimage's WORKDIR but since this is a part of u-boot, it is only present in DEPLOYDIR at this time.

Are we sure there is no way to make use of u-boot generated `bootscript.*` in fitimage recipe without adding this appends? I'd be very surprised if there wasn't a way - it would imply no one upstream/other-downstreams is building fitImages with bootscript or it would imply we've misunderstood on how the new framework is supposed to work. We would also need to make a fitImage for safemode in future so I'd like us to make sure we understand how the new framework is supposed to work.

Did we check how existing fitImage recipes in poky/OE build fitImages? Also check [variables glossary](https://docs.yoctoproject.org/dev/ref-manual/variables.html) and mailing list discussions if they offer any insights.

chaitu236 (approved):
I'm giving this a shipit for now. If we hear from upstream on best way to do this, we can put up another PR.

recipes-kernel/linux/linux-nilrt-fitimage.bb
rajendra-desai-ni: https://lists.openembedded.org/g/openembedded-core/topic/best_practice_for_including/119094498
Marking this as resolved since we haven't got any other workarounds other than the existing one.

PR #990, opened 2026-04-21
 Add env_config support for containers

AshwinUjjwal:
### Summary of Changes
Create wrapper scripts for fw_printenv and fw_setenv that read from grubenv instead of hardware partitions. Each container automatically gets a unique serial number based on hostname at startup.

Changes:
Add fw_printenv.wrapper and fw_setenv.wrapper scripts
Create env_config pointing to /boot/grub/grubenv
Create env-config-container.bb recipe
Update init script to generate unique grubenv per container

### Justification
[AB#3201993](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3201993)

### Testing
* [x] bitbake nilrt-slim-container
* [x] bitbake nilrt-runmode-container
On both containers:
* [x] Verified fw_printenv shows all env_config variables: serial#, TargetClass, DeviceCode, DeviceDesc
* [x] Each container received a unique serial number based on hostname


recipes-core/images/files/container/init
```diff
@@ -2,6 +2,37 @@
...
 # then executes the container's CMD..
 
+# Generate a unique grubenv for this container instance.
+# The serial number is derived from the container hostname, which is unique per container.
+GRUBENV_DIR="/etc/container"
+GRUBENV_FILE="$GRUBENV_DIR/grubenv"
+
+mkdir -p "$GRUBENV_DIR" || { echo "Failed to create $GRUBENV_DIR"; exit 1; }
+
+if [ ! -f "$GRUBENV_FILE" ] || ! grep -q "^serial#=" "$GRUBENV_FILE" 2>/dev/null; then
```
rajendra-desai-ni: You have defined the same grubenv variables that are already defined [here](https://github.com/ni/meta-nilrt/blob/nilrt/master/next/recipes-core/images/files/container/grubenv) and initialized to same values as before and again checking for those values and initializing them, which all doesn't make sense. If our main intent here is to update `serial#` with unique container ID which is also the hostname, then it is better to update just that. Other variables like TargetClass, DeviceCode, DeviceDesc need not to be verified since they are not going to change. You can update the logic accordingly.

recipes-ni/fw-printenv/files/env_config
```diff
@@ -0,0 +1 @@
+/etc/container/grubenv
```
rajendra-desai-ni: Missing whiteline.

recipes-ni/fw-printenv/files/fw_printenv
```diff
@@ -69,9 +85,9 @@ fi
 
 show_variable_smbios(){
 	# Return the value that was requested only on NI targets
-	if is_ni_bios || is_ni_device; then
+	if is_ni_bios 2>/dev/null || is_ni_device 2>/dev/null; then
```
rajendra-desai-ni: What is the need for masking error messages?

recipes-ni/fw-printenv/files/fw_printenv
```diff
@@ -69,9 +85,9 @@ fi
...
 		if [ "${smbios_funcs[$1]}" ]; then
-			value=`${smbios_funcs[$1]}`
+			value=$(${smbios_funcs[$1]} 2>/dev/null)
```
rajendra-desai-ni: What is the need for suppressing error messages? This is not a valid change.

recipes-ni/fw-printenv/files/fw_printenv
```diff
@@ -128,7 +152,7 @@ add_rem_vars(){
 	value=""
 
 	#SMBIOS - are read only variables
-	if is_ni_bios || is_ni_device; then
+	if is_ni_bios 2>/dev/null || is_ni_device 2>/dev/null; then
```
rajendra-desai-ni: What is the need for suppressing error messages? This is not a valid change.

recipes-ni/fw-printenv/files/fw_printenv
```diff
@@ -44,6 +46,20 @@ show_variable_grub(){
 	return 1
 }
 
+show_variable_config(){
```
rajendra-desai-ni: You are updating this file which is also included with the BSI and the changes made in this file might affect BSI's fw_printenv functionality as well. Isn't it possible to restrict the changes only to containers?

recipes-ni/fw-printenv/files/fw_printenv
```diff
@@ -151,6 +175,18 @@ add_rem_vars(){
 		echo "Can't overwrite" "\"$var_name\"" >&2
 		exit 1
 	fi
+	if [ -f "$ENV_CONFIG_FILE" ]; then
```
rajendra-desai-ni: Try to move all the changes to container specific scripts.

recipes-ni/fw-printenv/fw-printenv.bb
```diff
@@ -27,14 +30,20 @@ S = "${UNPACKDIR}"
 do_install () {
 	install -d ${D}${base_sbindir}
 	install -d ${D}${datadir}/fw_printenv
+	install -d ${D}${sysconfdir}
```
rajendra-desai-ni: Changes made here would also affect the BSI images. It is better to have a .bbappend recipe under containers and override fw_printenv functionalities there and keep them container specific. This way we can keep original recipe free from unneeded changes.

rajendra-desai-ni (changes requested):

recipes-core/images/files/container/init
AshwinUjjwal: Updated the init script logic to only set serial# with the unique container hostname. Removed unnecessary verification and reinitialization of static variables

recipes-ni/fw-printenv/files/fw_printenv
AshwinUjjwal: The original fw-printenv recipe remains unchanged. Now the env-config-container recipe (which installs wrapper scripts) is only added to container images (nilrt-slim-container and nilrt-runmode-container) via their IMAGE_INSTALL.BSI does not include this recipe, so it continues using the original fw-printenv without any modifications. The softlinks are created only at container startup via the container-specific init script , so BSI fw_printenv functionality is completely unaffected

recipes-ni/fw-printenv/fw-printenv.bb
AshwinUjjwal: Now created a separate env-config-container.bb recipe that's only included in container images. The original fw-printenv recipe remains completely unchanged. The wrapper scripts and container-specific logic are only deployed to containers via this dedicated recipe, ensuring BSI images are unaffected and use the standard fw_printenv functionality.

recipes-ni/fw-printenv/files/env_config
AshwinUjjwal: Now we are not adding this file

recipes-ni/fw-printenv/files/fw_printenv
AshwinUjjwal: All container-specific changes have been moved to a separate recipes-ni/env-config-container/ directory

recipes-ni/fw-printenv/files/fw_printenv
AshwinUjjwal: Now not modifying the original fw_printenv recipe, so error suppression changes have been reverted. The wrapper scripts and container-specific logic in the separate env-config-container recipe handle grubenv access directly, eliminating the need for any error suppression in the base fw_printenv.

recipes-ni/fw-printenv/files/fw_printenv
AshwinUjjwal: We are not modifying the original fw_printenv recipe, so error suppression changes have been reverted. The wrapper scripts and container-specific logic in the separate env-config-container recipe handle grubenv access directly, eliminating the need for any error suppression in the base fw_printenv.

recipes-ni/fw-printenv/files/fw_printenv
AshwinUjjwal: We are not modifying the original fw_printenv recipe, so error suppression changes have been reverted. The wrapper scripts and container-specific logic in the separate env-config-container recipe handle grubenv access directly, eliminating the need for any error suppression in the base fw_printenv.

recipes-ni/env-config-container/env-config-container.bb
```diff
@@ -0,0 +1,26 @@
+SUMMARY = "Container-specific wrappers and env_config for fw_printenv"
+DESCRIPTION = "Wrapper scripts and env_config file for containers"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+SECTION = "base"
+
+COMPATIBLE_MACHINE = "x64"
+PACKAGE_ARCH = "x64"
```
rajendra-desai-ni: Why are these 2 variables needed when they are not used anywhere?

recipes-ni/env-config-container/env-config-container.bb
```diff
@@ -0,0 +1,26 @@
...
+
+RDEPENDS:${PN} = "bash"
```
rajendra-desai-ni, line 8: Why do we need to RDEPEND on bash when `env-config-container` recipe is just installing the wrappers in the desired location and nothing else?

recipes-ni/env-config-container/files/fw_printenv.wrapper
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# Wrapper for fw_printenv that reads from grubenv
+GRUBENV="/boot/grub/grubenv"
+
+# Parse arguments
+SUPPRESS_NAME=0
+if [[ "$1" == "-n" ]]; then
```
rajendra-desai-ni, line 7: What is the use of this condition? `fw_printenv -n` is an invalid option right? Why do we need to suppress this specific option?

recipes-core/images/files/container/init
```diff
@@ -2,6 +2,43 @@
 # This is a custom init file for the NILRT container, that does some fixups and
 # then executes the container's CMD..
 
+# Update serial# in grubenv with unique container hostname
+GRUBENV_FILE="/boot/grub/grubenv"
+SERIAL=$(hostname)
+
+# Convert binary grubenv to text format for modification
```
rajendra-desai-ni: As we want to overwrite `serial#` with `hostname` value, can't we just use `fw_setenv`?
Something like this:
`fw_setenv serial# $(hostname)`
Instead of converting grubenv to txt and back?

recipes-core/images/files/container/init
```diff
@@ -2,6 +2,43 @@
...
+    while IFS= read -r line; do
+        if [[ ! -z "$line" ]]; then
+            grub-editenv "$GRUBENV_FILE" set "$line"
+        fi
+    done < /tmp/grubenv.txt
+    rm -f /tmp/grubenv.txt
+fi
+
+# Create env.config as softlink to grubenv for reference
+ln -sf "$GRUBENV_FILE" /etc/env.config || { echo "Failed to create softlink /etc/env.config"; exit 1; }
```
rajendra-desai-ni: What is the use of `env.config` here? Why do we need this line?

recipes-ni/env-config-container/files/fw_printenv.wrapper
```diff
@@ -0,0 +1,27 @@
...
+
+if [[ -z "$VAR_NAME" ]]; then
+    # List all variables
+    grub-editenv "$GRUBENV" list
+else
+    # Get specific variable
+    VALUE=$(grub-editenv "$GRUBENV" list | grep "^${VAR_NAME}=" | cut -d= -f2-)
+    if [[ -n "$VALUE" ]]; then
+        if [[ $SUPPRESS_NAME -eq 1 ]]; then
+            echo "$VALUE"
```
rajendra-desai-ni, line 22: Instead of suppressing the options and printing nothing, you could echo the standard error message:
`## Error: $VAR_NAME not defined`

recipes-ni/env-config-container/files/fw_setenv.wrapper
```diff
@@ -0,0 +1,38 @@
...
+        else
+            # Update or add variable
+            grub-editenv "$GRUBENV" set "$VAR_NAME=$VAR_VALUE"
+        fi
+    done < "$INPUT_FILE"
+else
+    VAR_NAME="$1"
+    VAR_VALUE="$2"
+    
+    if [[ -z "$VAR_VALUE" ]]; then
```
rajendra-desai-ni, line 41: What about `if [[ -z "$VAR_NAME" ]];` and `if [[ -z "$VAR_VALUE" && -z "$VAR_NAME" ]];` conditions?

recipes-ni/env-config-container/files/fw_setenv.wrapper
```diff
@@ -0,0 +1,38 @@
...
+    fi
+    
+    while IFS= read -r line; do
+        [[ "$line" =~ ^# ]] && continue
+        [[ -z "$line" ]] && continue
+        
+        VAR_NAME=$(echo "$line" | awk '{print $1}')
+        VAR_VALUE=$(echo "$line" | cut -d' ' -f2-)
+        
+        if [[ -z "$VAR_VALUE" ]]; then
```
rajendra-desai-ni, line 24: What about `if [[ -z "$VAR_NAME" ]];` and `if [[ -z "$VAR_VALUE" && -z "$VAR_NAME" ]];` conditions?

rajendra-desai-ni (changes requested):

recipes-core/images/files/container/init
AshwinUjjwal: Its purpose is to provide a reference point at a standard location (/etc/env.config)  that might expect to read environment configuration from there. However, if there are no use of /etc/env.config, then this line is unnecessary and adds no value. I will remove it .

recipes-ni/env-config-container/files/fw_printenv.wrapper
AshwinUjjwal: It is used to suppresses the variable name and outputs only the value.
fw_printenv serial# → outputs: serial#=e693b5cb5be0
fw_printenv -n serial# → outputs: e693b5cb5be0  like this it will give value only

recipes-ni/env-config-container/env-config-container.bb
AshwinUjjwal: The wrapper scripts require bash at runtime since they use bash-specific syntax (#!/bin/bash,). BitBake's QA file-rdeps checker automatically detects that the scripts need bash and fails the build if RDEPENDS is not declared. The RDEPENDS declaration ensures bash is available in the container when these wrapper scripts are executed.

recipes-ni/env-config-container/env-config-container.bb
AshwinUjjwal: I removed those two variables

recipes-ni/env-config-container/files/fw_setenv.wrapper
AshwinUjjwal: I  kept these validation conditions because they prevent errors if someone calls fw_setenv without providing both the variable name and value. They're safety checks to ensure the script fails gracefully with an error message instead of passing empty values to grub-editenv

recipes-ni/env-config-container/files/fw_setenv.wrapper
AshwinUjjwal: I added this validation conditions

recipes-ni/env-config-container/files/fw_printenv.wrapper
AshwinUjjwal: Yes, I have added explicit error messages echo "Error: Variable name not defined" to stderr instead of silently failing. This provides better feedback to users when they forget to provide required argument

recipes-core/images/files/container/init
AshwinUjjwal: Yeah I removed those converting of grubenv and added this fw_setenv serial# $(hostname) , now it is directly updating the serial with hostname

rajendra-desai-ni (approved):

chaitu236 (approved):
This probably could be architected better by creating a recipe that `PROVIDES` `fw_printenv` and install it instead of the default fw_printenv provider on the container.

But that'll probably be non trivial amount of work. So I won't gate the PR.

rajendra-desai-ni:
It is better to combine the 2 commits into a single commit since you are modifying the same files.

PR #997, opened 2026-05-05
nilrt-container: service startup on boot

Shreejit-03:
### Summary of Changes

Enable container services, NI MAX connectivity, and opkg feed support for NILRT runtime containers

Make NILRT run-mode/slim containers usable as RT targets: discoverable in NI MAX, able to accept VeriStand/LabVIEW RT deployments, and able to `opkg install` packages without a reboot — while keeping container-incompatible NI packages out.

- container/init:
  - Bind-mount `boot` so ni-arch-gen can detect it as mounted and generate `/etc/opkg/ni-arch.conf` at first boot.
  - Start D-Bus, Avahi, nirtmdnsd, SystemWebServer and sshd.
  - Fix `/run/natinst` ownership(`webserv:ni`) so SystemWebServer can create PID file.

- nilrt-container.inc:
  - Add `ni-arch-gen` so `/etc/opkg/ni-arch.conf` is generated dynamically at first boot (via `run-postinsts`) instead of shipping a static arch config.
  - Add `ni-sysapi-webservice`, `ni-auth`, `ni-system-webserver`, `ni-webdav-system-webserver-support`, and `nirtmdnsd` to `IMAGE_INSTALL_NODEPS` for NI MAX connectivity, WebDAV deployment, and target discovery.
  - Block container-incompatible packages by injecting dummy opkg status entries : `ni-dim`, `ni-dim-dkms`, `ni-dim-libs`, `ni-dim-sysapi`, `libnidimu1`, `ni-mdbg`, `ni-mdbg-dkms`, `ni-mdbg-errors`. Each is marked installed at version `99.0.0` with an empty `.list` file, so dependent packages' version constraints (e.g. `ni-daqmx-ef` needs `ni-dim >= 26.3.0`) are satisfied without actually installing packages that segfault without real PXI hardware in Docker.
  - Install opkg wrapper at `/usr/local/bin/opkg`.
  - Enable sshd by default in ni-rt.ini.

- env-config-container.bb:
  - Fix `fw_printenv.wrapper` and `fw_setenv.wrapper` permissions from `0550` to `0555` so the `webserv:ni` user can execute them. Without this, `libnitargetcfg` cannot read grubenv variables and NI MAX reports model as "Pele" / "not supported by drivers."

- nilrt-container.postinst:
  - Restore file capabilities lost during OCI image packaging (`setcap` for `hwclock` and `SystemWebServer`).

- opkg-wrapper.sh :
  - Thin wrapper over the real opkg. After a successful `install/upgrade`, restarts `nilvrt` so a newly-installed RT startup app (e.g. the VeriStand engine) is actually loaded — containers aren't rebooted and a running `lvrt` doesn't hot-reload the new app.
  - Guarded: restarts only for RT-relevant packages (`ni-veristand-engine*`, `ni-labview-realtime*`) and only when `lvrt` is idle (none of ports 2040/2050/3363 listening in tcp, so unrelated installs (e.g. `trace-cmd`) and actively-running deployments are never interrupted.

### Justification

[AB#3201988](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3201988).

### Testing

- Built `nilrt-runmode-container` and `nilrt-slim-container` images with bitbake successfully.
- Ran containers with `docker run --privileged --network=nilrt-net` and verified:
  -   SystemWebServer starts and runs as webserv user.
  - SSH is accessible from remote hosts on the macvlan network.
  - /nisysdetails/system API reports "model": "NI LinuxRT Container".
  - NI MAX discovers and connects to the container target.

- [x] ran a LabVIEW RT .vi on the containers.
- [x] ran VeriStand example on the containers.

### DIFF
--- a/recipes-core/images/files/container/init
+++ b/recipes-core/images/files/container/init
@@ -13,13 +13,11 @@ fi
 
 fw_setenv serial# $(hostname)
 
-setcap CAP_SYS_TIME+ep /sbin/hwclock.util-linux
-
 /etc/init.d/populateconfig start
+/etc/init.d/niauth start
 
-if [ -f /etc/init.d/niauth ]; then
-    /etc/init.d/niauth start
-fi
+# Bind-mount /boot so ni-arch-gen can detect it as mounted
+mount --bind /boot /boot
 
 /etc/init.d/run-postinsts start
 
@@ -27,6 +25,19 @@ if [ -x /postinst ]; then
     /postinst
     rm -f /postinst
 fi
+
+# Start D-Bus (required by Avahi)
+/etc/init.d/dbus-1 start
+
+/etc/init.d/avahi-daemon start
+/etc/init.d/nirtmdnsd start
+
+# Fix /run/natinst ownership so webserv can create PID file
+mkdir -p /run/natinst
+chown webserv:ni /run/natinst
+/etc/init.d/systemWebServer start
+
+/etc/init.d/sshd start
 
 # Execute the CMD passed to the container
 exec "$@"
--- a/recipes-core/images/files/container/nilrt-container.postinst
+++ b/recipes-core/images/files/container/nilrt-container.postinst
@@ -9,3 +9,16 @@
 # Update the package feed index and upgrade signing keys so users can
 # opkg install packages immediately.
 opkg update && opkg upgrade opkg-keyrings || true
+
+# Restore file capabilities lost during OCI image packaging (tar layers
+# do not preserve xattrs). These persist on the container filesystem
+# so they only need to be applied once.
+if command -v setcap >/dev/null 2>&1; then
+    if [ -x /sbin/hwclock.util-linux ]; then
+        setcap cap_sys_time+ep /sbin/hwclock.util-linux || true
+    fi
+    if [ -x /usr/local/natinst/share/NIWebServer/SystemWebServer ]; then
+        setcap cap_net_bind_service,cap_net_admin,cap_ipc_lock,cap_sys_rawio,cap_sys_boot,cap_sys_nice,cap_sys_time,cap_syslog+ep \
+            /usr/local/natinst/share/NIWebServer/SystemWebServer || true
+    fi
+fi
--- a/recipes-core/images/files/container/opkg-wrapper.sh
+++ b/recipes-core/images/files/container/opkg-wrapper.sh
@@ -1,0 +1,123 @@
+#!/bin/sh
+# opkg wrapper for NILRT containers.
+#
+# LabVIEW/VeriStand expect the target to reboot after installation so the
+# LabVIEW RT engine (lvrt) re-reads its startup application (e.g. the
+# VeriStand Engine rtexe). Containers are not rebooted, and an already-running
+# lvrt does not pick up a newly-installed startup app on its own.
+#
+# This wrapper runs the real opkg and, after a successful install/upgrade,
+# restarts nilvrt -- but ONLY when:
+#   1. an RT-relevant package was installed/upgraded (ni-veristand-engine*,
+#      ni-labview-realtime*), AND
+#   2. lvrt is currently idle, i.e. not serving an application.
+#
+# The idle check protects running work: installing an unrelated package
+# (trace-cmd, etc.) never restarts lvrt, and even an RT-relevant upgrade is
+# skipped while lvrt is actively running a VI/engine, so deployed
+# applications are not torn down underneath the user. lvrt opens its
+# application-level ports (VeriStand engine 2040/2050, VI Server 3363) only
+# while an application is loaded; when idle, none of them listen.
+#
+# Installed to /usr/local/bin/opkg, which precedes /usr/bin/opkg on PATH.
+
+REAL_OPKG=/usr/bin/opkg
+
+# lvrt_busy: return 0 (busy) if lvrt is currently serving an application, i.e.
+# any application-level port is in the LISTEN state (st == 0A in /proc/net/tcp):
+#   2040 -> 0x07F8 (VeriStand engine)
+#   2050 -> 0x0802 (VeriStand engine)
+#   3363 -> 0x0D23 (LabVIEW VI Server)
+# Returns 1 (idle) if none of them are listening.
+lvrt_busy() {
+    for f in /proc/net/tcp /proc/net/tcp6; do
+        [ -r "$f" ] || continue
+        if awk 'NR > 1 {
+                    split($2, a, ":")
+                    if ($4 == "0A" && (a[2] == "07F8" || a[2] == "0802" || a[2] == "0D23"))
+                        found = 1
+                }
+                END { exit found ? 0 : 1 }' "$f"; then
+            return 0
+        fi
+    done
+    return 1
+}
+
+# is_rt_pkg: return 0 if the given package name is one whose installation
+# affects the lvrt startup application and therefore warrants a restart.
+is_rt_pkg() {
+    case "$1" in
+        ni-veristand-engine*|ni-labview-realtime*) return 0 ;;
+        *) return 1 ;;
+    esac
+}
+
+"$REAL_OPKG" "$@"
+rc=$?
+
+# Parse the opkg command line: find the subcommand (skipping global options and
+# their arguments) and collect the package-name operands that follow it.
+subcommand=""
+pkgs=""
+expect_option_arg=0
+for arg in "$@"; do
+    if [ -n "$subcommand" ]; then
+        case "$arg" in
+            -*) ;;            # ignore per-command options
+            *) pkgs="$pkgs $arg" ;;
+        esac
+        continue
+    fi
+    if [ "$expect_option_arg" -eq 1 ]; then
+        expect_option_arg=0
+        continue
+    fi
+    case "$arg" in
+        -f|--conf|-o|--offline-root|-t|--tmp-dir|-l|--lists-dir|--cache|--add-arch|--add-dest)
+            expect_option_arg=1
+            ;;
+        --conf=*|--offline-root=*|--tmp-dir=*|--lists-dir=*|--cache=*|--add-arch=*|--add-dest=*)
+            ;;
+        -*)
+            ;;
+        *)
+            subcommand="$arg"
+            ;;
+    esac
+done
+
+# Decide whether an RT-relevant package was involved.
+rt_relevant=0
+case "$subcommand" in
+    install)
+        for p in $pkgs; do
+            if is_rt_pkg "$p"; then rt_relevant=1; break; fi
+        done
+        ;;
+    upgrade)
+        if [ -z "$pkgs" ]; then
+            # "opkg upgrade" with no operands upgrades everything, which may
+            # include an RT package; rely on the idle gate below.
+            rt_relevant=1
+        else
+            for p in $pkgs; do
+                if is_rt_pkg "$p"; then rt_relevant=1; break; fi
+            done
+        fi
+        ;;
+esac
+
+# Restart lvrt only on a successful, RT-relevant transaction while lvrt is idle.
+if [ "$rc" -eq 0 ] && [ "$rt_relevant" -eq 1 ] && [ -x /etc/init.d/nilvrt ]; then
+    if lvrt_busy; then
+        echo "opkg: lvrt is serving an application; skipping lvrt restart to avoid interrupting it." >&2
+    else
+        echo "opkg: restarting lvrt to launch newly-installed RT application..." >&2
+        /etc/init.d/nilvrt stop 2>/dev/null
+        sleep 1
+        /etc/init.d/nilvrt start 2>/dev/null
+    fi
+fi
+
+exit $rc
--- a/recipes-core/images/includes/nilrt-container.inc
+++ b/recipes-core/images/includes/nilrt-container.inc
@@ -9,9 +9,24 @@ FORCE_RO_REMOVE = "1"
 # The backing func will test the installed package manifest and only
 # issue an opkg remove if the package name is present. So no wildcards.
 ROOTFS_RO_UNNEEDED = "\
-	ni-arch-gen \
 	kernel-devsrc \
 "
+
+# Ensure ni-arch-gen is available in all container variants to generate
+# /etc/opkg/ni-arch.conf at first boot via run-postinsts.
+IMAGE_INSTALL_NODEPS += "ni-arch-gen"
+
+# ni-sysapi-webservice provides the /nisysapi/server endpoint needed
+# for NI MAX connectivity.
+# ni-auth is required by SystemWebServer (segfaults without niauth_daemon).
+# ni-system-webserver provides the web server for deployment and discovery
+# by NI MAX and VeriStand.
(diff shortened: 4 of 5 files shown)


recipes-core/images/files/container/init
```diff
@@ -12,5 +12,28 @@ fi
 
 /etc/init.d/run-postinsts start
 
+# Add PXI architecture definitions
+cat > /etc/opkg/ni-arch.conf <<'ARCHEOF'
```
chaitu236: Why not just run ni-arch-gen.py or run whichever initscript runs it?

recipes-core/images/files/container/init
```diff
@@ -12,5 +12,28 @@ fi
...
+ARCHEOF
+
+if [ -f /etc/init.d/systemWebServer ]; then
+    /etc/init.d/systemWebServer start
+fi
+
+# Enable and start sshd
+NI_RT_INI="/etc/natinst/share/ni-rt.ini"
+if ! grep -q '^\[systemsettings\]' "$NI_RT_INI" 2>/dev/null; then
+    printf '\n[systemsettings]\n' >> "$NI_RT_INI"
```
chaitu236: Why not make these changes in the container image itself instead of putting them in init?

recipes-core/images/files/container/init
```diff
@@ -12,5 +12,28 @@ fi
...
+fi
+if ! grep -q 'sshd.enabled' "$NI_RT_INI" 2>/dev/null; then
+    sed -i '/^\[systemsettings\]/a sshd.enabled="True"' "$NI_RT_INI"
+fi
+if [ -f /etc/init.d/sshd ]; then
```
chaitu236: Do we really need to check for presence of /etc/init.d/sshd? We always package it in images right?

recipes-core/images/files/container/init
```diff
@@ -12,5 +12,28 @@ fi
...
 /etc/init.d/run-postinsts start
 
+# Add PXI architecture definitions
+cat > /etc/opkg/ni-arch.conf <<'ARCHEOF'
+arch PXI 13
+arch x64-PXI 31
+arch DeviceCode77E1 32
+ARCHEOF
+
+if [ -f /etc/init.d/systemWebServer ]; then
```
chaitu236: Do we really need to check for presence of /etc/init.d/systemWebServer? We always package it in images right?

recipes-core/images/files/container/init
```diff
@@ -12,5 +12,28 @@ fi
 
 /etc/init.d/run-postinsts start
 
+# Add PXI architecture definitions
+cat > /etc/opkg/ni-arch.conf <<'ARCHEOF'
+arch PXI 13
+arch x64-PXI 31
+arch DeviceCode77E1 32
+ARCHEOF
```
chaitu236: EOF is the [standard](https://codesearch.ni.systems/search/text?q=%3C%3CEOF%20repo%3Ascarthgap&fold_case=auto&regex=true&context=true) and if there is no strong reason to change it, let's not change it.

recipes-core/images/files/container/init
Shreejit-03: `nilrt-slim-container` didn't include the `ni-arch-gen` package — it's now added via `IMAGE_INSTALL_NODEPS`.

recipes-core/images/files/container/init
Shreejit-03: The changes have been added to `container_image_pre()` in `nilrt-container.inc`.

recipes-core/images/files/container/init
Shreejit-03: No, it's always present in our images. I have now removed the guard.

recipes-core/images/files/container/init
Shreejit-03: `nilrt-slim-container` didn't include `ni-system-webserver` — it's now added via `IMAGE_INSTALL_NODEPS`, so the guard has been removed.

recipes-core/images/files/container/init
Shreejit-03: Agreed , this is now handled entirely by the `ni-arch-gen` package at first boot.

recipes-core/images/files/container/opkg-wrapper.sh
```diff
@@ -0,0 +1,61 @@
...
+            sleep 2
+        done
+
+        # Re-apply setgid after VeriStand creates its directories under /c
+        if [ -d /c ]; then
+            find /c -type d ! -perm -g+s -exec chmod g+s {} + 2>/dev/null
+        fi
+        ;;
+esac
+
```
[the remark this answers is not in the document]
Shreejit-03: Fixed.

recipes-core/images/includes/nilrt-container.inc
```diff
@@ -54,20 +74,37 @@ fakeroot container_rootfs_post() {
...
 
 	# Workaround for nirtcfg's ni-rt.ini.lock bug
 	# /run/lock cannot have the sticky-bit set.
 	install -m 777 -d ${IMAGE_ROOTFS}/run/lock
+
+	# Enable sshd by default in containers
+	if ! grep -q '^\[systemsettings\]' ${IMAGE_ROOTFS}${sysconfdir}/natinst/share/ni-rt.ini 2>/dev/null; then
+		printf '\n[systemsettings]\n' >> ${IMAGE_ROOTFS}${sysconfdir}/natinst/share/ni-rt.ini
+	fi
+	sed -i '/^\[systemsettings\]/a sshd.enabled="True"' ${IMAGE_ROOTFS}${sysconfdir}/natinst/share/ni-rt.ini
```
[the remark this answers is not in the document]
Shreejit-03: Fixed.

recipes-core/images/includes/nilrt-container.inc
```diff
@@ -54,20 +74,37 @@ fakeroot container_rootfs_post() {
...
 	# remake some removed volatile directories that do_rootfs removes
 	install -m 1777 -d ${IMAGE_ROOTFS}/${localstatedir}/volatile/tmp
 	install -m 755 -d ${IMAGE_ROOTFS}/${localstatedir}/volatile/log
 
 	# Workaround for nirtcfg's ni-rt.ini.lock bug
 	# /run/lock cannot have the sticky-bit set.
 	install -m 777 -d ${IMAGE_ROOTFS}/run/lock
+
+	# Enable sshd by default in containers
+	if ! grep -q '^\[systemsettings\]' ${IMAGE_ROOTFS}${sysconfdir}/natinst/share/ni-rt.ini 2>/dev/null; then
```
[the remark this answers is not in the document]
Shreejit-03: These containers are specifically for LabVIEW/VeriStand development and testing — SSH is a core requirement for all variants (debugging, file transfer, remote access). There is no "production" container variant where SSH would be unwanted. The init script also starts sshd unconditionally, so the ni-rt.ini entry just keeps the NI system configuration consistent (prevents NI MAX from showing SSH as "disabled" while it's actually running).

recipes-core/images/files/container/init
```diff
@@ -13,13 +13,11 @@ fi
...
-setcap CAP_SYS_TIME+ep /sbin/hwclock.util-linux
-
 /etc/init.d/populateconfig start
+/etc/init.d/niauth start
 
-if [ -f /etc/init.d/niauth ]; then
-    /etc/init.d/niauth start
-fi
+# Bind-mount /boot so ni-arch-gen can detect it as mounted
+mount --bind /boot /boot
```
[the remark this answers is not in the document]
Shreejit-03: Both `niauth` and `boot` are guaranteed present in all container variants — they come from `packagegroup-ni-base` (which both `nilrt-runmode-container` and `nilrt-slim-container` inherit). The `boot` directory is created during rootfs assembly and populated with `grubenv` by `container_rootfs_post()`.

recipes-ni/env-config-container/env-config-container.bb
```diff
@@ -18,7 +18,7 @@ do_install () {
 	install -d ${D}${base_sbindir}
 
 	# Install wrapper scripts with .wrapper suffix (softlinks will be created by init)
-	install -m 0550   ${S}/fw_printenv.wrapper ${D}${base_sbindir}/
-	install -m 0550   ${S}/fw_setenv.wrapper   ${D}${base_sbindir}/
+	install -m 0555   ${S}/fw_printenv.wrapper ${D}${base_sbindir}/
+	install -m 0555   ${S}/fw_setenv.wrapper   ${D}${base_sbindir}/
```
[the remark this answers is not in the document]
Shreejit-03: Fixed.

recipes-core/images/files/container/opkg-wrapper.sh
```diff
@@ -0,0 +1,61 @@
...
+/usr/bin/opkg "$@"
+rc=$?
+
+# After install/upgrade, neutralize crashing libraries and fix permissions
+case "$1" in
+    install|upgrade)
+        # Remove libnidimu (ni-dim) - segfaults without real PXI hardware
+        rm -f /usr/lib/x86_64-linux-gnu/libnidimu.so* 2>/dev/null
+        # Remove libnimdbgu (mDNS browsing) - segfaults in Docker networking
+        rm -f /usr/lib/x86_64-linux-gnu/libnimdbgu.so* 2>/dev/null
```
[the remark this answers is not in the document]
Shreejit-03: nilrt containers exclusively target x86_64. There are no ARM or other arch variants for these container images and none are planned as of now.

recipes-core/images/files/container/init
```diff
@@ -28,5 +26,26 @@ if [ -x /postinst ]; then
...
+/etc/init.d/nirtmdnsd start
+
+# Fix /run/natinst ownership so webserv can create PID file
+mkdir -p /run/natinst
+chown webserv:ni /run/natinst
+/etc/init.d/systemWebServer start
+
+# Start NI XNT services (required for NI-RPC interfaces)
+if [ -x /usr/sbin/nixntipstackd ]; then
+    /usr/sbin/nixntipstackd &
```
chaitu236: I see that these are not default installed in BSI - I suppose they're for cases when a customer creates container image that has them installed?

If so, are we intending to start initscripts for every potential h/w a customer may want to install right here?

I thought the only s/w install workflow we'll support on containers is via `opkg install` in custom docker images.. if so, why not let customers edit/append to init in the docker images themselves for each s/w they want to install (we could provide commands for what to add for each s/w)?

It seems to not be sustainable to add workarounds for any future pieces of s/w someone would want to install in the default init (like here and the dim stuff below, etc). But if there's a good reason to do this, I'd like to know.

recipes-ni/env-config-container/env-config-container.bb
```diff
@@ -18,7 +18,12 @@ do_install () {
...
 
 	# Install wrapper scripts with .wrapper suffix (softlinks will be created by init)
-	install -m 0550   ${S}/fw_printenv.wrapper ${D}${base_sbindir}/
-	install -m 0550   ${S}/fw_setenv.wrapper   ${D}${base_sbindir}/
+	# fw_printenv needs group 'ni' (gid 500) execute permission because
+	# SystemWebServer runs as webserv:ni and libnitargetcfg calls
+	# /sbin/fw_printenv to read DeviceCode/DeviceDesc. Without execute
+	# access, NI MAX shows model "Pele".
+	install -m 0550 ${S}/fw_printenv.wrapper ${D}${base_sbindir}/
+	chgrp 500 ${D}${base_sbindir}/fw_printenv.wrapper
```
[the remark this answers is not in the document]
Shreejit-03: I originally used `install -m 0550 -g ni` but it fails at build time with `install: invalid group 'ni'` — the group only exists in the target rootfs, not in pseudo's namespace during do_install. The chgrp 500 approach works because pseudo records the numeric GID and it resolves to `ni` in the final image.

GID 500 is a fixed allocation defined in [group ](https://github.com/ni/meta-nilrt/blob/bf67c4b7e3470f3ebda078adfe02dec3094efbd2/files/group#L5)— it's not fragile since NILRT always uses this value.

recipes-core/images/files/container/nilrt-container.postinst
```diff
@@ -9,3 +9,16 @@
...
+# Restore file capabilities lost during OCI image packaging (tar layers
+# do not preserve xattrs). These persist on the container filesystem
+# so they only need to be applied once.
+if command -v setcap >/dev/null 2>&1; then
+    if [ -x /sbin/hwclock.util-linux ]; then
+        setcap cap_sys_time+ep /sbin/hwclock.util-linux
+    fi
+    if [ -x /usr/local/natinst/share/NIWebServer/SystemWebServer ]; then
+        setcap cap_net_bind_service,cap_net_admin,cap_ipc_lock,cap_sys_rawio,cap_sys_boot,cap_sys_nice,cap_sys_time,cap_syslog+ep \
+            /usr/local/natinst/share/NIWebServer/SystemWebServer
```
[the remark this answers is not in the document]
Shreejit-03: Fixed.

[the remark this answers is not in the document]
Shreejit-03: Fixed.

recipes-core/images/files/container/opkg-wrapper.sh
```diff
@@ -0,0 +1,86 @@
...
+expect_option_arg=0
+for arg in "$@"; do
+    if [ "$expect_option_arg" -eq 1 ]; then
+        expect_option_arg=0
+        continue
+    fi
+    case "$arg" in
+        --)
+            continue
+            ;;
```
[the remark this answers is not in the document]
Shreejit-03: Fixed.

rajendra-desai-ni (approved):

recipes-core/images/files/container/opkg-wrapper.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/sh
```
chaitu236, line 1: @Shreejit-03 We had a discussion offline about exploring other options to prevent undesirable package installation by marking those packages as already installed. What came out of that research?

chaitu236 (changes requested):

recipes-core/images/files/container/opkg-wrapper.sh
Shreejit-03: I explored the "mark them as already installed" approach, and it does work as a blocker for the unwanted NI package families. The key requirement is that the rootfs opkg database must be updated at image-build time, and each injected package entry also needs a matching empty .list file or opkg will complain.

The first attempt to do this from a recipe was not reliable in a fresh image, because the fake installed state did not persist the way we needed. The stable solution was to inject the status entries in the image include during rootfs preprocessing, so the final container image always ships with those packages pre-marked as installed.

The opkg wrapper is still required, but only to restart lvrt so a newly installed RT startup app (e.g. the VeriStand engine) is actually loaded — containers aren't rebooted, and a running lvrt doesn't hot-reload the new app. The restart is guarded: it fires only for RT-relevant packages and only when lvrt is idle, so unrelated installs (e.g. trace-cmd) and active deployments are never interrupted.

chaitu236 (approved):

PR #1026, opened 2026-06-16
root-login: Use local root account instead of NIAuth admin

shubhanshumt26:
### Summary of Changes
In dist-next, `ni-auth` no longer ships its NSS and PAM modules, and BSI only pulls `ni-auth` (not `pam-plugin-niauth` / `libnss-niauth`/`ni-auth-networkcontroller`). The old interactive login was the NIAuth-backed `'admin'` user (empty password), which only worked because those modules were pulled in transitively. With them gone, that login path no longer exists, so make the local `root `account usable out of the box instead.

  * `nilrt-image-base.inc`: add the `'empty-root-password'` and `'allow-empty-password'` image features so root keeps an empty `/etc/shadow` field and pam_unix/sshd accept it (the rootfs postprocess would otherwise zap it and nullok_secure would reject empty passwords over SSH).

  * `nilrt.conf`: drop the NIAuth-era `ROOT_HOME = "/home/admin"` and create new override `ROOT_HOME = /root` to follow systemd convention and latest FHS suggestion ([FYI](https://dev.azure.com/ni/DevCentral/_git/labview/pullRequest/1303859?discussionId=10158552#1782913218)).

  * `niacctbase.bb`: drop the `/home/root -> /home/admin` symlink from useradd_preinst and create new symlink from `/home/admin/` -> `/root` to support backward combability for LabVIEW VIs ([FYI](https://dev.azure.com/ni/DevCentral/_git/labview/pullRequest/1303859?discussionId=10158552#1782772987)).

  * `nilrt-proprietary.inc`: remove `ni-auth` from `NI_PROPRIETARY_COMMON_PACKAGES`; it is pulled in via BSI dependencies and should not be force-installed here.
  * Update all refs of `admin->root` as admin user won't exist in dist-next BSI.
  * `ni-acctsync`: at boot, `insert_root_hash` copies root's password field from the config-partition `.shadow` onto `shadow`. Guard `hash_sanity_check` so a stale locked * (the zap_empty_root_password artifact carried across reflashes on the config partition) is no longer restored — otherwise it re-locks `root `on every boot even though the image ships root blank. Also accept a ` !` -prefixed hash so a deliberately locked password (`passwd -l`) syncs, which the old SHA512 check rejected.

### Justification

[AB#3650235](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3650235)

Posted [Announcement ](https://teams.microsoft.com/l/message/19:0d061f31186241a89957d9b050f28240@thread.tacv2/1782727878691?tenantId=eb06985d-06ca-4a17-81da-629ab99f6505&groupId=e7f1f87e-1a21-4b20-b242-fde2ebf4b8ce&parentMessageId=1782727878691&teamName=RTOS&channelName=Announcements&createdTime=1782727878691)over Teams in multiple channels so that client teams are aware of this major change in Dist-Next.

Note: Don't merge immediately. We have dependency on this [PR](https://dev.azure.com/ni/DevCentral/_git/labview/pullrequest/1313659).

### Testing
* [x] I have built the Safemode image with this PR in place. (`bitbake nilrt-safemode-rootfs`)
* [x] I have built the Base System Image feed with this PR in place. (`bitbake nilrt-base-system-image`)
* [x] I have boot tested Safemode Image & BSI on a `CRIO-9030` target. 
* [x] I have verified that password synchronization works for root user over Safemode & Runmode.
* [x] Verified `snac-configure` is working as expected.
* [x] Updated the [NI-Auth-Removal.md](https://dev.azure.com/ni/DevCentral/_git/AppCentral.wiki?path=/Organizations/Platform-HW-%26-Drivers/Squads/RTOS/Features/NI-Auth-Removal.md&version=GBwikiMaster&_a=preview) with all testing performed.
* [x] I verified a stale `*` in `/etc/natinst/share/.shadow` no longer locks root out on a freshly installed runmode — root stays blank/loginable (previously it re-locked to `root:*`), and the marker self-heals from `.shadow` after a clean reboot.
* [x] I verified a deliberate `passwd -l root` lock persists across reboot and is recoverable from `safemode`.
* [x] Provisioned a fresh VM via `recovery media` (set root password), installed this BSI → password synced to `runmode` and login succeeded.

Note: I have built the images using the custom packages (niauth & other split packages) produced by this [PR](https://dev.azure.com/ni/DevCentral/_git/labview/pullrequest/1313659). Then created a custom feed so they are installed in BSI by default while building.

## Snippets
**Safemode**

image

**Runmode**

image

**Hardware Manager**:
Safemode
image

Runmode
image

Note: In Hardware Manager we can't add a fresh target with root user (Dist-next case) directly we need to add it first from terminal using ssh. [Filed Bug](https://dev.azure.com/ni/DevCentral/_workitems/edit/3930353).

**Password Synchronization**
- Set password in runmode. Reboot. Then login with that password in safemode.
image

- Remove password from safemode. Reboot. Then login in runmode.
image

### Procedure

### DIFF
--- a/recipes-core/images/includes/nilrt-image-base.inc
+++ b/recipes-core/images/includes/nilrt-image-base.inc
@@ -23,6 +23,26 @@ do_image_build_test() {
 	fi
 }
 
+# Guarantee a usable LOCAL root login.
+#
+# Historically the default interactive login was the NIAuth-backed 'admin' user
+# (empty password), which was only visible/usable because NIAuth's NSS and PAM
+# modules were pulled in transitively by ni-auth. As of dist-next, ni-auth no
+# longer installs those modules, so the NIAuth 'admin'/"" login is gone. Make the
+# local root account loginable out of the box (the box ships open and is expected
+# to be secured on first configuration; see FEATURE 3835079):
+#
+#   * empty-root-password  - keep root's /etc/shadow password field empty
+#                            (otherwise the rootfs postprocess zaps it to '*').
+#   * allow-empty-password - rewrite 'pam_unix.so nullok_secure' to plain
+#                            'nullok' in /etc/pam.d/* and set sshd
+#                            PermitEmptyPasswords. nullok_secure only permits
+#                            blank passwords from ttys listed in /etc/securetty,
+#                            which excludes SSH (PAM_TTY=ssh), so without this an
+#                            empty root password could not log in over SSH.
+#
+IMAGE_FEATURES += "empty-root-password allow-empty-password"
+
 python install_additional_feeds() {
     for source_name, uri in d.getVarFlags('PACKAGE_FEED_URIS_ADDITIONAL').items():
         source_conf = os.path.join(d.getVar('IMAGE_ROOTFS'), 'etc', 'opkg', ('%s.conf' % source_name))
--- a/recipes-core/images/includes/nilrt-proprietary.inc
+++ b/recipes-core/images/includes/nilrt-proprietary.inc
@@ -6,7 +6,6 @@ NI_PROPRIETARY_SAFEMODE_PACKAGES = "\
 
 NI_PROPRIETARY_COMMON_PACKAGES = "\
         libnitargetcfg \
-        ni-auth \
         ni-auth-webservice \
         ni-avahi-client \
         ni-ca-certs \
--- a/recipes-kernel/kernel-tests/files/iperf-load
+++ b/recipes-kernel/kernel-tests/files/iperf-load
@@ -1,12 +1,12 @@
 #!/bin/bash
 source "$(dirname "$0")"/common.cfg
-source /home/admin/.iperf.info
+source /root/.iperf.info
 
 case "$1" in
 	check)
 		if [ -z "$IPERF_SERVER" ]; then
 			echo "Warning: iperf server not configured; skipping iperf based network load test."
-			echo "Create/edit /home/admin/.iperf.info file with IPERF_SERVER=<server> and IPERF_PORT=<port> to configure a server to connect to for this test."
+			echo "Create/edit /root/.iperf.info file with IPERF_SERVER=<server> and IPERF_PORT=<port> to configure a server to connect to for this test."
 			echo "SKIP: test_kernel_cyclictest_iperf"
 			exit 77
 		fi
--- a/recipes-kernel/kernel-tests/files/run-docker-cyclictest
+++ b/recipes-kernel/kernel-tests/files/run-docker-cyclictest
@@ -6,7 +6,7 @@ case "$#" in
 		docker run --privileged --network=host \
 		       -v ${LOG_DIR}:${LOG_DIR} \
 		       -v ${TEST_DIR}:${TEST_DIR} \
-		       -v /home/admin:/home/admin \
+		       -v /root:/root \
 		       -v /usr/share/fw_printenv:/usr/share/fw_printenv \
 		       -v /sbin/fw_printenv:/sbin/fw_printenv \
 		       -v /usr/share/nisysinfo:/usr/share/nisysinfo \
@@ -22,7 +22,7 @@ case "$#" in
 		       --cpuset-cpus "$2" \
 		       -v ${LOG_DIR}:${LOG_DIR} \
 		       -v ${TEST_DIR}:${TEST_DIR} \
-		       -v /home/admin:/home/admin \
+		       -v /root:/root \
 		       -v /usr/share/fw_printenv:/usr/share/fw_printenv \
 		       -v /sbin/fw_printenv:/sbin/fw_printenv \
 		       -v /usr/share/nisysinfo:/usr/share/nisysinfo \
--- a/recipes-ni/ni-acctsync/files/ni-acctsync
+++ b/recipes-ni/ni-acctsync/files/ni-acctsync
@@ -36,9 +36,19 @@ check_hash_equality()
 hash_sanity_check()
 {
     SHADOW_HASH="$1"
-    
-    if [ -z "$SHADOW_HASH" ] || [ "$SHADOW_HASH" = "*" ] || [ "$SHADOW_HASH" = "!" ]; then
+
+    # Empty (blank root) or bare '!' (deliberate passwd -l of a passwordless
+    # root) are both valid to restore as-is. '!'-prefixed hashes are handled below.
+    if [ -z "$SHADOW_HASH" ] || [ "$SHADOW_HASH" = "!" ]; then
         return 0
+    fi
+
+    # '*' is the zap_empty_root_password artifact for a never-set root, not a
+    # deliberate lock; skip it so a stale '*' on the config partition can't lock
+    # root out.
+    if [ "$SHADOW_HASH" = "*" ]; then
+        log "Shared root hash is '*' (zap artifact); skipping sync to avoid locking root out."
+        return 1
     fi
     
     if [ -f "/etc/login.defs" ]; then
@@ -55,7 +65,8 @@ hash_sanity_check()
     case "$ENCRYPT_METHOD" in
         SHA512)
             log "Checking sanity of SHA512 hash..."
-            if [[ "$SHADOW_HASH" =~ ^\$6\$[A-Za-z0-9./]{1,16}\$[A-Za-z0-9./]{86}$ ]]; then
+            # Optional leading '!' allows a deliberate 'passwd -l' lock (!$6$...) to sync.
+            if [[ "$SHADOW_HASH" =~ ^!?\$6\$[A-Za-z0-9./]{1,16}\$[A-Za-z0-9./]{86}$ ]]; then
                 log "Sanity check passed. Continue to sync."
                 return 0
             else
--- a/recipes-ni/ni-tracefs-utils/files/traceextract
+++ b/recipes-ni/ni-tracefs-utils/files/traceextract
@@ -4,5 +4,5 @@ set -e
 
 sudo -u lvuser touch "$1"
 chmod 440 "$1"
-chown admin:administrators "$1"
+chown root:root "$1"
 trace-cmd extract -o "$1"
--- a/recipes-ni/niacctbase/niacctbase.bb
+++ b/recipes-ni/niacctbase/niacctbase.bb
@@ -40,7 +40,6 @@ USERADD_PARAM:${PN} = " \
 useradd_preinst:append () {
 	eval ${PSEUDO} chmod g+sw ${SYSROOT}/home/${LVRT_USER} || true
 	eval ${PSEUDO} chmod g+s ${SYSROOT}/home/webserv || true
-	eval ${PSEUDO} ln -sf /home/admin ${SYSROOT}/home/root || true
 }
 
 
--- a/recipes-ni/niacctbase/niacctbase.bb
+++ b/recipes-ni/niacctbase/niacctbase.bb
@@ -39,6 +39,8 @@ USERADD_PARAM:${PN} = " \
 useradd_preinst:append () {
 	eval ${PSEUDO} chmod g+sw ${SYSROOT}/home/${LVRT_USER} || true
 	eval ${PSEUDO} chmod g+s ${SYSROOT}/home/webserv || true
+	# Keep legacy /home/admin working by pointing it at root's home (/root).
+	eval ${PSEUDO} ln -sf /root ${SYSROOT}/home/admin || true
 }
 
 
--- a/recipes-ni/niacctbase/niacctbase.bb
+++ b/recipes-ni/niacctbase/niacctbase.bb
@@ -8,7 +8,6 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 PV = "1.1"
 
 SRC_URI = "\
-	file://sudoers \
 	file://udev.rules \
 "
 
@@ -44,9 +43,6 @@ useradd_preinst:append () {
 
 
 do_install:append () {
-	install -d ${D}${sysconfdir}/sudoers.d/
-	install --mode=0660 ${S}/sudoers ${D}${sysconfdir}/sudoers.d/${PN}
-
 	install -d ${D}${sysconfdir}/udev/rules.d
 	install --mode=0644 ${S}/udev.rules ${D}${sysconfdir}/udev/rules.d/90-${PN}.rules
 }
@@ -54,18 +50,8 @@ do_install:append () {
 
 ## subpackages
 PACKAGE_BEFORE_PN += " \
-	${PN}-sudo \
 	${PN}-udev \
 "
-# -sudo : sudo integration
-SUMMARY:${PN}-sudo = "${SUMMARY} - sudo integration"
-FILES:${PN}-sudo = "${sysconfdir}/sudoers.d/*"
-CONFFILES:${PN}-sudo = "${sysconfdir}/sudoers.d/*"
-RDEPENDS:${PN}-sudo = "\
-	niacctbase \
-	sudo-lib \
-"
-
 # -udev : udev rules for the `ni` group
 SUMMARY:${PN}-udev = "${SUMMARY} - udev rules"
 FILES:${PN}-udev = "${sysconfdir}/udev/*"
@@ -74,5 +60,4 @@ RDEPENDS:${PN}-udev += " niacctbase udev"
 
 
 RDEPENDS:${PN} += " ${PN}-udev"
-RRECOMMENDS:${PN} += " ${PN}-sudo"
 BBCLASSEXTEND = "native nativesdk"
--- a/recipes-ni/niacctbase/niacctbase/sudoers
+++ b/recipes-ni/niacctbase/niacctbase/sudoers
@@ -1,3 +1,0 @@
-
-# Allow the admin user (UID 0) to invoke sudo without a password
-admin ALL=(ALL) NOPASSWD: ALL
--- a/conf/distro/nilrt.conf
+++ b/conf/distro/nilrt.conf
@@ -34,6 +34,11 @@ NILRT_BSI_FSTYPE:x64 = "tar.gz"
 # Use tar.xz on ARM to save space and prevent BSI installation failure.
 NILRT_BSI_FSTYPE:xilinx-zynq = "tar.xz"
 
+# Root home. NIAuth removed in dist-next (was /home/admin); use /root to match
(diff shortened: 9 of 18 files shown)


shubhanshumt26:
fyi: @doshirohan

recipes-core/images/includes/nilrt-proprietary.inc
```diff
@@ -6,7 +6,6 @@ NI_PROPRIETARY_SAFEMODE_PACKAGES = "\
 
 NI_PROPRIETARY_COMMON_PACKAGES = "\
         libnitargetcfg \
-        ni-auth \
         ni-auth-webservice \
```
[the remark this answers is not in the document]
shubhanshumt26: `ni-auth-webservice` has a [strict dependency](https://dev.azure.com/ni/DevCentral/_git/labview?path=/src/platform_shared/infrad/niauth_installer/niauthsvc_installer_rt_ipk.mak&version=GBmain&line=24&lineEnd=25&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) on ni-auth.
And as part of [US](https://dev.azure.com/ni/DevCentral/_workitems/edit/3650235) we don't want to install it explicitly in `dist-next` so later when dependency on `ni-auth-webservice` is removed this would be removed automatically. 

Also, I have validated that `ni-auth` is getting pulled by `ni-auth-webservice` on Dist-Next with my changes (See Testing Section) and the custom ni-auth packages that got generated by this [PR](https://dev.azure.com/ni/DevCentral/_git/labview/pullrequest/1313659).

recipes-core/images/includes/nilrt-image-base.inc
```diff
@@ -23,6 +23,28 @@ do_image_build_test() {
...
+#   * allow-empty-password - rewrite 'pam_unix.so nullok_secure' to plain
+#                            'nullok' in /etc/pam.d/* and set sshd
+#                            PermitEmptyPasswords. nullok_secure only permits
+#                            blank passwords from ttys listed in /etc/securetty,
+#                            which excludes SSH (PAM_TTY=ssh), so without this an
+#                            empty root password could not log in over SSH.
+#
+# Note: this previously "worked" only because login went through pam_niauth, not
+# pam_unix; with pam_niauth gone we must let pam_unix accept the empty password.
+IMAGE_FEATURES += "empty-root-password allow-empty-password"
```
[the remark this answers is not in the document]
shubhanshumt26: This was the behavior that we were following for `admin` user, so we'll be following the same for `root` user too.

recipes-core/images/includes/nilrt-image-base.inc
```diff
@@ -23,6 +23,28 @@ do_image_build_test() {
...
+# Guarantee a usable LOCAL root login.
+#
+# Historically the default interactive login was the NIAuth-backed 'admin' user
+# (empty password), which was only visible/usable because NIAuth's NSS and PAM
+# modules were pulled in transitively by ni-auth. As of dist-next, ni-auth no
+# longer installs those modules, so the NIAuth 'admin'/"" login is gone. Make the
+# local root account loginable out of the box (the box ships open and is expected
+# to be secured on first configuration; see US 3111349) with two image features:
+#
+#   * empty-root-password  - keep root's /etc/shadow password field empty
```
chaitu236, line 35: FYI: I expect this will need to be removed later for SOFU.

recipes-kernel/kernel-tests/files/iperf-load
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 source "$(dirname "$0")"/common.cfg
-source /home/admin/.iperf.info
+source /home/root/.iperf.info
```
chaitu236: We'll be updating these to `/root` right?

recipes-ni/niacctbase/niacctbase/sudoers
```diff
@@ -1,3 +0,0 @@
-
-# Allow the admin user (UID 0) to invoke sudo without a password
-admin ALL=(ALL) NOPASSWD: ALL
```
chaitu236, line 3: Does `sudo` work on dist-next without a password?

recipes-core/images/includes/nilrt-proprietary.inc
```diff
@@ -6,7 +6,6 @@ NI_PROPRIETARY_SAFEMODE_PACKAGES = "\
 
 NI_PROPRIETARY_COMMON_PACKAGES = "\
         libnitargetcfg \
-        ni-auth \
```
chaitu236, line 9: How are we coordinating these changes with ni-auth's pam being removed? Let's follow up on offline

chaitu236 (changes requested):

recipes-kernel/kernel-tests/files/iperf-load
shubhanshumt26: Yes, updated them to `/root`.

recipes-ni/niacctbase/niacctbase/sudoers
shubhanshumt26: yes, updated the Testing Snippets.

recipes-core/images/includes/nilrt-proprietary.inc
shubhanshumt26: We have already posted an announcement regarding breaking the dependencies of ni-auth. However, in the context of removing PAM from dist-next BSI, since it's only required by ni-auth, I believe we only need to be aware of this change. Is that correct?

recipes-core/images/includes/nilrt-image-base.inc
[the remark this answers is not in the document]
shubhanshumt26: allow-empty-password is a standard OE-core image feature. It's registered in IMAGE_FEATURES[validitems] and implemented via ssh_allow_empty_password in oe-core's [image.bbclass](vscode-file://vscode-app/c:/Users/shtripat/AppData/Local/Programs/Microsoft%20VS%20Code/fc3def6774/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[rootfs-postcommands.bbclass](vscode-file://vscode-app/c:/Users/shtripat/AppData/Local/Programs/Microsoft%20VS%20Code/fc3def6774/resources/app/out/vs/code/electron-browser/workbench/workbench.html), both inherited by every image recipe. So it passes feature validation and is applied at rootfs time in all builds — no meta-nilrt-local definition is needed. Already verified locally by building/booting the image.

recipes-ni/niacctbase/niacctbase.bb
```diff
@@ -40,33 +39,22 @@ USERADD_PARAM:${PN} = " \
 useradd_preinst:append () {
 	eval ${PSEUDO} chmod g+sw ${SYSROOT}/home/${LVRT_USER} || true
 	eval ${PSEUDO} chmod g+s ${SYSROOT}/home/webserv || true
-	eval ${PSEUDO} ln -sf /home/admin ${SYSROOT}/home/root || true
+	# Keep legacy /home/admin working by pointing it at root's home (/root).
+	# Reverses the old NIAuth wiring (/home/root -> /home/admin).
+	eval ${PSEUDO} ln -sf /root ${SYSROOT}/home/admin || true
```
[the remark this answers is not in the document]
shubhanshumt26: It's a false positive for how these images are actually built: opkg unpacks (preinst → symlink) before it configures (postinst → mkdir -p), so ni-auth can't create `/home/admin` before the symlink exists. 
Also we have confirmed that no one else created `/home/admin`. The only real-dir hazard would be an out-of-band `mkdir /home/admin` in a separate, earlier transaction, which doesn't occur here.

Also, we have verified the same with built image on the target.

image

shubhanshumt26:
FYI: @chaitu236 You can now review this and [nilrt PR](https://github.com/ni/nilrt/pull/354).

recipes-core/images/includes/nilrt-image-base.inc
```diff
@@ -23,6 +23,28 @@ do_image_build_test() {
...
+#                            (otherwise the rootfs postprocess zaps it to '*').
+#   * allow-empty-password - rewrite 'pam_unix.so nullok_secure' to plain
+#                            'nullok' in /etc/pam.d/* and set sshd
+#                            PermitEmptyPasswords. nullok_secure only permits
+#                            blank passwords from ttys listed in /etc/securetty,
+#                            which excludes SSH (PAM_TTY=ssh), so without this an
+#                            empty root password could not log in over SSH.
+#
+# Note: this previously "worked" only because login went through pam_niauth, not
```
chaitu236: > Note: this previously "worked" only because login went through pam_niauth, not
pam_unix; with pam_niauth gone we must let pam_unix accept the empty password.

Remove this comment. No need to mention how things used to work now that pam_niauth will be history.

recipes-kernel/kernel-tests/files/iperf-load
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 source "$(dirname "$0")"/common.cfg
-source /home/admin/.iperf.info
+source /root/.iperf.info
```
chaitu236, line 3: `rtos_oetests` (which copy files to/from `/home/admin`) are expected to still work after this change as we'll have `/home/admin -> /root` symlink right?

recipes-ni/niacctbase/niacctbase.bb
```diff
@@ -39,6 +39,9 @@ USERADD_PARAM:${PN} = " \
 useradd_preinst:append () {
 	eval ${PSEUDO} chmod g+sw ${SYSROOT}/home/${LVRT_USER} || true
 	eval ${PSEUDO} chmod g+s ${SYSROOT}/home/webserv || true
+	# Keep legacy /home/admin working by pointing it at root's home (/root).
+	# Reverses the old NIAuth wiring (/home/root -> /home/admin).
```
chaitu236: > Reverses the old NIAuth wiring (/home/root -> /home/admin).

No need to mention this. That's history now.

chaitu236 (approved):

recipes-core/images/includes/nilrt-proprietary.inc
chaitu236: You may close this when syscfg changes to handle both root, admin accounts are done or we have a reasonable plan to address ATS failures, etc.

recipes-ni/niacctbase/niacctbase.bb
shubhanshumt26: Removed.

recipes-core/images/includes/nilrt-image-base.inc
shubhanshumt26: Removed.

recipes-kernel/kernel-tests/files/iperf-load
shubhanshumt26: Yes, we shouldn't have any issues here since `/home/admin` will continue to work as before via the `/home/admin -> /root` symlink.

recipes-ni/ni-tracefs-utils/files/traceextract
```diff
@@ -4,5 +4,5 @@ set -e
 
 sudo -u lvuser touch "$1"
 chmod 440 "$1"
-chown admin:administrators "$1"
+chown root:root "$1"
```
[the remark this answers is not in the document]
shubhanshumt26: Thanks for flagging — but the file was never readable by `lvuser`, so this isn't a regression.

**Intent:** the introducing commit ([f73b9562](https://github.com/ni/meta-nilrt/commit/f73b9562947ff9e38767faa22523bb0d4e4900aa)) states the script "ensure[s] trace data can only be written to files with admin permissions." `lvuser` only *triggers* extraction via sudo; it's not meant to own, modify, or read the output.

**Why `admin:administrators` → `root:root` is correct:** `admin` and `root` are both uid 0 (owner unchanged), and `lvuser` was never in `administrators`, so with mode `440` it had no read access before either. With NIAuth removed, `admin`/`administrators` no longer resolve, so `chown admin:administrators` fails under `set -e` — `root:root` fixes that while preserving the exact access model.

Verified on hardware:

```text
# OLD (file was admin:administrators 440)
groups lvuser -> tty video plugdev network niwscerts ni   # NOT in 'administrators' => cannot read

# NEW (file is root:root 440)
groups lvuser -> ni tty video plugdev network niwscerts   # NOT in 'root' => still cannot read
```

Identical behavior. The suggested `root:ni` would actually *grant* `lvuser` new read access (it's in `ni`), contradicting the admin-only intent — so I'll keep `root:root`.

recipes-core/images/includes/nilrt-proprietary.inc
shubhanshumt26: Did you mean this [bug](https://dev.azure.com/ni/DevCentral/_workitems/edit/3930353) or the [MTO Feature](https://ni.visualstudio.com/DevCentral/_workitems/edit/3951308)?
Also, syscfg team had mentioned that they will create a feature for this to handle workflows without ni-auth etc. from Hardware Manager and that isn't committed for this cycle as changes are in dist-next.

chaitu236: I originally meant MTO feature as I was not aware of the bug or that syscfg needed to make other changes. So what's the plan regarding this PR (and https://github.com/ni/nilrt/pull/354)? We'll wait until that syscfg bug is addressed and syscfg handles other workflows?

shubhanshumt26: Oh, ok. 
For this & nilrt PR the plan is we'll merge them together with the ni-auth dependency break PR post ci-hardening completes. That's expected to complete in starting of next week.

And we can wait for sycfg to resolve this bug but I don't think it will be fixed soon and also if they fix it, it won't fix the functionality we require through Hardware Manager (basic ops like Format/Installing BSI etc. will be still broken).

In any case we shouldn't wait for sycfg to handle other workflows from Hardware Manager as that's not committed for this cycle from their side and this major change we should make sooner so if any thing breaks (hope it'll not) we can fix that early and others team also get sufficient time to adapt if they have to.

chaitu236: Ok, if we're confident that merging this PR won't break ATSes, it's ok to merge.

rajendra-desai-ni (approved):

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -36,10 +36,20 @@ check_hash_equality()
 hash_sanity_check()
 {
     SHADOW_HASH="$1"
-    
-    if [ -z "$SHADOW_HASH" ] || [ "$SHADOW_HASH" = "*" ] || [ "$SHADOW_HASH" = "!" ]; then
+
+    # Empty (blank root) or bare '!' (deliberate passwd -l of a passwordless
```
shubhanshumt26, line 40: FYI @erickshepherdNI As discussed, `hash_sanity_check` now skips the stale `*` (the `zap_empty_root_password` artifact) so it can't re-lock root from a `.shadow` carried across reflashes. Everything else stays as-is; I only added handling for a `!`-prefixed hash so a deliberately locked root password (`passwd -l`) still syncs. 

cc: @chaitu236 @rajendra-desai-ni

recipes-kernel/kernel-tests/files/run-docker-cyclictest
```diff
@@ -22,7 +22,7 @@ case "$#" in
 		       --cpuset-cpus "$2" \
 		       -v ${LOG_DIR}:${LOG_DIR} \
 		       -v ${TEST_DIR}:${TEST_DIR} \
-		       -v /home/admin:/home/admin \
+		       -v /root:/root \
```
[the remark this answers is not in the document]
shubhanshumt26: Keeping it as it is, as before also we were accessing the HOME_DIR and now also we are accessing the same.

recipes-kernel/kernel-tests/files/run-docker-cyclictest
```diff
@@ -6,7 +6,7 @@ case "$#" in
 		docker run --privileged --network=host \
 		       -v ${LOG_DIR}:${LOG_DIR} \
 		       -v ${TEST_DIR}:${TEST_DIR} \
-		       -v /home/admin:/home/admin \
+		       -v /root:/root \
```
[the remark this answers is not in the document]
shubhanshumt26: Keeping it as it is, as before also we were accessing the HOME_DIR and now also we are accessing the same.

recipes-core/images/includes/nilrt-image-base.inc
```diff
@@ -23,6 +23,26 @@ do_image_build_test() {
...
+#   * empty-root-password  - keep root's /etc/shadow password field empty
+#                            (otherwise the rootfs postprocess zaps it to '*').
+#   * allow-empty-password - rewrite 'pam_unix.so nullok_secure' to plain
+#                            'nullok' in /etc/pam.d/* and set sshd
+#                            PermitEmptyPasswords. nullok_secure only permits
+#                            blank passwords from ttys listed in /etc/securetty,
+#                            which excludes SSH (PAM_TTY=ssh), so without this an
+#                            empty root password could not log in over SSH.
+#
+IMAGE_FEATURES += "empty-root-password allow-empty-password"
```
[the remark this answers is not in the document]
shubhanshumt26: This is expected & later we have plans to ship with an expired password.

recipes-ni/ni-acctsync/files/ni-acctsync
```diff
@@ -55,7 +65,8 @@ hash_sanity_check()
     case "$ENCRYPT_METHOD" in
         SHA512)
             log "Checking sanity of SHA512 hash..."
-            if [[ "$SHADOW_HASH" =~ ^\$6\$[A-Za-z0-9./]{1,16}\$[A-Za-z0-9./]{86}$ ]]; then
+            # Optional leading '!' allows a deliberate 'passwd -l' lock (!$6$...) to sync.
+            if [[ "$SHADOW_HASH" =~ ^!?\$6\$[A-Za-z0-9./]{1,16}\$[A-Za-z0-9./]{86}$ ]]; then
```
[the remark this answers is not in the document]
shubhanshumt26: Thanks — good point, but this `[[ … =~ ]]` predates this PR (it was already in `hash_sanity_check`); my change only added the optional `!`. NILRT's sh supports it in practice, so I'll keep the existing behavior here and leave a POSIX (`grep -Eq`) conversion as a separate follow-up to avoid extra testing.

FYI @erickshepherdNI  — you wrote `ni-acctsync` originally; flagging in case you have a preference.

---

PR #1047, opened 2026-07-10
aide: integrate host file-integrity monitoring into the NILRT base image

Vairamani-NI:
### Summary of Changes

Add AIDE (Advanced Intrusion Detection Environment) as the host file-integrity monitoring (FIM) capability for NILRT, baked into the base system image (BSI):

- **Packaging:** add aide to packagegroup-ni-runmode so it ships on every device.
- **NI monitoring policy:** ship a managed /etc/aide.conf (overrides the upstream sample) monitoring the security-relevant, normally-immutable trees (/bin /sbin /lib /lib64 /usr /boot /etc) and kernel modules (.ko), excluding only volatile/runtime paths (incl. depmod-regenerated modules.* indices, /etc/natinst/share, DHCP/salt first-boot artifacts).
- **First-boot baseline:** sysvinit one-shot aide-init (S99, idempotent, backgrounded) builds the baseline on first boot; adds a reinit action for controlled rebuilds.
- **Scheduled checks:** aide-check wrapper runs via cron daily at 00:00 UTC, reporting a one-line summary to syslog (daemon.info clean / daemon.warning on differences) with per-run reports under /var/log/aide (logrotate + self-pruned).
- **Tamper-evident baseline:** DB stored outside monitored dirs (/var/lib/aide, 0700), sha256 sidecar, both made immutable (chattr +i), hash audited to syslog; checks fail closed (distinct exit codes) if the baseline can't be verified.
- **Runtime deps:** cronie, coreutils (for sha256sum, which the NILRT BusyBox config disables).
- **Review-driven hardening:**
  - Explicit PATH in aide-init and aide-check for boot-time/manual reliability.
  - Concurrent-run prevention via flock in both aide-init and aide-check; lockfile-open failures (e.g. run not writable) and flock error codes (rc=1 contention vs rc≥2 error) are distinguished and logged explicitly — no silent skips.
  - BusyBox-compatible unique report creation using mktemp template ending with XXXXXX, then rename to .log.
  - Removed awk dependency from checksum parsing (POSIX shell parsing only).
  - aide-init unknown action now prints usage and exits non-zero.
  - Orphan checksum handling when DB is missing.
  - Logrotate covers both aide.log and aide-init.log with root-only recreate policy.
  - Policy tightened for RT impact by using NORMAL = FIPSR (no redundant second content hash pass).

### Justification

https://dev.azure.com/ni/DevCentral/_workitems/edit/3470293

### Testing

**Build**

- nilrt-base-system-image builds for x64 and ARM32 (xilinx-zynq) via standard pipeline (build.core-feeds.sh and build.core-images.sh). No new build/QA warnings attributable to AIDE.
- AIDE packaging/layout checks pass: aide and aide-lic appear in nilrt-runmode-rootfs manifest; /etc/aide.conf ships NI policy override; /etc/init.d/aide-init (S99), /usr/sbin/aide-check, /etc/cron.d/aide-check, /etc/logrotate.d/aide, and /var/lib/aide (0700) are installed.
- ARM (xilinx-zynq): applicable to both x64 and ARM (no arch-specific code).

**On-target validation (x64, cRIO-9045, kernel 6.12.94-rt18 PREEMPT_RT)**

- Image reinstalled and validated on device.
- First boot: aide-init auto-builds baseline; DB and checksum created and protected.
- Permissions validated: /var/lib/aide 0700, /var/log/aide 0700, aide.db.gz 0600, aide.db.gz.sha256 0600.
- Clean check: /usr/sbin/aide-check returns exit 0 on clean image; per-run report created with unique mktemp-based name.
- Concurrency: concurrent aide-check invocation skips cleanly with syslog message; aide-init invalid action prints usage and returns exit 1.
- Persistent logging: aide.log written by AIDE via report_url; aide-init.log captures baseline build output; logrotate config present for both.
- Daily cron check confirmed running; sample report verified correct detection — post-baseline package install (rt-tests) flagged with 32 added entries including full filenames, hashes, and old vs new attribute diffs.
- RT impact: ran cyclictest at RT priority 80 (100k loops) with and without aide --check running concurrently; max latency 33 µs, no change under AIDE load. AIDE runs at normal priority and cannot preempt RT threads.

**On-target validation (ARM32, cRIO-9068, Cortex-A9, kernel 4.14.146-rt67-ni PREEMPT_RT)**

- BSI flashed and AIDE verified on device.
- aide --init: 2m21s wall time; aide --check: 2m20s — expected for single-core UBIFS-backed storage.
- Peak RSS ~59 MiB on 512 MiB device.
- RT cyclictest stress run — Min: 10 µs / Avg: 16 µs / Max: 60 µs  
(100,000 iterations, SCHED_FIFO prio 80, CPU0, concurrent aide --check, kernel 4.14.146-rt67-ni PREEMPT_RT)

**Footprint**

- AIDE installed size ~290 KiB (/usr/bin/aide ~253 KiB + aide-lic ~35 KiB); runtime libs already in base.
- Baseline DB aide.db.gz ~4.75 MiB for monitored system trees and kernel modules.
- Observed peak RSS ~45 MiB during --init and ~59 MiB during --check.

**Known gap / follow-up**

opkg 0.9.0 has no global post-transaction hook. A package install or update via opkg will cause AIDE to flag changed files as violations until the baseline is manually updated (run /etc/init.d/aide-init reinit after any opkg operation). Automatic rebaseline on package operations will be addressed in a successor feature: the team has decided to patch opkg to add a post-invoke hook, with a follow-up AIDE integration work item to wire the hook into aide-init reinit.

### Procedure

### DIFF
--- a/recipes-core/packagegroups/packagegroup-ni-runmode.bb
+++ b/recipes-core/packagegroups/packagegroup-ni-runmode.bb
@@ -6,6 +6,7 @@ PACKAGE_ARCH = "${MACHINE_ARCH}"
 inherit packagegroup
 
 RDEPENDS:${PN} = "\
+	aide \
 	cpp \
 	cpp-symlinks \
 	dosfstools \
--- a/recipes-ids/aide/aide/aide-check
+++ b/recipes-ids/aide/aide/aide-check
@@ -1,0 +1,132 @@
+#!/bin/sh
+# Periodic AIDE file-integrity check. A one-line summary of each run is sent to
+# syslog (syslog-ng via /dev/log) for alerting; the full per-run report is
+# written under /var/log/aide/ for detail.
+#
+# Exit status:
+#   0      no differences
+#   1|2|4  aide(1) difference bitmask (new|removed|changed); e.g. 7 = all three
+#   8      baseline could not be verified (checksum missing or mismatch); no scan run
+#   9      IO/setup error (cannot create or secure the log dir/report); no scan run
+#   >=14   aide(1) error (e.g. 18 = IO error) - details are in the per-run report
+
+# Set an explicit PATH so aide, sha256sum, mktemp, grep, tail, etc.
+# resolve for manual invocations too, not just the cron entry which sets PATH.
+PATH=/usr/sbin:/usr/bin:/sbin:/bin
+
+DB=/var/lib/aide/aide.db.gz
+LOGDIR=/var/log/aide
+TAG=aide-check
+INITLOCK=/run/aide-init.lock
+CHECKLOCK=/run/aide-check.lock
+
+# Reports can contain file paths and hashes -> keep them root-only. Fail closed
+# if the log directory cannot be created or secured, so we never run a scan we
+# cannot record.
+umask 077
+if ! mkdir -p "$LOGDIR" || ! chmod 700 "$LOGDIR"; then
+	logger -t "$TAG" -p daemon.err "cannot create or secure $LOGDIR; skipping check (IO error)"
+	exit 9
+fi
+
+# Prevent concurrent aide-check runs (e.g. a manual run during the daily cron
+# job). flock holds the advisory lock via fd 9 for the lifetime of this process;
+# the kernel releases it automatically on process death, so no PID-based
+# stale recovery is needed.
+if ! exec 9>"$CHECKLOCK" 2>/dev/null; then
+	logger -t "$TAG" -p daemon.err "cannot open lock file $CHECKLOCK; skipping check (IO error)"
+	exit 9
+fi
+flock -n 9; _rc=$?
+if [ "$_rc" -eq 1 ]; then
+	logger -t "$TAG" -p daemon.info "another aide-check is running; skipping"
+	exit 0
+elif [ "$_rc" -ne 0 ]; then
+	logger -t "$TAG" -p daemon.err "flock failed on $CHECKLOCK (rc=$_rc); skipping check (IO error)"
+	exit 9
+fi
+
+# No baseline database present. Distinguish three cases:
+#  - aide-init is mid-build (its lock is held): the checksum is written before
+#    the DB is promoted, so "checksum present, DB missing" is expected -> skip.
+#  - checksum exists with no build running: the DB was removed after
+#    provisioning (deletion/tampering) -> fail closed.
+#  - neither DB nor checksum: benign first boot before the baseline exists.
+if [ ! -f "$DB" ]; then
+	# flock -n fails when aide-init's background scan holds the lock (build in
+	# progress) and succeeds when nobody holds it. Stale locks are impossible:
+	# the kernel releases flock automatically when the holding process dies.
+	flock -n "$INITLOCK" true 2>/dev/null; _rc=$?
+	if [ "$_rc" -eq 1 ]; then
+		logger -t "$TAG" -p daemon.info "baseline build in progress; skipping integrity check"
+		exit 0
+	elif [ "$_rc" -ne 0 ]; then
+		logger -t "$TAG" -p daemon.err "cannot probe baseline-build lock $INITLOCK (rc=$_rc); skipping check (IO error)"
+		exit 9
+	fi
+	if [ -f "$DB.sha256" ]; then
+		logger -t "$TAG" -p daemon.crit "baseline database $DB missing but checksum present - possible deletion/tampering; refusing to run (run: /etc/init.d/aide-init reinit)"
+		exit 8
+	fi
+	logger -t "$TAG" -p daemon.info "no baseline database yet; skipping integrity check"
+	exit 0
+fi
+
+# Verify the baseline database itself has not been tampered with before we trust
+# it (the runmode rootfs is writable, so the DB is a target too). Fail closed: if
+# the baseline cannot be trusted, do NOT emit a (potentially false) "OK" - alert
+# and stop so an operator re-initialises the baseline (aide-init).
+if [ ! -f "$DB.sha256" ]; then
+	logger -t "$TAG" -p daemon.warning "baseline checksum missing; cannot verify baseline integrity - skipping check (run: /etc/init.d/aide-init reinit)"
+	exit 8
+fi
+want=$(cat "$DB.sha256" 2>/dev/null)
+have=$(sha256sum "$DB" 2>/dev/null)
+have=${have%% *}
+if [ -z "$want" ] || [ -z "$have" ]; then
+	# Could not read the checksum or hash the DB -> IO error, not tampering.
+	logger -t "$TAG" -p daemon.err "cannot read baseline checksum or hash the database (IO error); skipping check"
+	exit 9
+fi
+if [ "$want" != "$have" ]; then
+	logger -t "$TAG" -p daemon.crit "BASELINE DATABASE CHECKSUM MISMATCH - possible tampering; refusing to run check; investigate then run: /etc/init.d/aide-init reinit (want=$want have=$have)"
+	exit 8
+fi
+
+# Create a unique per-run report. mktemp guarantees uniqueness (so two runs in
+# the same second cannot overwrite each other's report) and fails closed: if the
+# log volume is full or read-only we skip rather than run a scan whose output
+# cannot be recorded, which would make the exit status meaningless. Keep XXXXXX
+# at the very end of the template (BusyBox mktemp requires it there), then rename
+# to add the .log suffix.
+tmp=$(mktemp "$LOGDIR/aide-check.$(date +%Y%m%d-%H%M%S).XXXXXX" 2>/dev/null) || {
+	logger -t "$TAG" -p daemon.err "cannot create report file in $LOGDIR; skipping check (IO error)"
+	exit 9
+}
+report="$tmp.log"
+if ! mv "$tmp" "$report" 2>/dev/null; then
+	logger -t "$TAG" -p daemon.err "cannot rename report file ($tmp -> $report); skipping check (IO error)"
+	rm -f "$tmp" 2>/dev/null
+	exit 9
+fi
+
+aide --check >> "$report" 2>&1
+rc=$?
+
+if [ "$rc" -eq 0 ]; then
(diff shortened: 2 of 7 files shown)


Vairamani-NI:
@ni/rtos Please review this.

chaitu236 (changes requested):
Many commits are unsigned. Also, many related changes seem to be spread across multiple commits please squash them.

Vairamani-NI:
> Many commits are unsigned. Also, many related changes seem to be spread across multiple commits please squash them.

Done.

recipes-core/packagegroups/packagegroup-ni-runmode.bb
```diff
@@ -6,6 +6,7 @@ PACKAGE_ARCH = "${MACHINE_ARCH}"
 inherit packagegroup
 
 RDEPENDS:${PN} = "\
+	aide \
```
chaitu236, line 9: PR description says

> ARM (xilinx-zynq): pending (not yet built).

but the change here is applicable to both x64 and ARM.

What is the scope of this change - is ARM expected to contain `aide`? If so, please update PR description.

recipes-ids/aide/aide/aide-check
```diff
@@ -0,0 +1,168 @@
...
+#   1|2|4  aide(1) difference bitmask (new|removed|changed); e.g. 7 = all three
+#   8      baseline could not be verified (checksum missing or mismatch); no scan run
+#   9      IO/setup error (cannot create or secure the log dir/report); no scan run
+#   >=14   aide(1) error (e.g. 18 = IO error) - details are in the per-run report
+
+# Set an explicit PATH so aide, sha256sum, mktemp, nice, ionice, grep, tail, etc.
+# resolve for manual invocations too, not just the cron entry which sets PATH.
+PATH=/usr/sbin:/usr/bin:/sbin:/bin
+
+DB=/var/lib/aide/aide.db.gz
```
chaitu236, line 17: Where does this DB come from. Is that something we package or does it get generated on target?

recipes-ids/aide/aide/aide-check
```diff
@@ -0,0 +1,168 @@
...
+LOGDIR=/var/log/aide
```
chaitu236, line 18: How frequently does this logfile get updated? Please note that `/var/log/` logs end up on disk and if the logging frequency is high this can degrade disk.

recipes-ids/aide/aide/aide-check
```diff
@@ -0,0 +1,168 @@
...
+# Reports can contain file paths and hashes -> keep them root-only. Fail closed
+# if the log directory cannot be created or secured, so we never run a scan we
+# cannot record.
+umask 077
+if ! mkdir -p "$LOGDIR" || ! chmod 700 "$LOGDIR"; then
+	logger -t "$TAG" -p daemon.err "cannot create or secure $LOGDIR; skipping check (IO error)"
+	exit 9
+fi
+
+# Prevent concurrent aide-check runs (e.g. a manual run during the daily cron
```
chaitu236, line 32: Does `aide` not have support to prevent concurrent runs? If so, can't we just use `flock` instead of rolling out our own implementation?

recipes-ids/aide/aide/aide-check
```diff
@@ -0,0 +1,168 @@
...
+	exit 8
+fi
+
+# Create a unique per-run report. mktemp guarantees uniqueness (so two runs in
+# the same second cannot overwrite each other's report) and fails closed: if the
+# log volume is full or read-only we skip rather than run a scan whose output
+# cannot be recorded, which would make the exit status meaningless. Keep XXXXXX
+# at the very end of the template (BusyBox mktemp requires it there), then rename
+# to add the .log suffix.
+tmp=$(mktemp "$LOGDIR/aide-check.$(date +%Y%m%d-%H%M%S).XXXXXX" 2>/dev/null) || {
```
chaitu236, line 102: How big is this report usually?

recipes-ids/aide/aide/aide-check
```diff
@@ -0,0 +1,168 @@
...
+	exit 9
+}
+report="$tmp.log"
+if ! mv "$tmp" "$report" 2>/dev/null; then
+	logger -t "$TAG" -p daemon.err "cannot rename report file ($tmp -> $report); skipping check (IO error)"
+	rm -f "$tmp" 2>/dev/null
+	exit 9
+fi
+
+# De-prioritise CPU and I/O to protect RT determinism; use ionice if available.
```
chaitu236: Do we really need to do this? i.e., have we determined that without this RT determinism is affected? I don't want us to prematurely optimize without determining that this is an issue.

recipes-ids/aide/aide/aide-check.cron
```diff
@@ -0,0 +1,6 @@
+# Daily AIDE file-integrity check. The wrapper logs a one-line summary to syslog
+# and writes the full per-run report under /var/log/aide/.
+# Runs at 03:17 to avoid busy periods; de-prioritised to protect RT determinism.
```
chaitu236: How will running at 03:17 avoid busy periods? Why not just use GMT 00:00 or some standard time. And I don't think we need a comment for this.

recipes-ids/aide/aide_%.bbappend
```diff
@@ -0,0 +1,58 @@
...
+# duplicate entries.
+SRC_URI += "file://aide.conf"
+AIDE_INCLUDE_DIRS = ""
+AIDE_SKIP_DIRS = ""
+
+# First-boot baseline initialization (sysvinit).
+SRC_URI += "file://aide-init"
+inherit update-rc.d
+INITSCRIPT_NAME = "aide-init"
+INITSCRIPT_PARAMS = "defaults 99"
```
chaitu236, line 18: Does the initscript run at startup? Wouldn't that be a timetaking process (and possibly block boot sequence)?

recipes-ids/aide/aide/aide.conf
```diff
@@ -0,0 +1,57 @@
...
+
+# ---- attribute rule groups ----
+# One strong content hash (sha256) is sufficient for tamper detection; adding a
+# second digest (e.g. sha512) would hash every file twice, enlarging the DB and
+# lengthening each scan for no security gain - avoided to limit RT impact.
+FIPSR  = p+u+g+s+acl+xattrs+sha256
+NORMAL = FIPSR
+DIR    = p+u+g+acl+xattrs
+
+# ---- monitored: security-relevant, normally-immutable system directories ----
```
chaitu236, line 29: > normally-immutable system directories

The dirs get updated when s/w is installed/upgraded. How is this handled?

chaitu236 (changes requested):

recipes-ids/aide/aide/aide-check
chaitu236: nvm, PR description mentions it gets created on first boot.

recipes-core/packagegroups/packagegroup-ni-runmode.bb
Vairamani-NI: For now, I haven't tested in ARM so that I have mentioned it, I will check with Gratian on the scope on this.

recipes-ids/aide/aide/aide-check
Vairamani-NI: For now, this will log only once, but we need to discuss with the team on the frequency of this check (on intrusion detection).

recipes-ids/aide/aide/aide-check
Vairamani-NI: Typically, 1-3 KB. Even if it gets bigger, logrotate caps it at 1M before rotating.

recipes-ids/aide/aide/aide-check
Vairamani-NI: I thought flock is just for file locking but after researching it does exactly what the custom code does, I will update it.

recipes-ids/aide/aide/aide-check
Vairamani-NI: While testing on the cRIO, I have noticed CPU/IO spikes from aide on the device without these flags, I can do more testing and update the results in the testing and if this is not needed we can remove it.

recipes-ids/aide/aide_%.bbappend
Vairamani-NI: The init script returns immediately, basically it just checks some condition and starts the `aide --init` command and that runs in the background.

recipes-ids/aide/aide/aide.conf
Vairamani-NI: You mean at the runtime, right? That will be flagged as intrusion, as per the design. We need to create a hook somewhere(?) to run `aide-init --reinit` to update the baseline, so the next check won't flag it.

recipes-ids/aide/aide/aide-check.cron
Vairamani-NI: Done.

recipes-ids/aide/aide/aide-check
Vairamani-NI: Done.

recipes-ids/aide/aide/aide-check
chaitu236: Does this create any RT threads? If not, it shouldn't affect RT determinism; RT processes will have priority over non-RT processes so they will always get CPU.
CPU/IO spikes are not necessarily problematic - that means the process is making full use of resources.

Vairamani-NI: Removed.

recipes-core/packagegroups/packagegroup-ni-runmode.bb
Vairamani-NI: tested in arm and is documented the results both in the WI and in the PR desc.

recipes-ids/aide/aide/aide.conf
Vairamani-NI: @chaitu236 This will be done as a separate work, and I will create a doc for these as well.

chaitu236 (approved):

PR #1056, opened 2026-08-05
ni-firewalld-servicedefs: add port-opening helper for NI services

Vairamani-NI:
### Summary of Changes
Adds a firewalld port-opening mechanism to `ni-firewalld-servicedefs` so NI services can be opened/closed in the default zone and wires the package to open the always-on core NI services (including the NI System Web Server) automatically.

- **New helper:** `ni-firewall-open` (POSIX sh) with `ni-firewall-close` as a symlink; the action is chosen from the invoked name.
- **Dual-mode:** works **offline** (`firewall-offline-cmd`, during image build) and **live** (`firewall-cmd --permanent` + a matching runtime change, no `--reload`, so other runtime rules are preserved).
- **Idempotent:** re-running is a no-op; an unknown/missing service name exits non-zero.
- **Default-open services** — opened by `pkg_postinst_ontarget`, closed by `pkg_prerm` on removal (guarded against offline `$D` removal): `ni-service-locator, ni-mxs, ni-rpc-server, ni-logos-xt, ni-sync-remote, http, https`. The `http/https` built-ins (ports 80/443) cover the NI System Web Server, which hosts Web-Based Configuration and the nisysapi/System Configuration channel used by NI MAX and Hardware Manager, so it stays reachable with firewalld active (matching legacy firewall behavior). `RDEPENDS` now includes `firewalld-offline-cmd`.

### Testing
- **Build:** `bitbake ni-firewalld-servicedefs` and full `bitbake nilrt-base-system-image` (x64). Verified `/usr/bin/ni-firewall-open` (0755) + `ni-firewall-close` symlink + `firewall-offline-cmd` land in the runmode rootfs; manifest shows `firewalld-offline-cmd` and `ni-firewalld-servicedefs`.
- **On-hardware (cRIO-9045, x64, run mode):**
  - `postinst` opened all core NI services in both runtime and permanent config (`INSTALL_RC=0`).
  - `prerm` closed them on removal with no stale rules; zone returned to baseline (`REMOVE_RC=0`).
  - `firewalld-offline-cmd` pulled automatically via RDEPENDS.
  - Functional suite: open, idempotent re-open, close, idempotent re-close, unknown service → rc 1, no-args → rc 2, multi-service open/close.

- NI System Web Server (http/https) auto-enable after fresh image install:
  - On a freshly flashed runmode image (firewalld active) with no manual firewall config, the default zone lists http and https alongside the core NI services — opened automatically by postinst, no first-login/manual step.
  - External reachability from a separate host with firewalld active: TCP 80 and 443 OPEN (verified together with 22 and 3580/RPC).
  - Persists across reboot (written to firewalld's permanent config, not runtime-only).
  - Removed on package removal via prerm (returns zone to baseline, no stale rules).

### Justification
[AB#3964859](https://dev.azure.com/ni/DevCentral/_workitems/edit/3964859)

### Procedure

### DIFF
--- a/recipes-ni/ni-firewalld-servicedefs/files/ni-firewall-open
+++ b/recipes-ni/ni-firewalld-servicedefs/files/ni-firewall-open
@@ -1,0 +1,147 @@
+#!/bin/sh
+# ni-firewall-open / ni-firewall-close
+#
+# Open or close a firewalld service in the default zone. The same script backs
+# both commands; the action is chosen from the name it is invoked as
+# (ni-firewall-close is a symlink to ni-firewall-open).
+#
+# It works in two situations:
+#   * offline - firewalld is not running (e.g. early first boot). Changes are
+#     written with firewall-offline-cmd and take effect when firewalld starts.
+#   * live    - firewalld is running. Changes are applied to the running config
+#     (immediate) and the permanent config (persisted), without a full
+#     'firewall-cmd --reload', so other runtime rules are left undisturbed.
+#
+# Usage: ni-firewall-open  <service> [<service> ...]
+#        ni-firewall-close <service> [<service> ...]
+#
+# Re-running is idempotent. An unknown service name causes a non-zero exit.
+
+set -u
+
+# Strip the directory prefix with shell parameter expansion instead of spawning
+# basename; keeps the helper self-contained in minimal/early-boot environments.
+prog=${0##*/}
+case "$prog" in
+	ni-firewall-open)  action=add;    state=open   ;;
+	ni-firewall-close) action=remove; state=closed ;;
+	*)
+		echo "$prog: must be invoked as ni-firewall-open or ni-firewall-close" >&2
+		exit 2
+		;;
+esac
+
+if [ $# -lt 1 ]; then
+	echo "Usage: $prog <service> [<service> ...]" >&2
+	exit 2
+fi
+
+# Ask firewalld itself which service definitions exist instead of scanning a
+# hard-coded path: XMLs may live under /etc/firewalld/services or the platform
+# ${libdir} (/usr/lib or /usr/lib64), alongside firewalld's own built-ins. The
+# permanent/offline views read straight from disk, so a freshly installed
+# service file is visible without a daemon reload.
+service_exists() {
+	for s in ${available_services:-}; do
+		[ "$s" = "$1" ] && return 0
+	done
+	return 1
+}
+
+# Decide once whether firewalld is running (live) or not (offline).
+#
+# The default zone is a single global setting (DefaultZone in firewalld.conf),
+# not a per-view one: firewalld has no '--permanent --get-default-zone', and
+# --set-default-zone always updates runtime and permanent together. So the
+# runtime default zone reported here is also the permanent default zone, and it
+# is safe to target for the permanent changes below.
+if firewall-cmd --state >/dev/null 2>&1; then
+	live=1
+	zone=$(firewall-cmd --get-default-zone 2>/dev/null)
+	available_services=$(firewall-cmd --permanent --get-services 2>/dev/null) || available_services=
+else
+	live=0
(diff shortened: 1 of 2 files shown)


Vairamani-NI:
@ni/rtos please review this

recipes-ni/ni-firewalld-servicedefs/files/ni-firewall-open
```diff
@@ -0,0 +1,147 @@
+#!/bin/sh
+# ni-firewall-open / ni-firewall-close
```
rajendra-desai-ni, line 2: From what I understand, this script acts as a wrapper to `firewall-cmd` and `firewall-offline-cmd` commands with additional functionalities of some validation, idempotency checks, error reporting etc. Isn't `firewalld` as a package itself provide these functionalities? Did you check on this?

Having an additional wrapper script should only be considered when the underlying package doesn't provide required functionalities, otherwise it would be an overhead maintaining it.

recipes-ni/ni-firewalld-servicedefs/ni-firewalld-servicedefs.bb
```diff
@@ -30,17 +30,45 @@ SRC_URI = "\
...
+	ln -sf ni-firewall-open ${D}${bindir}/ni-firewall-close
+}
+
+RDEPENDS:${PN} += "firewalld firewalld-offline-cmd"
+
+# Open the always-on NI services once the package is on the target. Deferred to
+# first boot (never runs against $D) so firewalld's tools operate natively.
+pkg_postinst_ontarget:${PN} () {
+	# Best effort: never let a firewall hiccup block package configuration.
+	ni-firewall-open ${CORE_NI_SERVICES} || true
```
rajendra-desai-ni, line 63: What if we directly run `firewall-cmd` here? Why will it block package configuration if the line always tends to succeed?

recipes-ni/ni-firewalld-servicedefs/ni-firewalld-servicedefs.bb
```diff
@@ -30,17 +30,45 @@ SRC_URI = "\
...
 
-RDEPENDS:${PN} += "firewalld"
+# Close them again on package removal, leaving no stale rules. Skip during any
+# offline (image-build) removal.
+pkg_prerm:${PN} () {
+	if [ -n "$D" ]; then
+		exit 0
+	fi
+	# Best effort: never let a firewall hiccup block package removal.
+	ni-firewall-close ${CORE_NI_SERVICES} || true
```
rajendra-desai-ni, line 73: What if we directly run `firewall-cmd` here instead of a wrapper script? Why will it block package configuration if the line always tends to succeed?

rajendra-desai-ni (changes requested):
- Please squash the commits into one.
- Follow 50/72 rule for commit messages.

Vairamani-NI:
@rajendra-desai-ni 

Squashed to a single commit, corrected the message.

On the wrapper: you're right that firewall-cmd already returns 0 on ALREADY_ENABLED/NOT_ENABLED and validates input, so idempotency/validation aren't the reason. 

The reasons are: 
1. firewall-cmd needs the running daemon — during image build or before firewalld starts you must use firewall-offline-cmd, and using the wrong one desyncs runtime vs permanent;
2. it writes permanent + runtime in one step without a disruptive --reload
3. the declarative drop-in [feature](https://dev.azure.com/ni/DevCentral/_workitems/edit/3964887) and any package that opens a port reuse this one helper instead of re-implementing that offline/live logic. The || true in postinst/prerm is because a non-zero there aborts the opkg transaction, and firewall setup is best-effort — a firewall hiccup shouldn't block installing or removing the package.

I have already discussed these in the firewalld successor feature review, and we have decided to have a wrapper.

rajendra-desai-ni (approved):

chaitu236 (approved):

PR #1061, opened 2026-08-11
sshd_config: Enable KbdInteractiveAuth

erickshepherdNI:
### Summary of Changes

Updates the OpenSSH bbappend to modify the NILRT sshd_config file to enable KbdInteractiveAuth and disable PasswordAuthentication.

[AB#3918226](https://dev.azure.com/ni/DevCentral/_workitems/edit/3918226)

### Justification

This will allow users to detect an expired password in non-interactive scenarios while keeping SSH and login behaviors the same. The change is required in the scarthgap branch to support the changes already being made in Hardware Manager for our set on first use plan. Dan's PR that relies on this is here: https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/1279679

### Testing

- On a cRIO-9033 with ni-auth installed:
  - Confirmed that users are still able to log in with admin/<blank>
- On a cRIO-9033 without ni-auth
  - Confirmed that users are able to log in with root/<blank>
  - Confirmed that users are able to change the root password with passwd and then log in with it
  - Confirmed that users are able to update an expired root password and log in

### Procedure

### DIFF
--- a/recipes-connectivity/openssh/openssh_%.bbappend
+++ b/recipes-connectivity/openssh/openssh_%.bbappend
@@ -27,7 +27,8 @@ do_install:append () {
 	# customize sshd_config
 	sed -e 's|^[#[:space:]]*Banner .*|Banner /etc/issue.net|' \
 		-e 's|^[#[:space:]]*UseDNS .*|UseDNS no|' \
-		-e 's|^[#[:space:]]*PasswordAuthentication .*|PasswordAuthentication yes|' \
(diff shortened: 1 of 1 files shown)


chaitu236:
image

Should this be targeted to `nilrt/master/next` instead?

erickshepherdNI:
> Should this be targeted to `nilrt/master/next` instead?

This change was requested to go into the current dist by the Hardware Manager team to support changes they have already made for set on first use that will be releasing in C3.

chaitu236 (approved):

PR #1064, opened 2026-08-12
ni-firewall-dropins: add declarative drop-in contract and apply hook

Vairamani-NI:
### Summary of Changes

Adds `ni-firewall-dropins`: a declarative mechanism for opening firewalld services in the default zone. Any package that needs a service opened ships a marker file under `/etc/ni/firewall/open.d/<name>`.  A marker names one or more firewalld services (one per line; blank lines and `#` comments — whole-line or inline — are ignored); an **empty** marker defaults to the service named after the file, with any leading `NN-` ordering prefix and trailing `.conf` suffix removed.

A boot-time hook, `ni-firewall-apply`, reconciles the markers with firewalld on each run: it opens every service named by a current marker and closes every service it opened on a
previous run whose marker has since been removed. The set it has opened is tracked in `/var/lib/ni/firewall/applied`, so services opened outside this mechanism (e.g. the core NI services opened by `ni-firewalld-servicedefs`) are
never disturbed. It reuses the `ni-firewall-open` / `ni-firewall-close` helper for all offline/live and permanent/runtime handling. Wired into `packagegroup-ni-runmode` (x64).

### Justification
[AB#3964887](https://dev.azure.com/ni/DevCentral/_workitems/edit/3964887)

### Documentation
Internal Doc: [NILRT-Firewall-Setup-and-Configuration.md – Repos](https://dev.azure.com/ni/DevCentral/_git/AppCentral.wiki?path=/Organizations/Platform-HW-%26-Drivers/Products/RTOS/Tutorials-%26-Manuals/NILRT-Firewall-Setup-and-Configuration.md&version=GBwikiMaster)
Nilrt-doc: [Firewall Configuration with ](https://nilrt-docs.ni.com/firewalld/firewalld.html)[firewalld](https://nilrt-docs.ni.com/firewalld/firewalld.html)[ — ](https://nilrt-docs.ni.com/firewalld/firewalld.html)[nilrt](https://nilrt-docs.ni.com/firewalld/firewalld.html)[-docs 1.0 documentation](https://nilrt-docs.ni.com/firewalld/firewalld.html)

### Testing
Built and hardware-validated on an x64 cRIO-9045 (image flashed from a build of
this branch rebased on current `next`):

* Manual reconcile: open via marker, idempotent re-run, close on marker removal,
  and non-disturbance of core services — all pass.
* Boot-time reconcile: marker present at boot opens the service
  (`applied=[<svc>]`, runtime + permanent); marker removed before boot closes it
  (`applied=[]`); core `http/https/ni-*/ssh` services remain untouched.
* Init ordering verified: `S20firewalld` → `S25ni-firewall-apply`.

### Procedure

### DIFF
--- a/recipes-ni/ni-firewall-dropins/files/ni-firewall-apply
+++ b/recipes-ni/ni-firewall-dropins/files/ni-firewall-apply
@@ -1,0 +1,187 @@
+#!/bin/sh
+# ni-firewall-apply - reconcile declarative NI firewall drop-ins with firewalld.
+#
+# Contract:
+#   A package that needs a firewalld service opened in the default zone ships:
+#     1. the service definition  <service>.xml  in the firewalld services dir
+#        (e.g. ${libdir}/firewalld/services or /etc/firewalld/services), and
+#     2. a drop-in marker file   /etc/ni/firewall/open.d/<name>
+#   The marker names one or more firewalld services to open, one per line;
+#   blank lines and lines starting with '#' are ignored. A marker with no
+#   service lines defaults to a single service named after the marker file with
+#   any leading "NN-" ordering prefix and trailing ".conf" suffix removed.
+#
+# On each run this hook:
+#   * opens every service named by a current marker (idempotent), and
+#   * closes every service it opened on a previous run whose marker (or service
+#     line) has since been removed.
+# It reuses the ni-firewall-open / ni-firewall-close helper, so all offline/live
+# and permanent/runtime handling is shared. It only ever closes services it
+# recorded in its own state file, so services opened outside this mechanism
+# (such as the core NI services opened by the ni-firewalld-servicedefs postinst)
+# are never disturbed.
+
+set -u
+# Service names are literals; disable globbing so unquoted list expansions
+# (for svc in $desired, in_list ... $desired) can't match stray filenames.
+set -f
+
+OPEN_D=${NI_FIREWALL_OPEN_D:-/etc/ni/firewall/open.d}
+STATE_DIR=${NI_FIREWALL_STATE_DIR:-/var/lib/ni/firewall}
+STATE_FILE=$STATE_DIR/applied
+LOCK_FILE=$STATE_DIR/apply.lock
+
+prog=${0##*/}
+
+if ! mkdir -p "$STATE_DIR"; then
+	echo "$prog: cannot create state directory '$STATE_DIR'" >&2
+	exit 1
+fi
+
+# Serialize concurrent runs (boot plus a later package operation). Non-blocking:
+# if another run already holds the lock it will do the reconciliation, so just
+# exit cleanly rather than stacking a second pass. flock is required (RDEPENDS);
+# a missing flock or an unopenable lock file is a hard error, distinct from the
+# lock being held by another run.
+if ! command -v flock >/dev/null 2>&1; then
+	echo "$prog: flock not found; refusing to run without locking" >&2
+	exit 1
+fi
+if ! touch "$LOCK_FILE" 2>/dev/null; then
+	echo "$prog: cannot create lock file '$LOCK_FILE'" >&2
+	exit 1
+fi
+exec 9>"$LOCK_FILE"
+if ! flock -n 9; then
+	echo "$prog: another apply run is in progress; skipping" >&2
+	exit 0
+fi
+
+# in_list <item> [list...] - return 0 if <item> is present in the argument list.
+in_list() {
+	item=$1
+	shift
+	for e in "$@"; do
+		[ "$e" = "$item" ] && return 0
+	done
+	return 1
+}
+
+# Build the desired service set from the current markers.
+desired=""
+if [ -d "$OPEN_D" ]; then
+	# Enable globbing only to list the marker files, then disable it again.
+	set +f
+	set -- "$OPEN_D"/*
+	set -f
+	for marker in "$@"; do
+		[ -f "$marker" ] || continue
+		got_line=0
+		while IFS= read -r line || [ -n "$line" ]; do
+			# Drop inline/whole-line comments, then all whitespace.
+			svc=$(printf '%s' "${line%%#*}" | tr -d '[:space:]')
+			[ -n "$svc" ] || continue
+			got_line=1
+			in_list "$svc" $desired || desired="$desired $svc"
+		done < "$marker"
+		# Empty marker: derive the service from the file name.
+		if [ "$got_line" = 0 ]; then
+			svc=${marker##*/}
+			svc=${svc%.conf}
+			svc=${svc#[0-9][0-9]-}
+			in_list "$svc" $desired || desired="$desired $svc"
+		fi
+	done
+fi
+
+# Services opened by the previous run.
+previous=""
+if [ -f "$STATE_FILE" ]; then
+	while IFS= read -r svc || [ -n "$svc" ]; do
+		[ -n "$svc" ] || continue
+		previous="$previous $svc"
+	done < "$STATE_FILE"
+fi
+
+# Detect once whether firewalld is running (live) or not (offline) and the
+# default zone, mirroring ni-firewall-open, so we can tell whether a service was
+# already enabled before this mechanism opened it.
+if firewall-cmd --state >/dev/null 2>&1; then
+	fw_live=1
+	fw_zone=$(firewall-cmd --get-default-zone 2>/dev/null)
+else
+	fw_live=0
+	fw_zone=$(firewall-offline-cmd --get-default-zone 2>/dev/null)
+fi
+if [ -z "$fw_zone" ]; then
+	echo "$prog: could not determine the default firewalld zone" >&2
+	exit 1
+fi
+
+# svc_enabled_perm <svc> - return 0 if <svc> is enabled in the permanent/default
+# zone. Returns non-zero if the zone is unknown or the service is not enabled.
+svc_enabled_perm() {
+	[ -n "$fw_zone" ] || return 1
+	if [ "$fw_live" = 1 ]; then
+		firewall-cmd --permanent --zone="$fw_zone" --query-service="$1" >/dev/null 2>&1
+	else
+		firewall-offline-cmd --zone="$fw_zone" --query-service="$1" >/dev/null 2>&1
+	fi
+}
+
+rc=0
+applied=""
+
+# Open every desired service (idempotent). Record a service as "ours" only if we
+# opened it on a previous run, or if this run actually transitioned it from
+# disabled to enabled. This ensures we never later close a service that was
+# already enabled by something else (e.g. the core NI services).
+for svc in $desired; do
+	if in_list "$svc" $previous; then
+		ni-firewall-open "$svc" || rc=1
+		in_list "$svc" $applied || applied="$applied $svc"
+		continue
+	fi
+	pre=0
+	svc_enabled_perm "$svc" && pre=1
+	ni-firewall-open "$svc" || rc=1
+	post=0
+	svc_enabled_perm "$svc" && post=1
+	if [ "$pre" = 0 ] && [ "$post" = 1 ]; then
+		in_list "$svc" $applied || applied="$applied $svc"
+	fi
+done
+
+# Close services we opened before whose marker is now gone. Keep ownership if a
+# close fails so the next reconcile retries it instead of leaking the rule.
+for svc in $previous; do
+	in_list "$svc" $desired && continue
+	if ni-firewall-close "$svc"; then
+		continue
+	fi
+	rc=1
+	in_list "$svc" $applied || applied="$applied $svc"
+done
+
+# Persist the set we opened (one service per line) via a temp file + rename so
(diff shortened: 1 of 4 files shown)


Vairamani-NI:
@ni/rtos pls review this.

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply
```diff
@@ -0,0 +1,163 @@
...
+	svc_enabled_perm "$svc" && post=1
+	if [ "$pre" = 0 ] && [ "$post" = 1 ]; then
+		in_list "$svc" $applied || applied="$applied $svc"
+	fi
+done
+
+# Close services we opened before whose marker is now gone.
+for svc in $previous; do
+	in_list "$svc" $desired && continue
+	ni-firewall-close "$svc" || rc=1
```
rajendra-desai-ni: Can you comment on this info by copilot?

Ownership state can be lost when close fails, causing permanent drift in the reconciliation script, services removed from markers are closed in the loop at ni-firewall-apply:144. If ni-firewall-close fails (line 146), the service is still dropped from the next persisted applied set because state is rewritten from applied (desired-owned services) only (lines 152-160).
Result: the rule may remain enabled in firewalld, but this mechanism forgets ownership and will not retry closure on future runs.

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply.init
```diff
@@ -0,0 +1,25 @@
...
+# Required-Stop:
+# Default-Start:     2 3 4 5
+# Default-Stop:
+# Short-Description: Apply NI declarative firewall drop-ins
+# Description:       Reconcile /etc/ni/firewall/open.d markers with firewalld by
+#                    opening/closing services through ni-firewall-open.
+### END INIT INFO
+
+case "$1" in
+	start|restart|reload|force-reload)
```
rajendra-desai-ni, line 14: In the scenario when a new marker package is installed on a running system, how do we make sure `ni-firewall-apply` is automatically run for marker changes to take effect? Are we considering having postinst scripts for each of those packages that would run this script whenever new package is installed or are we not considering this?

recipes-ni/ni-firewall-dropins/ni-firewall-dropins.bb
```diff
@@ -0,0 +1,40 @@
...
+	# Persistent state directory for the apply hook.
+	install -d ${D}${localstatedir}/lib/ni/firewall
+}
+
+FILES:${PN} += "${sysconfdir}/ni/firewall/open.d ${localstatedir}/lib/ni/firewall"
+
+# Reuses the WI-A helper (ni-firewall-open/close) and firewalld's tools.
+RDEPENDS:${PN} += "ni-firewalld-servicedefs firewalld"
+# flock is used for run serialization; the hook degrades gracefully without it.
+RRECOMMENDS:${PN} += "util-linux-flock"
```
rajendra-desai-ni: Why just recommend and why not `RDEPENDS` on `util-linux-flock`?

rajendra-desai-ni (changes requested):

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply
Vairamani-NI: Fixed. A failed close now keeps the service in the persisted state so the next reconcile retries closing it, instead of forgetting ownership.

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply.init
Vairamani-NI: The opkg post invoke [feature](https://dev.azure.com/ni/DevCentral/_workitems/edit/4005638) is not done yet, I can create a separate ticket for the auto open port after package installation (@chaitu236 what do you think?), for now the apply script runs only after reboot.

rajendra-desai-ni (approved):

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply
```diff
@@ -0,0 +1,173 @@
...
+#        (e.g. ${libdir}/firewalld/services or /etc/firewalld/services), and
+#     2. a drop-in marker file   /etc/ni/firewall/open.d/<name>
+#   The marker names one or more firewalld services to open, one per line;
+#   blank lines and lines starting with '#' are ignored. A marker with no
+#   service lines defaults to a single service named after the marker file with
+#   any leading "NN-" ordering prefix and trailing ".conf" suffix removed.
+#
+# On each run this hook:
+#   * opens every service named by a current marker (idempotent), and
+#   * closes every service it opened on a previous run whose marker (or service
```
chaitu236, line 16: So, we're trying to only change port status of newly created/deleted services? What happens if we don't do this and do a generic `firewall-cmd --reload`? Would existing open ports or in-flight traffic be disrupted?

Vairamani-NI: --reload only re-applies firewalld's own permanent config; it has no knowledge of the open.d/ markers, so it can't open a service from a new marker or close one when a marker is removed. It would also drop other runtime-only rules. The hook translates markers into --add-service/--remove-service and only touches services it opened itself, so it never disturbs the core NI services or manual changes.

recipes-ni/ni-firewall-dropins/ni-firewall-dropins.bb
```diff
@@ -0,0 +1,38 @@
...
+	file://ni-firewall-apply \
+	file://ni-firewall-apply.init \
+"
+
+inherit update-rc.d
+
+INITSCRIPT_NAME = "ni-firewall-apply"
+# Start after firewalld (installed with update-rc.d "defaults" => S20) so the
+# reconciliation runs against the live daemon; no meaningful stop action.
+INITSCRIPT_PARAMS = "start 25 2 3 4 5 . stop 75 0 1 6 ."
```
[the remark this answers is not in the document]
Vairamani-NI: firewalld is not rcS.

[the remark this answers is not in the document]
chaitu236: @Vairamani-NI do we start firewalld at S20 in rc5.d? If so I think that may be too late as other network related services seem to start earlier (e.g., sshd, lldp, etc). And wouldn't this initscript also need to start before network services start? Ideally right after firewalld?

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply.init
chaitu236: It's ok to have another WI to track open-ports-post-package-installation work. But if we implement that, does that mean that `ni-firewall-apply` initscript will become obsolete (as all marker files will have already been processed post opkg operation and on a new boot there won't be any pending marker files to process)?

chaitu236: Also, why are we using marker files rather than just having the postinst scripts call `ni-firewall-apply` directly?

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply
```diff
@@ -0,0 +1,187 @@
+#!/bin/sh
+# ni-firewall-apply - reconcile declarative NI firewall drop-ins with firewalld.
+#
+# Contract:
+#   A package that needs a firewalld service opened in the default zone ships:
+#     1. the service definition  <service>.xml  in the firewalld services dir
+#        (e.g. ${libdir}/firewalld/services or /etc/firewalld/services), and
+#     2. a drop-in marker file   /etc/ni/firewall/open.d/<name>
```
chaitu236, line 8: As discussed offline, can you please attach an example marker file and .xml file to this PR.

Vairamani-NI: example is updated in the internal doc, pls refer PR description.

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply
```diff
@@ -0,0 +1,187 @@
...
+# are never disturbed.
+
+set -u
+# Service names are literals; disable globbing so unquoted list expansions
+# (for svc in $desired, in_list ... $desired) can't match stray filenames.
+set -f
+
+OPEN_D=${NI_FIREWALL_OPEN_D:-/etc/ni/firewall/open.d}
+STATE_DIR=${NI_FIREWALL_STATE_DIR:-/var/lib/ni/firewall}
+STATE_FILE=$STATE_DIR/applied
```
chaitu236, line 31: How do we handle package upgrade cases?

e.g., a package that previously only opened ports `A` and `B` is now upgraded and now has to open ports `B` and `C`.

After package installation, marker file is still present and state file is still present so `ni-firewall-apply` wouldn't process the changes; correct? If so this is a problem.

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply.init
Vairamani-NI: No — the two cover different situations.

A package's marker gets onto a device one of two ways:

1. It ships in the image (built into the rootfs at image-build time), or
2. It's installed later with opkg install on the running device.

In both cases, the same script — `ni-firewall-apply` — is what actually opens the ports by reading the marker files. The only difference is what triggers it to run:

1. image-shipped markers → the boot init script runs it at startup;
2. later opkg install → a post-invoke hook would run it after the opkg operation.

**The marker** file is the list of "what this package wants open." ni-firewall-apply takes no arguments — it always works by reading the markers. So even if a postinst called it, the package would still have to drop a marker first.

The marker also makes two other things work automatically:

- Removal: delete the package → its marker is gone → the next run closes only that service. With no marker there's nothing telling us to close it.
- Upgrade: the script compares the current markers to what it opened last time (see below).

If a `postinst` instead called `ni-firewall-open <service>` directly, the port would open, but we'd lose the automatic close-on-removal and the upgrade handling, and every package would carry its own open/close logic. The marker keeps packages simple: they just drop a file.

recipes-ni/ni-firewall-dropins/files/ni-firewall-apply
Vairamani-NI: It doesn't skip — it compares. On its next run ni-firewall-apply:

reads the current marker → should be open: `B, C`
reads its saved state (what it opened last time) → was open: `A, B`
opens C (new), leaves B (already open), closes A (no longer wanted), and saves the new state `(B, C)`.

So, the upgrade is handled correctly — the files being present is exactly what it needs to do the comparison, not a reason to skip. The one caveat is timing: this happens on the next boot or a manual run (or the future post-invoke hook), not the instant the upgrade finishes.

recipes-ni/ni-firewall-dropins/ni-firewall-dropins.bb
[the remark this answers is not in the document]
Vairamani-NI: Checked the actual boot order on a device (runlevel 5):
```
S01networking
S09sshd
S20firewalld
S20lldpd
S25ni-firewall-apply
```

Our hook (S25) is correctly right after firewalld (S20) — that's all it needs, since it has to run against a live firewalld and can't run before it.

You're right that firewalld itself (S20) starts after sshd (S09). But that's firewalld's own start position (it comes from the firewalld recipe's default update-rc.d setting), not something ni-firewall-dropins introduces or can change here — our hook can only run after firewalld. Also note our hook opens (allows) NI ports; opening them a little later at boot is harmless. Making firewalld start before sshd would be a separate change to the firewalld recipe, we need to discuss if thats needed.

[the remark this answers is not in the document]
chaitu236: > But that's firewalld's own start position (it comes from the firewalld recipe's default update-rc.d setting), not something ni-firewall-dropins introduces or can change here

We need to add a bbappend to change firewalld's init order. And change this hook to be earlier as well (right after firewall starts). If you want to address this in another PR, that's fine (please create a WI in this case) but this should happen.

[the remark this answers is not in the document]
Vairamani-NI: I will create a new WI and add it under successor of main firewalld feature.

recipes-ni/ni-firewall-dropins/ni-firewall-dropins.bb
```diff
@@ -0,0 +1,40 @@
...
+
+DEPENDS += "update-rc.d-native"
```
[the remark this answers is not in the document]
Vairamani-NI: removed.

recipes-ni/ni-firewall-dropins/ni-firewall-dropins.bb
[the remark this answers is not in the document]
Vairamani-NI: WI: https://dev.azure.com/ni/DevCentral/_workitems/edit/3470632

PR #1084, opened 2026-09-10
firewalld: start before network-facing services

Vairamani-NI:
### Summary of Changes

Order firewalld symmetrically around the network-facing services in the SysV boot/shutdown sequence, so the firewall is up before any service starts listening and torn down only after the network path is gone.

- **New file:** `recipes-connectivity/firewalld/firewalld_%.bbappend` sets `INITSCRIPT_PARAMS = "start 03 2 3 4 5 . stop 81 0 1 6 ."`, moving firewalld from the default `S20/K20` to `S03` on start and `K81` on stop.

firewalld defaults to `INITSCRIPT_PARAMS ?= "defaults"` (start 20 / stop 20), which is not ordered relative to the network-facing services:

**Start (rc5.d):** the default `S20` runs *after* `S09sshd`, and around `S21avahi-daemon` and `S25ni-firewall-apply`, so a listener can accept traffic before the ruleset is loaded. D-Bus (`S02`) is firewalld's only hard runtime dependency, so `S03` brings the firewall up before every listener:

| Priority | Service |
|---|---|
| S01 | networking |
| S02 | dbus-1 |
| **S03** | **firewalld** |
| S09 | sshd |
| S19 | syslog |
| S21 | avahi-daemon |
| S25 | ni-firewall-apply |

**Stop (rc6.d):** the default `K20` tears firewalld down *before* `K35systemWebServer` (80/443), `K36avahi-daemon` (5353) and `K39nisvcloc` (3580), while `K80networking` keeps the interface up. Since `firewalld.conf` ships `CleanupOnExit=yes`, stopping the daemon flushes all rules, leaving those ports exposed for the rest of the shutdown. `K81` (just after `K80networking`) tears the firewall down only once the network path is already down:

| Priority | Service |
|---|---|
| K09 | sshd |
| K35 | systemWebServer |
| K36 | avahi-daemon |
| K39 | nisvcloc |
| K80 | networking |
| **K81** | **firewalld** |

### Justification

[AB#4044873](https://dev.azure.com/ni/DevCentral/_workitems/edit/4044873)

### Testing

- **Build:** `build.core-feeds.sh && build.core-images.sh` green; firewalld ipk postinst registers the configured priorities via `update-rc.d`.
- **On hardware (cRIO-9045, x64):** flashed the BSI and confirmed the live `rc5.d` start order places `S03firewalld` before `S09sshd`, `S19syslog`, `S21avahi-daemon`, and `S25ni-firewall-apply`; `firewall-cmd --state` = `running`, default zone `public` with the expected services; clean reboot with ordering intact and no boot regression.
- **Shutdown:** verified the `rc6.d` `K`-link order and `CleanupOnExit=yes` on the device; `K81` places firewalld teardown after `K80networking`. Device re-validation of the updated stop priority is in progress.

### Procedure

* [ ] I certify that the contents of this pull request complies with the [Developer Certificate of Origin](https://github.com/ni/nilrt/blob/HEAD/docs/CONTRIBUTING.md#developer-certificate-of-origin-dco).


Vairamani-NI:
@ni/rtos pls review and approve.

chaitu236:
> The stop priority stays at 20 so firewalld shuts down before D-Bus (K37).

And do all other network services stop before firewalld stops?

Vairamani-NI:
> > The stop priority stays at 20 so firewalld shuts down before D-Bus (K37).
> 
> And do all other network services stop before firewalld stops?

No;
K39	-> nisvcloc
K80	-> networking

I will update to 81.

chaitu236:
> > > The stop priority stays at 20 so firewalld shuts down before D-Bus (K37).
> > 
> > 
> > And do all other network services stop before firewalld stops?
> 
> No; K39 -> nisvcloc K80 -> networking
> 
> I will update to 81.

It is odd that firewalld starts after networking but also stops after networking. Is that really correct?

Vairamani-NI:
> > > > The stop priority stays at 20 so firewalld shuts down before D-Bus (K37).
> > > 
> > > 
> > > And do all other network services stop before firewalld stops?
> > 
> > 
> > No; K39 -> nisvcloc K80 -> networking
> > I will update to 81.
> 
> It is odd that firewalld starts after networking but also stops after networking. Is that really correct?

Yeah, it's correct — the odd look is just because of one hard constraint. firewalld needs D-Bus, and D-Bus comes up after networking, so at boot firewalld simply can't start before the network (it lands right after, at S03). On shutdown there's no such constraint, so I keep firewalld alive until after networking goes down (K81). That way the firewall is up the whole time the network is actually carrying traffic. So, its simply firewall stays up for the entire time the network is up.

chaitu236:
> > > > > The stop priority stays at 20 so firewalld shuts down before D-Bus (K37).
> > > > 
> > > > 
> > > > And do all other network services stop before firewalld stops?
> > > 
> > > 
> > > No; K39 -> nisvcloc K80 -> networking
> > > I will update to 81.
> > 
> > 
> > It is odd that firewalld starts after networking but also stops after networking. Is that really correct?
> 
> Yeah, it's correct — the odd look is just because of one hard constraint. firewalld needs D-Bus, and D-Bus comes up after networking, so at boot firewalld simply can't start before the network (it lands right after, at S03). On shutdown there's no such constraint, so I keep firewalld alive until after networking goes down (K81). That way the firewall is up the whole time the network is actually carrying traffic. So, its simply firewall stays up for the entire time the network is up.

In that case, any reason why d-bus and firewalld shouldn't be moved to start before networking?

Vairamani-NI:
> > > > > > The stop priority stays at 20 so firewalld shuts down before D-Bus (K37).
> > > > > 
> > > > > 
> > > > > And do all other network services stop before firewalld stops?
> > > > 
> > > > 
> > > > No; K39 -> nisvcloc K80 -> networking
> > > > I will update to 81.
> > > 
> > > 
> > > It is odd that firewalld starts after networking but also stops after networking. Is that really correct?
> > 
> > 
> > Yeah, it's correct — the odd look is just because of one hard constraint. firewalld needs D-Bus, and D-Bus comes up after networking, so at boot firewalld simply can't start before the network (it lands right after, at S03). On shutdown there's no such constraint, so I keep firewalld alive until after networking goes down (K81). That way the firewall is up the whole time the network is actually carrying traffic. So, its simply firewall stays up for the entire time the network is up.
> 
> In that case, any reason why d-bus and firewalld shouldn't be moved to start before networking?

Starting D-Bus and firewalld before networking isn't really possible — networking is at S01 and the only free slot below it is S00, which isn't enough room for both (firewalld has to come after D-Bus anyway), and renumbering networking itself is a shared core service I don't want to disturb. It's also not worth it: nothing is listening that early, so there's no real exposure in that window.

On the other side, we don't want to stop firewalld before networking either. Stopping it flushes all the rules, so if it went down first, we'd be left with a live network interface and no firewall for the rest of the shutdown — which is exactly the state this whole change is trying to avoid. So, we keep firewalld up until after networking is down: as long as the network is up, the firewall is up with it.

rajendra-desai-ni (approved):

